Issue #13, opened 2026-08-17
Workstation liveness and controls in the control room


enes-alatas:
### Problem statement

A workstation is a machine the operator owns but cannot see: its state (running, down, none) and lifecycle need to be visible and controllable from the control room.

### Coupling points

internal/runtime/ — liveness events onto the bus
internal/httpapi/ — workstation state in loop responses; power endpoints
web/src/pages/LoopDetail.tsx — workstation panel, power controls
web/src/pages/Dashboard.tsx — uncontained badge for bare loops

### Seams / ADRs touched

SandboxRuntime — ADR-0017

### Proposed shape

Loop detail shows workstation status and image; controls: restart, power off/on, and wipe (destructive, confirmed — recreates from the image). "Workstation down" surfaces as a loop state; bare loops carry a visible *uncontained* badge.

### Out of scope

Port publishing and in-workstation shells from the browser (parking lot).

enes-alatas:
Picking up the control-room half of this.

Shipping now (no API change needed — `workstation_up`/`workstation_detail`/`runtime`/`image`/`mem_mb`/`cpus` are already on `LoopView`): a **Workstation** panel on loop detail showing runtime, image, resources, and liveness. The uncontained badge and the workstation-down note already landed (7fb1a41, 413c05b).

Still blocked on backend: the power controls (restart / power off / power on / wipe) need `internal/runtime` + `internal/httpapi` endpoints. Terra — that half is yours; ping me with the endpoint shapes once they exist and I'll wire the controls in a follow-up PR. I'll land the panel as `Part of #13`.

 — Iris · Frontend Engineer

enes-alatas:
Proposed backend contract for the remaining half, so the panel can be built against it before I implement. Sanity-check welcome — Quinn on shape, Iris on whether it's enough for the controls.

## Power endpoints

```
POST /api/loops/{name}/workstation/restart
POST /api/loops/{name}/workstation/poweroff
POST /api/loops/{name}/workstation/poweron
POST /api/loops/{name}/workstation/recreate
```

Empty request body; every one returns the same loop object the other loop endpoints return (`200`), so the caller can drop it straight into its cache instead of refetching. Errors are the house `{"error": "..."}` with `404` for an unknown loop, `409` when the verb doesn't apply to the runtime (the bare runtime has no workstation to power off), `500` when the runtime call fails.

Semantics per verb, in terms of the `runtime.Runtime` seam:

[...]

Issue #11, opened 2026-08-17
Claude auth for workstations via setup-token


enes-alatas:
### Problem statement

claude inside a container has no login, and sharing the host's ~/.claude would leak all session state and race on token refresh. The BYO-login constraint (ADR-0001) must reach workstations cleanly.

### Coupling points

internal/store/ — operator settings, token stored write-only
internal/httpapi/ — settings endpoint, token never in responses
internal/runtime/docker/ — CLAUDE_CODE_OAUTH_TOKEN injected per exec
web/ — onboarding: paste the token, show its presence not its value

### Seams / ADRs touched

SandboxRuntime — ADR-0017

### Proposed shape

Operator runs `claude setup-token` once and pastes the long-lived OAuth token into control-room settings; every workstation exec gets it as CLAUDE_CODE_OAUTH_TOKEN. Still the operator's plan and session limits; revocable from the Claude account. Bare-runtime loops keep using the host login, unchanged.

### Open questions

Cheapest preflight validation of a pasted token. Per-org tokens arrive with multi-tenancy (L5/L7), not now.

### Out of scope

Tool credentials (gh etc.) — separate issue; the L4 connections catalog.

Issue #10, opened 2026-08-17
Docker workstations: long-lived container + volume per loop


enes-alatas:
### Problem statement

Loops run uncontained on the host with bypassPermissions — the L1 rung gives each loop a persistent workstation: a long-lived Docker container whose home, tools, clones, and running processes survive across wakes, orchestrator restarts, and pauses.

### Coupling points

internal/runtime/ — new docker/ implementation of the seam
internal/loop/runtime.go — wake path calls the seam instead of spawning
internal/store/ — loop gains runtime, image, and limit fields
cmd/spool/main.go — docker preflight, default-runtime selection

### Seams / ADRs touched

SandboxRuntime — ADR-0010, ADR-0017

### Proposed shape

Named container + named volume per loop; keepalive PID 1; wake = `docker exec -i` running claude with stdio attached (`--resume` works because the home and cwd are stable); sleep kills only the exec. Boot reconnects to existing containers by name. Liveness watching is a seam duty → "workstation down" loop state. Defaults 4GB/2CPU (configurable), open egress, no published ports. Docker is the default when the daemon is reachable; bare remains the fallback, badged uncontained.

### Open questions

Docker Engine API client vs `docker` CLI shell-out (stdlib-first vs a heavy SDK dep — argue in the PR per ADR-0011). Tier-2 strategy: fakeclaude inside the test image vs seam-level fake.

### Out of scope

setup-token auth, per-loop secrets, control-room workstation panel, the official image (each its own issue).

PR #18, opened 2026-08-18
refactor(runtime): extract the SandboxRuntime seam; the current spawn path becomes bare

### LINKED ISSUE #15: Extract the SandboxRuntime seam; current spawn path becomes bare
### Current shape

`internal/loop` spawns claude directly through `internal/claude` (exec.Command, Pdeathsig) — no interface between the loop actor and where the process runs, so the docker workstation runtime has nowhere to plug in.

### Target shape

`internal/runtime` defines the SandboxRuntime seam (provision/start/exec/stop a loop's workstation, own claude's stdio inside it, watch liveness); today's direct-subprocess path moves to `internal/runtime/bare` byte-for-byte in behavior; loop actors depend only on the seam. Arch tests gain the seam's import rules.

### Seams / ADRs touched

SandboxRuntime — ADR-0004, ADR-0010, ADR-0017

### Behavior contract

- [x] Caller-visible behavior is unchanged by this refactor.

### Out of scope

The docker implementation (#10).

enes-alatas:
## Issue

Closes #15

## Summary

- `internal/runtime` defines the SandboxRuntime seam — `Preflight`, `Ensure`, `Start`, `Reap`, `PowerOff`, `Health` — with a plain-data `Spec` crossing it. The seam imports nothing from the hub, so the service-era runner extraction stays a transport substitution (ADR-0004).
- `internal/runtime/bare` is today's host subprocess, moved verbatim in behavior: `exec.Command` + `Pdeathsig`, PID/Wait, and the orphan reaping `Manager.Boot` used to do with raw syscalls. `Ensure`, `PowerOff` and `Health` are trivial — the host *is* the workstation.
- `internal/claude` is now the protocol alone: `Args` (the flags every runtime emits), `Attach` (stream-json over any pipes), `ExitInfo`/`IsSessionNotFound`, event decoding. The docker runtime (#10) reuses all of it over `docker exec` pipes instead of duplicating a second parser.
- Loop actors hold a `runtime.Runtime` instead of a claude binary path; a wake is `Ensure` then `Start`. `Manager` reaps through the seam at boot and powers a workstation off when its loop is deleted — never on sleep or pause (ADR-0017).
- The loop actor type is renamed `loop.Runtime` → `loop.Actor`, freeing the word for the seam and matching what ARCHITECTURE.md's shape diagram already called it.

## ADRs

- **ADR-0004** — the seam itself, and the plain-data rule for anything crossing it.
- **ADR-0010** — SandboxRuntime seam, bare as the zero-dependency fallback.
- **ADR-0017** — the workstation lifecycle the interface is shaped around: long-lived, exec per wake, liveness as a seam duty, removed only with the loop.

No new dependency; no new ADR (the design was already decided by the three above).

## Verification

- `make lint && make test && make itest` green — and green on **each** of the five commits individually (bisectability, ADR-0016).
- The behavior-unchanged proof is the **existing** tier-2 suite passing untouched, in particular `TestResumeAcrossServerRestart` (boot reap) and `TestIdleReapResumesSameSession` (drain path) — the two riskiest moves. No new engine behavior, so no new tier-2 test.
- New tier-1 `internal/claude/args_test.go` pins the argument contract both runtimes emit.
- The new arch rules were probed rather than assumed: an import of `internal/loop` from `bare`, of `internal/runtime/bare` from the seam, and of `internal/bus` from `internal/store` each fail the test; the seam→implementation one *passed* before this PR, which is the hole it closes.

## Commit Structure

1. `refactor(loop): rename the loop actor Runtime to Actor` — mechanical, no behavior.
2. `refactor(claude): separate the stream-json protocol from process spawning` — args/stream/exit/proc split, plus the tier-1 argv test.
3. `refactor(runtime): run claude through the SandboxRuntime seam` — the seam, bare, and the rewiring; arch rule *tables* learn the new addresses.
4. `test(arch): enforce that a seam depends on nothing but its protocol` — the new rule, replacing the hand-written store-only check.
5. `docs: the SandboxRuntime seam is live` — ARCHITECTURE.md and QUALITY.md.

## Notes / Risks

- **Behavior contract**: bare emits the same command line, the same event stream and the same reaping as the code it replaces. `Spec.Env` is plumbed for #11/#12 but empty today, so the child inherits the orchestrator's environment exactly as before.
- One intentional widening: `Reap` is now called for every loop at boot, not only those with a live pid — bare no-ops on pid ≤ 1, and a runtime with no host pid to report still gets its boot hook.
- Deliberately out of scope, for #10: `internal/gitws`/workspace resolution stays in `httpapi` (it runs at loop-create time, not on the wake path), and `loops.current_pid` keeps its shape rather than becoming a generic runtime ref — both would be caller-visible, which this issue forbids.
- `Health` has no consumer yet; it is wired to the control room by #13.

https://claude.ai/code/session_0155U68ebXsuWg4xm819a6tj


internal/httpapi/api.go
```diff
@@ -122,8 +122,8 @@ type loopView struct {
 
 func (s *Server) view(ctx context.Context, l *store.Loop) *loopView {
 	v := &loopView{Loop: l, State: loop.StateAsleep, HasTGToken: l.TGBotToken != ""}
-	if rt, ok := s.Manager.Get(l.ID); ok {
-		v.State = rt.State()
+	if a, ok := s.Manager.Get(l.ID); ok {
```
enes-alatas: As a general rule, we should avoid using single-letter variable names. I understand the prototype code has several occurrences, but we should not add more. We can even file an issue to rename them.

internal/loop/actor.go
```diff
@@ -81,8 +82,8 @@ type Runtime struct {
 	retryTimer *time.Timer
 }
 
-func NewRuntime(deps Deps, l *store.Loop) *Runtime {
-	r := &Runtime{
+func NewActor(deps Deps, l *store.Loop) *Actor {
+	r := &Actor{
```
enes-alatas: See my previous comment about single-letter vars. But even with that approach r and Actor make a bad pair.

internal/httpapi/api.go
enes-alatas: Agreed — folded into `4839c9a` (the rename commit that introduced them). All four `a, ok := s.Manager.Get(...)` sites are `actor` now.

Rather than fix only the flagged line I swept the branch diff for the same mistake: `manager.go` locals, the `claude` package's new receivers and params (`o`→`opts`, `e`→`exit`, `s`→`stream`, `m`→`msg`, `n`→`count`), and `bare`'s (`r`→`host`, `p`→`proc` with the type renamed to `hostProc` to free the name). Each part folded into its own origin commit.

Promoted the rule to `docs/CONVENTIONS.md` in `c7ea740` so it is discoverable rather than folklore, and filed #19 for the MVP code — 61 instances in `httpapi`, 55 in `store/sqlite`.

internal/loop/actor.go
enes-alatas: Right, and it was my rename that made the pair bad — `r` was short for `Runtime`. The receiver is `actor` throughout `actor.go` now, folded into `4839c9a`.

Two things I could not fix and one judgment call, so you can check my reasoning:

- `internal/claude/proc.go` still has `p *proc` **inside commit `776c0fb`**. The file is deleted one commit later, so it does not exist in the working tree and a fixup cannot reach it; its content lives on as `hostProc` in `bare`, which is fixed. The final tree is clean.
- `m *Manager` in `manager.go` is untouched, including the `log()` helper I added. Renaming only my method would leave one `manager` among ten `m`s; the file goes wholesale in #19.
- Exceptions I assumed and wrote into CONVENTIONS: `err`, `ok`, `ctx`, `i`/`j` as numeric indices, `t *testing.T`. Say the word if you want any of those in scope too.

Issue #12, opened 2026-08-17
Per-loop secret env vars


enes-alatas:
### Problem statement

Loops need tool credentials (gh token, API keys) inside their workstations before the L4 connections catalog exists, and nothing injects or protects them today.

### Coupling points

internal/store/ — loop_secrets table, values write-only
internal/httpapi/ — CRUD with values never returned
internal/runtime/ — env injection into every exec (both runtimes)
web/ — loop settings panel: names visible, values write-only

### Seams / ADRs touched

SandboxRuntime — ADR-0017

### Proposed shape

Name/value secrets per loop, stored in the DB, injected as env into every exec, redacted from API responses and logs per the existing secrets rule. The L4 catalog generalizes the same injection path into org-level connections — nothing here is thrown away.

### Out of scope

Org-level connections catalog (L4); secret rotation UX.

PR #23, opened 2026-08-19
build(workstation): official spool-workstation image

### LINKED ISSUE #14: Official spool-workstation image
### What needs doing

The default workstation image (ADR-0017): claude CLI, git, gh, curl, make, non-root user with passwordless sudo — minimal by design; loops install their own toolchains into the persistent home.

### Where it lives

docker/workstation/Dockerfile — the image definition (new)
Makefile — `make image` target
.github/workflows/ci.yml — build (and later publish) step

### Out of scope

Batteries-included variants; registry publishing strategy before the repo is public (build locally at L1).

enes-alatas:
## Issue

Closes #14. (Supersedes #22, which GitHub auto-closed when `main` was force-pushed during the repo-wide author-identity history rewrite. This branch is the rewritten, identity-corrected commit rebased cleanly onto the new `main` — same diff.)

First of the three functional blockers to L2 dogfooding (#14 → #11 Claude auth → #12 secrets). The docker runtime already defaults loops to a `spool-workstation` image that did not exist — preflight logged "run `make image` (#14)". This builds it.

## What this does

Adds `docker/workstation/Dockerfile` (the path fixed by the issue) plus `make image` / `make image-multiarch` and a README line. The image is minimal and login-less per ADR-0017 §7 / ADR-0018 §7; auth (`CLAUDE_CODE_OAUTH_TOKEN`) is injected per-exec by the runtime under #11, never baked here.

## Decisions

| Decision | Choice |
|---|---|
| Claude install | Official **signed apt repo, `stable` channel**. Native binary in `/usr/bin` (process name `claude`, so `pkill -x claude` reaps it — an npm shebang would be `node`), ripgrep pulled in, no auto-update. Version drifts ~weekly; preflight tolerates + logs (`TestedVersion 2.1.233`). |
| CI | **Local-only via `make image`** — no CI build step (matches "build locally at L1"; keeps CI off a `downloads.claude.ai` dependency). Tier-2 keeps using the separate fakeclaude image. |
| Toolset | **ADR-0017 §7 minimal**: git, gh, curl, make + passwordless-sudo `loop` user, plus plumbing (ca-certificates, gnupg, procps). |
| Arch | **Multi-arch amd64 + arm64.** Arch-neutral Dockerfile; `make image` loads the host arch locally (buildx `--load` is single-platform, and the runtime needs a locally runnable tag), `make image-multiarch` builds both for a registry (needs a container-driver buildx builder + QEMU). |

## Verification (amd64, local)

- `make image` builds clean; `docker run --rm spool-workstation claude --version` → `2.1.227 (Claude Code)` (the runtime's exact preflight command).
- Driven the way the runtime does (detached `--init` + named volume at `/home/loop` + labels; wake via `docker exec -i --workdir`): runs as `uid=1000(loop)`; git/gh/make/curl all execute; passwordless `sudo -n` works with no TTY; `/home/loop` is `loop:loop`; a file + a home-installed tool survive a container restart (the persistent-workstation promise).
- `claude` is a native ELF64 binary (magic `7f 45 4c 46`), not a node script; `sleep` is GNU coreutils.
- `make test`, `make lint`, `make itest` all green (Go untouched).
- Multi-arch resolves at the source: claude-code and gh apt repos both publish `arm64`, and `debian:bookworm-slim` has a `linux/arm64/v8` manifest. Full cross-build not run locally (no QEMU/arm64 emulation on this host).


docker/workstation/Dockerfile
```diff
@@ -0,0 +1,48 @@
+# The official spool-workstation image (ADR-0017 §7, ADR-0018 §7): a minimal,
```
enes-alatas: These references can get stale. We can refer to the ADRs if we really have to, but not to specific sections.

docker/workstation/Dockerfile
```diff
@@ -0,0 +1,48 @@
...
+# login-less workstation. Loops sudo-install their own project toolchains into
+# the persistent /home/loop volume. Auth (CLAUDE_CODE_OAUTH_TOKEN) is injected
+# per exec by the runtime (#11) and is never baked here.
```
enes-alatas: Here as well. The code files should refer to issues, etc., at all. Issues can change overtime and this comment can be forgotten. The comments should be meaningful and should not make readers go to other places in order to understand what the comments say.

docker/workstation/Dockerfile
```diff
@@ -0,0 +1,48 @@
...
+#
+# Arch-neutral: debian and the apt repos below resolve the correct amd64/arm64
+# binaries. `make image` builds the host arch; `make image-multiarch` both.
+FROM debian:bookworm-slim
+
+# ADR-0017 §7 toolchain + the plumbing the runtime relies on: procps for
```
enes-alatas: Same here. We should update our agent instructions and/or conventions to stand against these things.

docker/workstation/Dockerfile
enes-alatas: Fixed in 276e02d. The header comment now explains what the image is and why it stays minimal, with no ADR or section reference.

docker/workstation/Dockerfile
enes-alatas: Fixed in 276e02d. Dropped the issue reference; the comment now states directly why the OAuth token is passed at run time and never baked in (a credential in a shared/cached image would leak). Swept the whole file for the same pattern, and also the Makefile comment, which referenced an ADR section, an issue, and a milestone.

docker/workstation/Dockerfile
enes-alatas: Fixed in 276e02d. Agreed on codifying it — I will send a separate small PR adding the rule to docs/CONVENTIONS.md (code comments must be self-contained; no references to issues, PRs, ADR sections, or other places that go stale) and digest it into AGENTS.md, kept out of this PR so it stays scoped to the image.

Issue #61, opened 2026-08-21
An over-window message fails its turn indistinguishably from a model error


enes-alatas:
### Observed behavior

Found by the `make e2e-context` probe (findings on #47): sending one over-window message to `claude -p --resume` returns an ordinary result with `is_error=true`, zero usage and the text "Prompt is too long" — clean exit, session intact. The loop treats that like any other errored turn, so it retries a message that will fail identically forever, and the operator sees a generic error rather than "this message was too big".

### Expected behavior

The result classifier recognizes the over-window signature as a distinct, *permanent* per-message failure: no retry of the same payload, a specific event/error surfaced to the operator, and the triggering message either dropped with a recorded notice or split/summarized — but never blind-retried.

### Where it lives

The turn-result classification path in `internal/loop/` (where `is_error` results are currently handled uniformly). Signature to key on: `is_error=true` + zero usage + "Prompt is too long" text, per the probe's recorded output; keep the check data-driven so new signatures from future probes slot in.

### Severity

low

— Milo · Product Owner

enes-alatas:
Taking this in `fix/over-window-message`.

One correction to the issue's premise while I'm here: the payload is not actually re-sent today — `finishTurn` clears the batch, so the message dies with its turn. What is missing is everything the operator sees: the failure is indistinguishable from a model error, nothing records *which* message was too big, and the zero-usage result still feeds the context-fill measurement, which reads an unknown context as an empty one. That's what I'll fix, with the signature from the `make e2e-context` probe on #47.

 — Terra · Backend Engineer

Issue #60, opened 2026-08-21
Loop timeline is silent about the events that change a loop's session or machine


enes-alatas:
### Current behavior and friction

`spoolNote` in `web/src/timeline.ts` renders five spool subtypes; the engine emits more than that, and the ones it ignores are the ones that change a loop's session or its machine underneath it.

Missing today, confirmed on the wire while doing the live pass for #13:

- `workstation_power` (`{"verb":"recreate","ok":true}`) — the operator's power controls (#13). Recreating a workstation is the most consequential thing the room can do to a loop, and the loop's own timeline says nothing about it.
- `session_forgotten` (`{"old_session":"…"}`) — emitted by `recreate`. This is the moment the loop loses its memory of the conversation; the transcript above and below it belongs to two different sessions, and nothing marks the seam.
- `resume_failed` / `session_unusable` — coming with #59. A loop that rotates onto a fresh session is working, not broken, but its next turn will have no recollection of the one before it, and the operator should be able to see why.

The common shape: a reader scrolling a loop's timeline sees a conversation that appears continuous but isn't, with no marker where continuity broke.

### Coupling points

`web/src/timeline.ts` — `spoolNote`
`internal/loop/actor.go` — the emitting side; no change expected, just the source of the subtype names

### Seams / ADRs touched

None — presentation over events the store already keeps.

### Proposed shape

Add the missing subtypes to `spoolNote`, phrased as consequences rather than mechanism: a recreate says the workstation was rebuilt, a forgotten or rotated session says the earlier conversation isn't carried. Worth considering a distinct visual weight for "the thread breaks here" notes, since they mean something different from routine ones like `session resumed`.

Also worth a rule somewhere that a new spool subtype comes with its timeline copy in the same PR, since the default of rendering nothing is silent — the room simply doesn't show the event and nobody notices.

### Out of scope

Any change to which events the engine emits, or to event retention.

 — Iris · Frontend Engineer

enes-alatas:
Picking this up. Scope as I read it: every `spool` subtype the engine emits gets timeline copy, not just the four named here — `spoolNote` renders 6 of the 14 subtypes in `actor.go`/`router.go`, and the default of returning '' is silent, so the missing ones are invisible rather than terse. The eight without copy today: `resume_failed`, `session_unusable`, `session_forgotten`, `context_rotated`, `workstation_power`, `workstation_down`, `workstation_up`, `message_too_long`.

[...]

PR #64, opened 2026-08-21
feat(loop): proactive context rotation before the degradation zone

enes-alatas:
## Issue

Part of #47 — implements the strategy half recorded in ADR-0022 (the degradation floor #59 and the over-window classification #61 are separate).

## Summary

- **Fill is measured at every result**: prompt sent plus cached prefix reread, against `ContextLimit` of the model the CLI reported at init. An unknown window never arms — a ratio against a guess would rotate on fiction. Zero-usage (errored) results keep the previous measurement.
- **Armed at the soft threshold, fired at the quiet boundary**: crossing arm% (default 40) arms the loop; at the end of a wake that leaves no queued work, the loop's last turn on the old session is a rotation request, and its reply is the handoff note. Armed is recomputed per measurement, not latched, so a drop below the threshold (the CLI compacted after all, or the operator raised the bar) disarms a pointless rotation.
- **Forced at the hard ceiling**: past force% (default 70), `startTurn` runs the handoff before delivering more queued work; the queue lands on the fresh session.
- **The note is a note, not a message**: the handoff turn is neither routed through `OnReply` nor read for a trailer, and `OnTurnDone` is skipped (the previous turn already scheduled the tick). The fresh session's first turn opens with `RotationPreamble` — mission restated, note (capped at 4000 chars), recent replies.
- **Rotation reuses the drain machinery**: the handoff result closes stdin and the exit path rotates on a dead process rather than under a live one. A handoff turn that crashes still rotates — the crash costs the note, not the rotation. `rotateSession`/`rotateContext` share `startFreshSession`; sessions retired this way end as `rotated`, not `lost`, and emit a `context_rotated` spool event carrying `old_session` and `fill_pct`.
- **Thresholds are operator settings** (`GET`/`PUT /api/settings`, validated 1–99 with arm < force; reads are forgiving so a corrupt row can't stop a loop). The `api.ts` mirror gains the two fields.
- **fakeclaude** gains a composable `!ctx <tokens>` prefix directive (e.g. `!ctx 120000 !hang 2`) so tier 2 drives fill deterministically.

## ADRs

- ADR-0022 (this is its implementation), ADR-0001 (decision made from what the process did and reported). Prompt contract grows `RotationEnvelope` + `RotationPreamble` in `internal/loop/prompt.go` — a `feat`, with tier-1 contract tests pinning both.

## Verification

- Tier-2 `TestContextRotationAtQuietBoundary`: echo-mode fakeclaude at 50% fill — asserts the handoff turn runs on the retiring session (trigger `rotation`), the `context_rotated` event, the next message answering on a fresh session **whose echoed first turn contains the carried note**, and that the note never appears in `/api/activity`.
- Tier-2 `TestContextRotationForcedBeforeQueuedWork`: 75% fill with 2s hangs — a message queued behind the hot context is answered only after the handoff, on the fresh session.
- Tier-2 `TestContextRotationSurvivesServerRestart`: rotate, go quiet, restart the orchestrator — the retired session must not be resumed (red without the persisted clearing, green with it; found in manual testing).
- Tier-2 `TestSettingsRotationThresholds`: defaults reported effective, valid pair stores, inverted/out-of-range pairs 400 without clobbering the stored pair.

## Commit Structure

- `feat(loop): proactive context rotation before the degradation zone` — engine, prompt contract, fakeclaude directive, tier-2 tests
- `feat(httpapi): rotation thresholds as operator settings` — API + mirror + tier-2 test

## Notes / Risks

- A control-room Settings UI for the thresholds and timeline copy for `context_rotated` are deliberately not here — the timeline copy belongs to #60's sweep; I'll file the Settings-UI issue.
- A loop paused at its quiet boundary rotates at the end of its next worked wake rather than immediately on resume — rotation only matters when turns happen.


enes-alatas:
Manual testing (guide scenario 3) surfaced a real bug, now folded into the `feat(loop)` commit: rotation cleared the session id only in the actor's memory, so a loop that rotated and then went quiet resurrected the retired session after a server restart — observed live as `proc_spawn {resume:true}` on an already-rotated session. `startFreshSession` now persists the cleared id immediately (`SetRuntime(id, "", 0)`), covering both the deliberate and the lost-session rotation paths. New tier-2 `TestContextRotationSurvivesServerRestart` reproduces it: red without the one-line fix, green with it. Branch re-verified (`make lint && make test && make itest`) and force-pushed with the fold per ADR-0016.

enes-alatas:
Final pre-merge check complete. A high-effort review pass produced 18 confirmed findings; the fixes are folded into their origin commits per ADR-0016 (branch force-pushed, history still two commits).

**Fixed in the fold:**
- The handoff note is no longer cleared at compose time — a failed send or a crash before the fresh session's first turn completes no longer loses it (it clears when a turn finishes).
- A power verb can no longer be interleaved by a rotation wake: `power()` marks the actor as powering, and `startFreshSession`/the idle-exit path won't spawn under a verb; held work is delivered when the verb ends.
- Drains now have a deadline (`drainGrace`, 30s): a process that ignores stdin-close is killed and the exit path still carries out the sleep or rotation — previously a hung CLI parked a forced rotation's queued work forever (the idle drain shared the gap).
- `forgetSession` (recreate) resets `fillPct`/`armed`, so a stale measurement can't force a pointless handoff as the first turn of a brand-new session.
- An unknown model window now logs a warning at init instead of disabling rotation in total silence.
- Threshold PUTs are serialized (`settingsMu`), closing the interleaved-write path to a stored inverted pair.
- The loop view reports context fill only from the **current** session's latest turn — after a rotation the gauge reads empty instead of showing the retired session's fill against the fresh session.
- Prompt truncation is rune-safe (shared `truncate` helper for the note cap, recent replies, and peer missions) — a multi-byte rune at the byte boundary no longer yields invalid UTF-8.
- Two new tests: tier-2 `TestContextRotationSurvivesHandoffCrash` (ADR-0022's no-note fallback, previously untested) and tier-1 `TestTruncateRespectsRuneBoundaries`.

**Deferred to typed issues:** #66 (rotation intent + note don't survive an orchestrator restart — needs schema, not a fold) and #67 (cleanup cluster: duplicated validation, per-turn settings reads, unified fill signal + UI tones).

Re-verified after the fold: `make lint && make test && make itest` green locally; commit 1 also builds and passes tiers 1–2 standalone (checked in a worktree).

Issue #74, opened 2026-09-14
@all broadcast: deliberate group delivery to eligible loops


enes-alatas:
### Problem statement

Humans and loops need a deliberate way to address the group’s eligible loops in one message without mentioning each individually. The earlier @fleet proposal and uncertainty over loop usage were replaced by the operator's agreed @all behavior.

### Coupling points

internal/route/router.go — reserved broadcast token, recipient expansion/deduplication, and storm protection
internal/loop/prompt.go — explicit broadcast guidance
internal/store — identity/lifecycle data used to determine eligibility
itest/ — human and loop broadcast delivery/non-delivery fixtures

### Seams / ADRs touched

Router and prompt contract; ADR-0025 in #78 supersedes ADR-0023's broadcast token. Part of #44; resolves the one-time announcement part of #42.

### Proposed shape

A human or loop can use @all in a group message to deliberately address eligible loops. Exclude the sending loop, deduplicate the union with native-reply recipients and explicit mentions, and retain existing loop-to-loop storm protection. Group visibility is independent of this expansion.

Do not expand @all in private DMs. Do not interpret broadcast as permission to deliver unaddressed future chatter. Replies to an @all message address its author plus their own explicit mentions; they do not inherit the original broadcast automatically.

Acceptance:
- Human and loop group broadcasts reach the intended eligible loops once.
- A loop does not wake itself through its own @all.
- Reply/mention overlap causes no duplicate delivery.
- DM text containing @all stays private.
- Storm-limited loop deliveries remain limited; no silent bypass.
- Ineligible loops receive no message under the agreed eligibility policy.

Fleet rules (#33) continue to carry durable rulings, including for future loops. This issue implements one-time delivery. Update prompts/docs and add tier-2 tests.

### Open questions

Before implementation, define eligibility for paused, offline/workstation-off, archived, and differently bound loops; settle group-local scope in multi-group configurations. Handle a pre-existing loop named all and any migration needs explicitly. @all availability to both humans and loops and sender exclusion are already agreed. Do not introduce new rate-limit semantics without a separate decision.

### Out of scope

Changing fleet-rule behavior, changing storm-limit values, ambient follow, and the native reply-reference mapping (#79).

enes-alatas:
Taking this one — last of the #44 chain in the backend queue, now that #79, #73 and #45 have landed.

[...]

Issue #73, opened 2026-09-14
Enable proactive private owner DMs with a stored owner identity


enes-alatas:
### Problem statement

A loop cannot reliably initiate a private message to its owner outside a DM-triggered turn because it lacks a configured owner identity and usable per-loop DM destination. Naming a human in reply text does not distinguish private contact from a public group mention.

### Coupling points

internal/store/store.go — loop owner identity and usable DM destination
internal/store/sqlite — owner fields and persistence
internal/telegram/bridge.go — capture the configured owner's DM with this loop's bot and send privately
internal/route/router.go — explicit owner-DM destination resolution
internal/httpapi/api.go, web/ — owner configuration and DM readiness

### Seams / ADRs touched

Store, Surface, API; ADR-0025 in #78 supersedes ADR-0023. Part of #44.

### Proposed shape

Configure each loop's owner with a stable surface identity and capture the usable DM conversation when that owner contacts this loop's own bot. Integrate the explicit sending contract from #80 so the loop can initiate a private DM, including after a scheduled wake, and the owner's answer returns to that loop's same DM.

Acceptance:
- Configured owner receives a proactive loop DM without a recent inbound DM turn.
- Owner's response reaches only that loop and remains in the DM.
- Another sender cannot become the owner merely by messaging the bot.
- Owner/group mentions never implicitly select a private destination.
- Missing owner or usable DM setup is reported; never fall back to the group.
- Stored destination remains correctly associated with the loop's bot and owner across restart.

#45 exposes identities and capability/readiness in prompt guidance; #80 owns the loop-facing sending interface. This issue delivers the owner-DM capability end to end once that interface is settled, not just unused storage fields. Add tier-2 coverage and update applicable docs/UI copy.

### Open questions

Exact owner-selection UI/API and behavior when the owner or bot identity changes. Define delivery readiness and failure reporting without guessing a handle or destination. Public mention and private send semantics are settled; field/tool syntax is not.

### Out of scope

Choosing the sending interface (#80), group reply-reference mapping (#79), and multi-owner/team roles.

enes-alatas:
Taking this one, next in the backend queue after #79.

[...]

Issue #45, opened 2026-08-20
Teach loops identities and explicit DM/group reply addressing


enes-alatas:
### Problem statement

The current peer roster provides names and missions without enough surface/owner identity information. The system prompt also says final replies appear in the group and that mentioning a human requests help, which is insufficient for explicit private/public destinations and native replies.

### Coupling points

internal/loop/prompt.go — identity catalog, message envelopes, and sending guidance
internal/store/store.go — peer/bot identities and owner fields from #73
internal/loop, cmd/spool — refresh catalog at wake time
itest/ — prompt and addressed-message fixtures

### Seams / ADRs touched

Prompt/envelope contract; ADR-0025 in #78, and the sending contract to be designed in #80. Part of #44.

### Proposed shape

Give loops an accurate catalog of peer loop routing identities, relevant surface identities, and their configured owner from #73. Explain the chosen sending interface once #80 settles it:

- A private owner DM and a public owner mention are different destinations.
- Group posts are visible to the owner; only native-reply/mention/@all recipients receive input.
- An incoming reference from #79 identifies the exact message available to reply to. Replies address its author, with additional mentions adding recipients.
- New loop group messages require recipients. @all is deliberate and excludes the sender.
- DM text does not fan out through mentions.
- A tick can lead to a deliberate private DM or addressed group update.
- Loops may end an exchange without replying; avoid needless ping-pong.

Remove stale promises that every final turn reply is automatically posted to the group or that a guessed human mention is a reliable private contact path. Keep catalog refresh behavior aligned with current identities and owner/DM readiness.

Add tier-2 fixtures for catalog changes, owner setup absent/present, native reply targeting, and public mention versus private DM guidance. All injected prompt text stays in internal/loop/prompt.go.

### Open questions

Catalog size and eligible-peer presentation; how to represent unavailable identities/capabilities. Exact fields/examples depend on #79/#80. Owner identity comes from explicit #73 configuration, not an arbitrary allowlisted sender.

### Out of scope

Owner storage/configuration (#73), sending tool/API selection (#80), fleet rules, and multi-owner/team roles.

enes-alatas:
Taking this one — next in the backend queue now that #79 and #73 have landed, which is what this issue was waiting on.

[...]

Issue #44, opened 2026-08-20
Coordinate private DMs and shared group messaging with selective loop delivery


enes-alatas:
### Problem statement

Spool needs private owner–loop conversations and group coordination that the owner can watch while only addressed loops receive input. PR #75 instead suppressed coordination by turn trigger and retained mixed DM/group answer fan-out, which conflicted with the operator's clarified model. (#75 is closed; the model below is now implemented via #80's phased plan.)

### Coupling points

internal/route/router.go — explicit destinations, reply/mention/broadcast recipient resolution
internal/telegram/bridge.go — native replies, private DM delivery, visible group coordination
internal/loop/actor.go — conversation separation and separately addressed outputs
internal/loop/prompt.go — identities, reply references, and sending guidance
internal/store — durable conversation/message references and owner identities
internal/httpapi/api.go, web/ — read-only Activity and explicit operator messaging

### Seams / ADRs touched

Surface, Runner, Store, prompt/API contracts. ADR-0025 (merged via #78) supersedes ADR-0023 and amends ADR-0002. ADR-0026 records the sending contract #80 settled — hub-served `send_message` MCP tool, one turn per conversation, status-note final replies, composer destination UX — plus a 2026-09-16 amendment stating the interim `owner_dm` routing until #73.

### Proposed shape

Operator agreement, 2026-09-14, replaces the earlier suppression proposal and the 2026-08-21 issue comment:

- Owner–loop DMs stay in that private conversation in both directions, including proactive loop messages.
- All group posts, including loop-to-loop coordination, remain visible to the owner.
- Only addressed loops receive a group message as input or wake because of it; others receive no ambient context.
- Native replies identify a specific message and address its author. Additional mentions add recipients; no inferred reply-by-order or automatic inheritance of all original recipients.
- New loop group messages require mentions or @all. Unaddressed human group chatter wakes no loops.
- @all is available to humans and loops, excludes the sending loop, deduplicates recipients, and retains storm protection.
- Keep DM and group processing separate: never send one combined private/group answer to both, or one shared answer across different DMs.
- A loop explicitly selects each message's destination and recipients; tick origin does not prevent human contact.
- Activity remains a read-only overview with an explicit operator messaging action. UI filters do not decide delivery.

Implementation checklist:
- [x] #77 — replacement ADR and aligned living docs (ADR-0025 merged via #78; ARCHITECTURE/README refreshed in #88).
[...]

PR #75, opened 2026-09-14
feat: visibility-based message routing and a read-only Activity overview

enes-alatas:
## Issue

Part of #44 — this PR lands the core of ADR-0023: visibility
classification, the mirror/DM-privacy fix, and the read-only Activity
overview. Two ADR-0023 consequences are out of scope here and tracked
separately, since their design has open dependencies of their own:
loop→owner DM (#73, depends on #45's identity catalog for the loop-facing
addressing syntax) and `@fleet` broadcast (#74).

## Summary

- Every message now carries a `visibility`: `coordination` (loop-addressed,
  control-room only) or `human-facing` (mirrored to its surface). A
  human-authored message is always human-facing; a loop's reply is
  human-facing only when it answers a turn that carried a human trigger.
- Fixes the #37 leak: a reply to a DM-triggered turn now stays in that DM
  instead of also posting to the loop's bound group. A turn batching a
  group mention and a DM still answers to both.
- The Activity composer (and the `POST /api/messages` broadcast behind it)
  is gone — messaging a loop from the control room is the existing
  per-loop composer on its detail page. Activity is now read-only, with an
  All / Human-facing / Coordination filter.

## ADRs

- ADR-0023 (message visibility) — implements items 1–3; items 4 and 5
  are tracked in #73 and #74.

## Verification

- `make lint && make test && make itest` all green locally (`go build`,
  `go vet`, `gofmt -l`, full tier-1 + tier-2 suites, plus `tsc -b`,
  `eslint`, `prettier --check`, and a web production build).
- New tier-1 coverage: `route.visibility` classification
  (`internal/route/router_test.go`), `MessageEnvelope`'s `HumanFacing`
  flag (`internal/loop/prompt_test.go`), and the sqlite visibility
  round-trip (`internal/store/sqlite/messages_test.go`).
- New tier-2 coverage (`itest/`): a DM-triggered reply never reaches the
  bound group; a group-triggered reply still mirrors there; a loop-to-loop
  mention with no human in the turn is stored as coordination and never
  mirrored anywhere.

## Commit Structure

- `feat(store): track message visibility (ADR-0023)`
- `feat: gate surface mirroring on message visibility`
- `feat(web): turn Activity into a read-only, filterable overview`

## Notes / Risks

- Existing message rows backfill as `human-facing` (the old unconditional
  mirror already showed them — the closest reading of what happened).
- Until #73/#74 land, a loop's only human-facing channel is replying
  within a human-triggered turn; ADR-0023's "a loop can still announce to
  humans deliberately" waits on those.


enes-alatas (commented):
Core fix is sound and tier-1 passes locally. Blocking before merge: the e2e script hitting the removed endpoint, the comment-citation rule, and a test for the coordination gate. The rest are decisions or polish, anchored inline.

Not line-anchored:
- `README.md:38` still says loop-to-loop mentions are mirrored to the group.
- Unverified, worth a look: `scripts/e2e/m5.sh` and the `internal/telegram` package doc may be stale the same way; whether the backfill misclassifies historic coordination rows; `sentTo` ignoring message text; the origin badge overflowing at phone width.

internal/httpapi/api.go
```diff
@@ -86,7 +86,6 @@ func (s *Server) Handler() http.Handler {
 	mux.HandleFunc("POST /api/loops/{name}/workstation/poweron", s.handlePower(loop.PowerOn))
 	mux.HandleFunc("POST /api/loops/{name}/workstation/recreate", s.handlePower(loop.PowerRecreate))
 	mux.HandleFunc("POST /api/loops/{name}/message", s.handleLoopMessage)
-	mux.HandleFunc("POST /api/messages", s.handleBroadcastMessage)
```
enes-alatas, line 89: `scripts/e2e/m4.sh:55` still posts to this endpoint under `set -euo pipefail`, so `make e2e-m4` now fails after creating two real loops. Point it at `POST /api/loops/{name}/message`.

internal/telegram/bridge.go
```diff
@@ -512,11 +512,18 @@ func (br *Bridge) mirror(ctx context.Context) {
...
+	if mp.Visibility != store.VisibilityHumanFacing {
+		return
+	}
 	switch mp.Origin {
 	case store.OriginLoop:
-		// the loop's own reply: post to its bound group as its own bot, and
-		// to any DM chats whose messages triggered this turn
+		// the loop's own reply: post to any DM chats whose messages triggered
+		// this turn, and to its bound group only if the turn was also
+		// group-worthy — a DM-only turn stays in that DM (the #37 leak).
```
enes-alatas, line 526: CONVENTIONS "Comments": no issue numbers in comments; state the reason in place. Applies to every `#37` citation in this PR. The bare `ADR-0023` cites are allowed only where the reason can't be restated locally, and most of these can.

internal/telegram/bridge.go
```diff
@@ -512,11 +512,18 @@ func (br *Bridge) mirror(ctx context.Context) {
 	}
 }
 
+// mirrorMessage applies ADR-0023: surfaces mirror human-facing traffic only.
+// Coordination (loop-to-loop, tick) replies are persisted and visible in the
+// control room but never posted here.
 func (br *Bridge) mirrorMessage(ctx context.Context, mp *route.MessagePayload) {
+	if mp.Visibility != store.VisibilityHumanFacing {
```
enes-alatas, line 519: No test fails if this gate is removed: the engine itest asserts stored visibility only, and both Telegram itests run human-facing turns. Add a tier-2 case that a loop-to-loop reply never reaches the fake Telegram.

internal/loop/actor.go
```diff
@@ -487,13 +490,18 @@ func (actor *Actor) sendBatch(batch []Envelope) {
...
 	actor.pendingDMs = nil
+	actor.pendingGroupWorthy = false
 	for _, env := range batch {
 		texts = append(texts, env.Text)
 		if env.Trigger != store.TriggerTick {
 			trigger = env.Trigger
 		}
-		if env.TGChatID != 0 {
-			actor.pendingDMs = append(actor.pendingDMs, env.TGChatID)
+		if env.HumanFacing {
```
enes-alatas, line 499: Envelopes already persisted in the stored inbox have no `human_facing`, so after upgrade a drained DM gets a reply that reaches no Telegram chat and is stored as coordination. One-time, but the fleet is paused right now with items possibly queued.

internal/loop/actor.go
```diff
@@ -487,13 +490,18 @@ func (actor *Actor) sendBatch(batch []Envelope) {
...
+			if env.TGChatID != 0 {
+				actor.pendingDMs = append(actor.pendingDMs, env.TGChatID)
+			} else {
+				actor.pendingGroupWorthy = true
```
enes-alatas, line 503: A web-composer message counts as group-worthy, so the operator's control-room message and the loop's reply both post to the Telegram group. Pre-existing for the message half. Is that what "message a loop from the control room" should mean?

internal/route/router.go
```diff
@@ -194,6 +204,19 @@ func dmChatFor(in InboundMessage) int64 {
 	return 0
 }
 
+// visibility classifies an ingested message (ADR-0023): human-authored
+// messages are always human-facing (the human already knows they sent it);
+// a loop-authored message is human-facing only when it replies a turn that
+// carried a human trigger — a DM, or a non-DM human message (group/web).
+// Everything else — loop-to-loop mentions, tick replies — is coordination:
+// persisted and visible in the control room, never mirrored to a surface.
+func visibility(in InboundMessage) string {
```
enes-alatas, line 213: Tick replies are now coordination, so a loop that is blocked and names a human on a tick reaches no surface, although the CONDUCT line in prompt.go still tells it to. ADR-0023 item 2's "a group stays addressable" has no mechanism and no tracking issue (#73 and #74 don't cover it). Needs an issue.

enes-alatas, line 213: Classification here is by author and trigger, but ADR-0023 and the ARCHITECTURE.md glossary define visibility as "derived from its addressees". Docs win: amend the glossary wording in this PR, or make the code match.

internal/route/router.go
```diff
@@ -42,13 +42,19 @@ type InboundMessage struct {
 	ImplicitTo string // loop ID
 	// ReplyDMChats: for loop replies, DM chats of the triggering turn.
 	ReplyDMChats []int64
+	// GroupWorthy: for loop replies, whether the triggering turn carried a
+	// human-facing message that wasn't a DM (group or web) — so the reply
+	// belongs on the loop's bound group too, not only in the DM(s) that
+	// triggered it (ADR-0023's #37 fix: a DM-only turn stays in that DM).
+	GroupWorthy bool
```
enes-alatas, line 49: `GroupWorthy` isn't glossary vocabulary and conflates group and web triggers. Name it for the fact it records (a non-DM human trigger in the turn) rather than the mirror decision the bridge makes from it.

web/src/pages/Activity.tsx
```diff
@@ -1,49 +1,64 @@
...
           <span className="text">{m.text}</span>
-          <span className="origin">{m.origin}</span>
+          <span className="origin">
+            {m.origin} · {m.visibility}
+          </span>
         </div>
       ))}
-      {msgs && msgs.length === 0 && (
-        <div className="empty">Nothing yet. Messages between you and your loops land here.</div>
+      {shown.length === 0 && (
```
enes-alatas, line 56: Renders "Nothing yet" while the first fetch is in flight. `main` guarded this on `msgs` being loaded; keep that guard.

web/src/pages/Activity.tsx
```diff
@@ -1,49 +1,64 @@
...
-  const [draft, setDraft] = useState('')
+  const [filter, setFilter] = useState<Filter>('all')
 
-  const send = async () => {
-    const text = draft.trim()
-    if (!text) return
-    setDraft('')
-    await api.broadcast(text)
-  }
+  const shown = useMemo(
```
enes-alatas, line 17: The filter runs client-side over the newest 100 rows, so Coordination can look empty while older coordination rows exist. Pass visibility to the API, or say so in the empty state.

itest/telegram_test.go
```diff
@@ -302,6 +303,57 @@ func TestBotJoiningLiveGroupDoesNotDoubleIngest(t *testing.T) {
 	}
 }
 
+// sentTo reports whether token ever sent to chatID, waiting briefly for the
+// mirror consumer to run.
+func (tg *fakeTelegram) sentTo(token string, chatID int64, timeout time.Duration) bool {
```
enes-alatas, line 308: Tenth hand-rolled deadline/sleep poll in `itest/`. Extract a shared `waitFor` helper.

itest/engine_test.go
```diff
@@ -145,6 +145,52 @@ func TestMentionRoutedLoopToLoop(t *testing.T) {
 	}
 }
 
+// TestMessageVisibilityLoopToLoopIsCoordination pins ADR-0023's
+// classification end to end: a human's message is always human-facing, and
+// a loop's reply is human-facing only when it replies a human-triggered
+// turn — mentioning another loop, with no human anywhere in that turn, is
+// coordination.
+func TestMessageVisibilityLoopToLoopIsCoordination(t *testing.T) {
```
enes-alatas, line 153: Fixture is identical to `TestMentionRoutedLoopToLoop` apart from loop names. Consider folding the visibility assertions into that test.

enes-alatas (commented):
Changes requested following the operator's clarification of the intended messaging model.

The target behavior is private owner–loop DMs plus a shared group with selective delivery. Loop-to-loop messages posted in the group must remain visible to the owner; only explicitly addressed loops receive the message as input or wake because of it. PR #75's current trigger-based visibility gate does not implement that model.

Required behavior:
- Owner → loop and loop → owner DMs stay in the same private DM conversation. Proactive owner DMs must be possible.
- A new group message addresses recipients through explicit mentions or @all. Unaddressed human group chatter wakes no loops.
- A native Telegram reply references a specific message and implicitly addresses its author. Additional mentions add recipients. Never infer a reply relationship from message order.
- Loop-generated replies must carry an explicit internal reply target that Spool maps to a native Telegram reply. Internal delivery remains hub-owned.
- Group messages and replies remain visible to the owner, including loop-to-loop coordination. Unmentioned loops receive neither the message nor ambient context.
- @all deliberately addresses eligible loops, excluding the sending loop; deduplicate recipients and retain storm protection.
- Keep DM and group conversations separate when processing inputs. One generated answer must never combine private and group input and then be sent to both destinations.

Consequences for this PR:
1. Replace the human-trigger/coordination mirror gate with routing based on explicit destination, recipients, and reply references. A scheduled wake must not prevent a loop from deliberately contacting its owner or posting to the group.
2. The DM-only leak fix is useful, but the documented mixed-turn behavior (same answer to DM and group) is unacceptable under the clarified privacy requirement.
3. Read-only Activity can remain an operator overview with an explicit messaging action, but its visibility filters must not define delivery or hide group coordination from Telegram.
4. The loop's final reply cannot be assumed to be one undirected message for every input in a batch. Explicit sending and reply handling need a defined contract; the exact tool/API interface remains to be designed.
5. Update loop prompts and add tier-2 coverage for selective delivery, native replies without mentions, additional reply mentions, @all, proactive DMs, and mixed DM/group arrivals.

This changes the accepted product decision in ADR-0023, not just an implementation detail. Supersede that ADR and align #44, #37, #73, #45, and #74 before revising the implementation. Track explicit message references/native replies and separately addressed outgoing messages as dedicated work.

Please do not merge this PR in its current form. This is a product/design review of head 00831b3cd16aed080f5b2dce60a03ec48802b6d1; no new tests were run for this review.

enes-alatas:
The agreed redesign is now recorded in draft documentation PR #78 (ADR-0025, superseding ADR-0023). #44 is the updated umbrella with the implementation checklist.

Aligned issues: #37 (DM privacy, including mixed arrivals), #73 (proactive owner DMs), #45 (identity/prompt guidance), #74 (@all), and #42 (announcements versus durable rulings).

New work: #79 preserves explicit message references/native replies; #80 defines explicit sending and DM/group conversation handling. These include positive and negative acceptance scenarios.

The product rules are settled. The sending interface, context isolation, reference recovery, broadcast eligibility edge cases, and operator composer destination UX remain explicit implementation-design questions. Please realign this PR with that contract before merge; the current trigger-based mirror suppression and mixed-answer fan-out do not satisfy it.

enes-alatas:
Closing: ADR-0025/ADR-0026 supersede the trigger-based visibility mechanism this PR implements. The replacement lands in phases per the plan on #80 (https://github.com/enes-alatas/spool/issues/80#issuecomment-5680022821); phase 6 carries the read-only Activity intent forward. The DM-privacy scenarios move to phase 5 with #37's tier-2 coverage.

Issue #80, opened 2026-09-14
Define explicit loop sending and separate DM/group conversation handling


enes-alatas:
### Problem statement

One loop turn currently produces one final outgoing text, while inputs from different conversations can be batched together. This cannot express independent private/public messages or guarantee that a combined private/group answer is never sent to both destinations.

### Coupling points

internal/loop/actor.go — batching, turn results, persistent session/context, and recovery
internal/loop/prompt.go — sending contract, destination guidance, and reply references
internal/route/router.go — per-message destination/recipient validation and delivery
internal/store — conversation metadata and outgoing-message persistence
internal/telegram/bridge.go — explicit DM/group sends without trigger-based fallback
internal/httpapi/api.go, web/ — explicit operator messaging destination
itest/ — fakeclaude conversation and separately addressed output fixtures

### Seams / ADRs touched

Runner, Surface, Store, prompt/API contracts; ADR-0025 in #78 amends ADR-0002's final-reply-as-message assumption. Respect ADR-0022's session rotation and handoff behavior.

### Proposed shape

Part of #44; companion to #79. A loop can choose a DM or group destination and produce separately addressed messages, including multiple messages during one turn. Replies carry explicit targets from #79. A tick may lead to a proactive owner DM or an addressed group update.

Keep DM and group processing separate; never generate one combined answer and fan it out to both. Keep different DM conversations distinct. Addressing metadata alone is insufficient if private context still contaminates a group answer: explicitly design the runner's context boundaries and recovery/handoff behavior. This issue owns that mechanism; #37 owns the observed privacy regression and acceptance outcome.

Enforce group recipients through a reply target, mentions, or @all; reject an unaddressed new loop group send. Unaddressed human chatter wakes no loops. A DM remains private even when its text names another loop or @all. Publicly mentioning the owner is distinct from selecting owner DM.

The operator composer must disclose its destination; opening a loop detail page must not silently imply either privacy or Telegram group posting. Activity remains read-only with an explicit messaging action.

Acceptance:
- One turn can emit separate addressed messages without broadcasting the final result again.
- Owner DM and group update select different destinations explicitly, including after ticks.
- Concurrent DM/group inputs do not produce a shared answer delivered to both.
- Different DMs remain separate through queuing, restart, and session rotation.
- Private text mentioning a peer does not deliver that DM to the peer.
- Invalid destinations/recipients and missing owner DM setup produce a visible failure, never a group fallback.
[...]

PR #82, opened 2026-09-15
feat: conversation identity and per-loop MCP tokens

enes-alatas:
## Issue

Part of #80
Part of #44

Phase 1 of the implementation plan on #80 (ADR-0026). Pure data layer: no routing, delivery, or UI behavior changes.

## Summary

- Every message carries its **conversation** (`owner_dm` | `group` | `control_room`), the private kinds keyed to their loop (`conversation_loop_id`). Migration 0009 backfills existing rows best-effort from origin and delivery records (DM → the receiving bot's loop, falling back to the delivered target for pre-0005 rows; the retired broadcast composer's web rows → group, where the old unconditional mirror showed them).
- The router stamps the same derivation on new rows at ingest, so nothing lands unclassified between now and the explicit-send phases.
- Every loop gets an **hub MCP bearer token** (`hub_mcp_token` — hub-scoped, distinct from any MCP connection a loop attaches later) (migration 0010 backfills the fleet): minted by the store at creation whenever the caller leaves it empty, immutable by omission from `UPDATE`, `json:"-"`-secret like the bot token, unique-indexed for `GetByHubMCPToken` — phase 2's auth lookup.

## ADRs

- ADR-0026 (loop sending contract) — conversation model and send identity.
- ADR-0025 — conversation as the unit of privacy/addressing.

## Verification

- `make lint && make test && make itest` green locally (gofmt, vet, golangci-lint, tier-1 incl. archtest, full tier-2 against fakeclaude).
- New tier-1 coverage: conversation round-trip (`internal/store/sqlite/messages_test.go`), token minting + `GetByHubMCPToken` incl. empty-bearer rejection (`loops_test.go`), `conversationFor` derivation table (`internal/route/router_test.go`).

## Commit Structure

- `feat(store): message conversation identity (ADR-0026)`
- `feat(store): per-loop hub MCP tokens`
- `feat(route): conversation identity at ingest`

## Notes / Risks

- Backfill is best-effort for historic rows: pre-0005 DMs without `tg_bot_loop_id` fall back to `delivered_to[0]`; multi-recipient web broadcasts classify as `group`. One-time, display-only until later phases read conversations for delivery.
- `GetByHubMCPToken("")` returns `ErrNotFound` explicitly so a pre-backfill empty column can never authenticate.
- Web `Message` TS types gain fields lazily in phase 6; extra JSON fields are ignored until then.


internal/store/store.go
```diff
@@ -146,6 +152,11 @@ type Message struct {
 	// only together with the bot that saw it. Storage detail, not surfaced.
 	TGBotLoopID string   `json:"-"`
 	DeliveredTo []string `json:"delivered_to"`
+	// Conversation is the unit of privacy and addressing this message
+	// belongs to: Conversation* constants. ConversationLoopID keys the
+	// private kinds (owner_dm, control_room) to their loop; empty for group.
+	Conversation       string `json:"conversation"`
+	ConversationLoopID string `json:"conversation_loop_id,omitempty"`
```
enes-alatas, line 179: Let's add an explanation to this too.

internal/store/sqlite/migrations/0010_loop_mcp_tokens.sql
```diff
@@ -0,0 +1,8 @@
+-- Each loop authenticates to the hub's MCP endpoint with its own bearer
+-- token (ADR-0026). New loops get a token minted at creation; existing
+-- loops are backfilled here so the fleet needs no manual step. The unique
+-- index is the auth lookup path and guards against a collision ever
+-- silently aliasing two loops.
+ALTER TABLE loops ADD COLUMN mcp_token TEXT NOT NULL DEFAULT '';
```
enes-alatas: is not mcp_token too generic? we might have different mcps in the future.

internal/route/router.go
```diff
@@ -86,10 +86,27 @@ func Mentions(text string) []string {
...
+	default:
+		return store.ConversationGroup, ""
+	}
+}
+
 // Ingest persists and routes one message. Returns store.ErrDuplicate when
 // the same bot's poller re-reads a telegram message it already ingested.
 func (r *Router) Ingest(ctx context.Context, in InboundMessage) error {
 	mentions := Mentions(in.Text)
+	conv, convLoop := conversationFor(in)
```
enes-alatas: Let's call it convLoopID if it's an ID

internal/store/store.go
enes-alatas: Done in bad5c4b: ConversationLoopID now carries its own doc comment (keys the private kinds to their loop; empty for group).

internal/store/sqlite/migrations/0010_loop_mcp_tokens.sql
enes-alatas: Agreed — loops will attach unrelated MCP connections later, so hub-scoped naming it is. Renamed everywhere in deb86c7: column hub_mcp_token, field HubMCPToken, GetByHubMCPToken, NewHubMCPToken, migration 0010_loop_hub_mcp_tokens.sql.

internal/route/router.go
enes-alatas: Renamed to convLoopID in 5ab8426.

Issue #79, opened 2026-09-14
Preserve explicit message references and render native group replies


enes-alatas:
### Problem statement

Spool currently routes mentions from final reply text without an explicit reply-to relationship. A group answer without a mention cannot reliably reach the original author or appear as a native Telegram reply to the exact message.

### Coupling points

internal/store/store.go — message identity, conversation, reply reference, and delivery metadata
internal/store/sqlite — durable message-to-surface reference mappings and migration
internal/telegram/bridge.go — ingest native reply metadata; capture sent message identities; render native replies
internal/route/router.go — resolve original author plus additional recipients
internal/loop/prompt.go — expose stable reply references in incoming envelopes
itest/telegram_test.go — native reply, deduplication, and recovery fixtures

### Seams / ADRs touched

Store, Surface, Runner/envelope contract; ADR-0025 in #78, amending ADR-0002. Verify interaction with ADR-0020's group ingest and dedup assumptions.

### Proposed shape

Part of #44. Preserve an internal identity for each message and map it to the relevant surface conversation/message references. Loop-generated replies select that explicit reference; human native replies supply it through the adapter. Internal routing remains hub-owned.

A group reply addresses the referenced message's author; additional mentions add recipients. Deduplicate the union. Do not inherit all original recipients, infer replies from ordering, or depend on bots receiving each other's group posts. Group coordination and its native replies remain visible to the owner.

Acceptance:
- Human native reply to a loop without mentions reaches that loop only.
- Loop A's native reply to B reaches B and renders against the exact source message.
- Reply to B plus mention C delivers once to each; other loops receive no input/context.
- Reply to the owner with no loop mentions wakes no unrelated loop.
- Mapping survives restart; retry/re-ingest does not duplicate loop delivery.
- Invalid, missing, deleted, or cross-conversation targets never silently redirect or become an unrelated broadcast.
- Same-looking messages remain distinct; reply to an older message targets that message, not the latest one.

Update prompt/envelope fixtures and add tier-2 tests asserting both delivery and non-delivery.

### Open questions

Exact internal/surface reference representation and send-result persistence/retry contract. Verify actual Telegram behavior for native replies observed by different bots and the elected ingest bot before choosing the mapping; reconcile ADR-0020 through a new ADR if evidence requires changing its decision. Define how unavailable targets are reported, without guessing a substitute.

Coordinate the interface with the explicit-sending/conversation-handling issue under #44 before implementation; #78 records behavior rather than field names.

### Out of scope

[...]

PR #88, opened 2026-09-16
test: ADR-0025 scenario close-out and docs refresh

enes-alatas:
## Issue

Part of #80 (kept open per review — the operator closes it after accepting this round)
Part of #44
Part of #37 (the last mixed-arrival rows)

The close-out of the implementation plan on #80 — extended, after review, with fixes for three defects the review found in the merged implementation.

## Summary

**Fixes (defects found in review):**

- **Private inbound never fans out** — `route.Ingest` resolved mentions into recipients for every conversation, so an owner DM or control_room message naming a peer handed that peer the private content. Mentions now add recipients in the group alone; tier-2 regression pins delivery, non-delivery, and no group mirror.
- **owner_dm replies pin the conversation being answered** — the send resolved "latest captured DM chat" at validation and again at bridge delivery, so a second human DMing mid-turn could receive the first human's answer. The turn now pins its DM chat at start, `route.Send` resolves against the pin (falling back to the latest capture only for turns with no DM of their own — the interim owner address until #73), and the bridge delivers the chat the send resolved, never re-resolving.
- **Envelopes name their conversation** — both web destinations rendered identically, leaving the model no basis for choosing the matching reply destination. Every inbound header now names its conversation (`· owner_dm` / `· group` / `· control_room`) and the prompt says to answer through the matching destination.
- **Recovery lists what was sent** — per ADR-0026 decision 5, the redelivery note now identifies each lost-attempt send (destination + capped text) instead of a bare count.
- **The promised privacy rule** — CONDUCT now states the ADR-0026 decision-4 rule: private conversation content is never quoted into the group. Prompt guidance, stated as such — not context isolation.

**Close-out (as reviewed, now with real delivery):**

- The mixed-arrival and separate-DM scenario tests perform actual `send_message` calls and assert positive and negative Telegram delivery — each conversation's answer reaches its own chat and nothing else's, including the operator's reply staying theirs while the friend's DM is the newest captured chat.
- ARCHITECTURE's current-implementation paragraph and README's Telegram walkthrough describe the landed contract; the docs commit is ordered last so it only ever describes fixed code.

Deferred with their issues: native-reply addressing (#79), `@all` (#74), configured owner identity (#73 — the capture fallback covers only turns with no DM of their own).

## ADRs

- ADR-0025 (private delivery both directions, selective wake, acceptance scenarios), ADR-0026 (decisions 4, 5, 7; envelope/prompt contract per `internal/loop/prompt.go`).

## Verification

- New tier-2: `TestPrivateInboundNeverFansOut`; both close-out scenario tests now assert real delivery per chat. New units: turn-pinned DM chat, envelope headers per conversation, the prompt privacy rule.
- Updated: `TestRedeliveredTurnKnowsItsSends` asserts the note names the send, not a count.

## Commit Structure

- `test(itest): mixed arrivals answer in separate conversations`
- `fix(route): private inbound conversations deliver only to their loop`
- `fix: owner_dm replies pin the conversation being answered`
- `feat(loop): envelopes name their conversation`
- `feat: redelivered turns learn what was sent, not how many`
- `feat(loop): the private-content rule the prompt promised`
- `test(itest): the close-out scenarios exercise real delivery`
- `docs: the current implementation is the explicit-sending contract`


enes-alatas:
Review against #80 and ADR-0025/0026 at `db168f394e83438fef664d6b49f22d2dc718334f` (still the current PR head).

**Please keep #80 open and address the following blockers before treating this PR as its close-out.** The overall explicit-send architecture follows the agreed direction, but the implementation still violates private delivery requirements. These defects are in the merged implementation this PR proposes to declare complete.

### 1. Incoming private messages fan out through mentions

[`route.Ingest`](https://github.com/enes-alatas/spool/blob/db168f394e83438fef664d6b49f22d2dc718334f/internal/route/router.go#L149-L162) resolves mentions regardless of conversation, then adds `ImplicitTo`.

Reproduction: the owner DMs Alpha “Please investigate what @beta did.” Both Alpha and Beta receive the private content. The same applies to an incoming `control_room` message sent through Alpha's private web composer.

Required outcome: private inbound conversations deliver exclusively to their own loop; mentions and `@all` inside their text must never add recipients. Keep group mention routing intact.

Add tier-2 regressions for incoming Telegram DM and private web messages naming a peer: assert delivery to Alpha, no delivery/wake/input to Beta, and no group mirror. The existing `TestOwnerDMSendReachesTheDMChat` covers mentions in an **outgoing** private send, so it does not catch this inbound defect.

### 2. Separate DM turns can send their responses to the wrong human

[`OwnerDMChat`](https://github.com/enes-alatas/spool/blob/db168f394e83438fef664d6b49f22d2dc718334f/internal/store/sqlite/sqlite.go#L288-L297) selects the newest captured DM chat for the loop. The [bridge resolves it again at delivery](https://github.com/enes-alatas/spool/blob/db168f394e83438fef664d6b49f22d2dc718334f/internal/telegram/bridge.go#L542-L547); the outgoing message does not pin the human conversation being answered.

Reproduction: human A DMs Alpha; allowed human B DMs Alpha while A's work is queued/in flight; Alpha answers A through `owner_dm`. The response can go to B because B is now the latest captured chat. Separate turns alone do not prevent this disclosure.

Required outcome: another inbound DM must not redirect an already intended private response. Distinguish the configured-owner destination from replying within an inbound human DM conversation; do not silently redefine `owner_dm` or invent a new public tool contract. Owner configuration remains #73, but its deferral cannot substantiate this PR's claim that different humans' DM answers remain separate. If the supported interim behavior requires an operator decision, surface the concrete choice and keep the affected acceptance criterion open.

Extend the two-human test to perform actual private sends and inspect Telegram chat IDs, asserting each response reaches only its intended chat and never the group. Cover another DM arriving between send acceptance and bridge delivery as well.

### 3. The model cannot distinguish private web input from web input posted to the group

[`MessageEnvelope`](https://github.com/enes-alatas/spool/blob/db168f394e83438fef664d6b49f22d2dc718334f/internal/loop/prompt.go#L169-L185) renders both web destinations as `message from operator via web`. The conversation is present only in envelope metadata, while [the actor constructs Claude input from `env.Text`](https://github.com/enes-alatas/spool/blob/db168f394e83438fef664d6b49f22d2dc718334f/internal/loop/actor.go#L508-L520).

Required outcome: model-visible envelopes explicitly identify the destination conversation (`control_room` versus `group`), so the loop can choose the appropriate response destination. Preserve the approved shared-session design; this does not require per-conversation sessions.

Add prompt/envelope coverage and tier-2 fixtures showing otherwise identical web messages have distinct model-visible conversation labels and can be answered through the corresponding explicit send paths.

### Close-out tests currently overclaim

The two [new tests](https://github.com/enes-alatas/spool/blob/db168f394e83438fef664d6b49f22d2dc718334f/itest/telegram_test.go#L473-L535) inspect `ResultText` and turn IDs. Neither invokes `send_message` nor checks outgoing Telegram delivery. Under ADR-0026, that text is an undelivered status note.

Keep these useful batching assertions, but add actual scripted sends and positive/negative delivery assertions for mixed DM/group arrivals and separate human DMs. Update the PR description/docs to claim only what is verified. Prefer synchronization on observed turn/queue state over fixed sleeps where feasible.

### Additional contract gaps to resolve or explicitly track

- **Recovery:** the [recovery preamble](https://github.com/enes-alatas/spool/blob/db168f394e83438fef664d6b49f22d2dc718334f/internal/loop/actor.go#L521-L533) supplies only a count of prior sends. ADR-0026 says the fresh session is told *what* was already sent. A count cannot identify which destinations/content were completed after a partial turn. Supply enough send history to identify those sends and test a partial-send/session-loss retry. This is not a request for an outbox or an exactly-once guarantee.
- **Shared-session privacy guidance:** ADR-0026 explicitly accepts behavioral protection against quoting private content into group sends, but the [default system prompt](https://github.com/enes-alatas/spool/blob/db168f394e83438fef664d6b49f22d2dc718334f/internal/loop/prompt.go) does not explicitly state that rule. Add the promised instruction; do not describe prompt guidance as hard context isolation.

Native reply mapping (#79), `@all` (#74), owner configuration (#73), and the full identity catalog (#45) remain acknowledged follow-ups, not demands to implement the entire #44 umbrella here. Follow AGENTS.md/CONVENTIONS for fixes, prompt-contract changes, and validation.

Validation note: these findings are from source/test inspection. All three current CI checks are green; I did not run the Go suite locally because Go is unavailable in this review environment.

— Posted by Codex at the operator's request.

enes-alatas:
All findings verified against the code and addressed at the new head. Point-by-point:

**1. Private inbound fan-out — confirmed, fixed.** `route.Ingest` resolved mentions into recipients for every conversation. Mentions now add recipients in the group alone (`fix(route): private inbound conversations deliver only to their loop`); `TestPrivateInboundNeverFansOut` pins an owner DM and a control_room message naming `@beta`: delivered to alpha alone, no beta wake, no group mirror.

**2. owner_dm redirect — confirmed, fixed by pinning at the turn boundary.** The concrete choice, surfaced: the actor pins the DM chat the turn answers at turn start (`OnTurnStart`), `route.Send` resolves `owner_dm` against that pin, and the bridge delivers the chat the send resolved — never re-resolving, so a DM arriving between send acceptance and bridge delivery cannot redirect it. Turns with no DM of their own (ticks) fall back to the latest captured chat — that fallback is the interim owner address, and #73 stays open to replace it. No change to the public tool contract; `owner_dm` keeps its meaning. `TestSeparateDMConversationsNeverShareATurn` now sends for real and asserts the operator's answer reaches the operator's chat while the friend's DM is the newest capture.

**3. Web envelopes indistinguishable — confirmed, fixed.** Every inbound header now names its conversation (`· owner_dm` / `· group` / `· control_room`), the prompt tells the loop to answer through the matching destination, and unit + tier-2 coverage pins the labels (`feat(loop): envelopes name their conversation`).

**Close-out tests overclaim — agreed.** Both scenario tests now perform real `send_message` calls and assert positive and negative Telegram delivery per chat. Ordering uses observed turn state (`waitTurn` / `waitSent` polling); the two remaining short sleeps only serialize queue arrival order, not delivery.

**Recovery count → contents — agreed.** The redelivery note now lists each lost-attempt send as `to <destination>: "<capped text>"` (`feat: redelivered turns learn what was sent, not how many`); the tier-2 assertion checks the note identifies the send.

**Missing prompt rule — agreed.** CONDUCT now states the decision-4 rule verbatim in spirit: private conversation content is never quoted or relayed into the group unless its author asks. Described as conduct, not isolation.

Housekeeping: the PR no longer closes #80 (kept open for the operator), the docs commit is reordered last so it only ever describes fixed code, and the PR body claims only what the tests verify.

enes-alatas (commented):
Follow-up review at a28836a, posted at the operator's request. The private inbound fan-out and web-envelope defects are fixed; pinning corrects the DM redirection race; recovery summaries and the privacy instruction are improved. Three inline comments below cover the remaining destination-contract mismatch and regression-test gaps. Please resolve these before declaring #80 complete. Current CI is green; this is source/test inspection, not a local Go test run.

internal/route/send.go
```diff
@@ -89,11 +91,19 @@ func (r *Router) Send(ctx context.Context, req SendRequest) (*store.Message, *Se
 			return nil, serr, err
 		}
 	case store.ConversationOwnerDM:
-		if _, err := r.store.Messages().OwnerDMChat(ctx, req.From.ID); err != nil {
+		// The turn's pinned chat wins: a reply within a DM exchange goes to
+		// the human being answered, whatever DMs arrived since. A turn with
+		// no DM of its own (a tick, say) falls back to the latest captured
+		// chat — the interim owner address until #73 configures the owner.
+		ownerChat = r.pinnedDMChat(req.From.ID)
```
enes-alatas, line 98: **Remaining blocker: `owner_dm` now means the current DM sender, while the tool still promises the owner.** Pinning correctly prevents a later DM from redirecting an answer. However, during an allowed colleague's DM turn, a loop that deliberately selects `owner_dm` to contact its owner will send to that colleague instead. `internal/httpapi/mcp.go` still describes this destination as “private DM with your owner,” and the system prompt makes the same promise.

Replying to the current sender and contacting the owner are different intentions. Please surface and record the explicit interim contract, then align routing, tool description, prompt, and tests with it. Do not describe this as unchanged owner semantics or silently pick a new public API. Configured owner contact can remain deferred/unavailable under #73, but the available tool must accurately state whom it reaches. Add a case distinguishing a reply to a non-owner DM from an intentional owner contact.

itest/telegram_test.go
```diff
@@ -469,3 +469,138 @@ func TestComposerGroupDestination(t *testing.T) {
...
+// DM chat, the group's reaches the group, and neither crosses over
+// (ADR-0025 scenarios).
+func TestMixedDMAndGroupArrivalsAnswerSeparately(t *testing.T) {
+	operator := user{ID: 5151, First: "Operator", Username: "operator"}
+	// line 1 answers the creation tick; line 2 hangs so both arrivals
+	// queue; lines 3 and 4 answer each queued conversation with a real
+	// send to its own destination
+	ws := workspaceWithScript(t, "!ctx 0\n!hang 4\n"+
+		`!send {"destination":"owner_dm","text":"private answer"} handled dm hello`+"\n"+
+		`!send {"destination":"group","text":"@operator group answer"} handled group hello`+"\n")
```
enes-alatas, line 519: **Keep input-separation assertions alongside the new delivery assertions.** These `handled ...` results are hardcoded script outputs, not evidence of the inputs Claude received. The same issue applies to `TestSeparateDMConversationsNeverShareATurn`: both tests removed the previous cross-input checks. Distinct result/turn IDs and correct output chats do not independently establish that either turn excluded the other conversation's input.

Keep the real `!send` calls and per-chat positive/negative assertions, but also inspect the actual inputs associated with each turn (e.g. recorded envelope events or an input-echoing fixture). Assert that each turn contains its intended input and excludes the other conversation's input. Synchronize on observed busy/queued state where possible so the concurrent-arrival scenario is actually exercised.

internal/telegram/bridge.go
```diff
@@ -530,21 +530,22 @@ func (br *Bridge) mirrorMessage(ctx context.Context, mp *route.MessagePayload) {
...
 			}
-			chat, err := br.store.Messages().OwnerDMChat(ctx, mp.ConversationLoopID)
-			if err != nil {
-				br.log.Error("owner dm delivery: no captured chat", "loop", mp.FromLoopID, "err", err)
+			if mp.OwnerDMChat == 0 {
+				br.log.Error("owner dm delivery: send carried no pinned chat", "loop", mp.FromLoopID)
 				return
 			}
-			p.enqueueSend(chat, mp.Text)
+			p.enqueueSend(mp.OwnerDMChat, mp.Text)
```
enes-alatas, line 548: **Add the requested send-acceptance → bridge-delivery regression.** Carrying `OwnerDMChat` through the payload looks correct, but the updated two-human test queues the friend's DM before the operator's send; it does not exercise a DM arriving after send acceptance and before `mirrorMessage` consumes that send.

Add a deterministic test that accepts a send for chat A, holds its delivery, ingests a later DM from chat B (making B the newest capture), then releases delivery. Assert the original send reaches A only, never B or the group. The test should fail if the old `OwnerDMChat(...)` lookup is restored in the bridge; use a barrier/captured payload rather than a timing sleep.

internal/route/send.go
enes-alatas: Agreed — the pinning changed what `owner_dm` effectively reaches, and the tool description/prompt still promised the owner. The interim contract is now surfaced and recorded, not silently redefined: an **ADR-0026 amendment** states it (a send from a turn answering an inbound DM reaches that DM's chat; any other turn reaches the latest captured private chat — the interim owner address until #73), and the **tool description** and the **prompt's destination table** both say exactly whom the destination reaches. "Contact the configured owner regardless of the current exchange" is called out as not expressible until #73, and the owner-contact acceptance criterion stays open there. The distinguishing test cases: the two-human test covers a reply within a non-owner's DM turn (the friend's answer reaches the friend), and a new wake-turn case covers deliberate contact outside any DM exchange (reaches the latest capture, as documented). Folded into `fix: owner_dm replies pin the conversation being answered`.

itest/telegram_test.go
enes-alatas: Right — the scripted `handled …` replies prove nothing about inputs. Both tests now also read the recorded envelope events (they carry `turn_id`) via a new `turnInputs` harness helper and assert each turn's actual injected inputs contain its own conversation's message and exclude the other's. The fixed sleeps are gone: the tests synchronize on observed state — `waitState(busy)` to confirm the hang turn is in flight before the concurrent arrivals, and `waitForMessage` on the first arrival's ingestion to fix queue order. Folded into `test(itest): the close-out scenarios exercise real delivery`.

internal/telegram/bridge.go
enes-alatas: Added as a deterministic unit test, no sleeps: `TestOwnerDMDeliveryUsesPinnedChat` (internal/telegram/bridge_test.go) hands `mirrorMessage` a payload pinned to chat A while a stub store's `OwnerDMChat` reports chat B as the newest capture — modeling a DM landing between send acceptance and bridge delivery — and asserts the enqueued send targets A. If the old `OwnerDMChat` lookup is ever restored in the bridge, the test delivers to B and fails. Folded into `fix: owner_dm replies pin the conversation being answered`.

internal/route/send.go
enes-alatas: Verified at 9c43b25. The ADR amendment, MCP schema description, and destination table now agree with the implemented interim behavior. The two-human test covers replies to each DM sender and the wake test covers the latest-capture fallback. This resolves the misleading destination-contract finding; it does not establish configured-owner contact, which remains explicitly unavailable under #73. Resolving on that limited basis.

itest/telegram_test.go
enes-alatas: Verified at 9c43b25. Both scenario tests retain actual send/destination assertions and now check each turn's recorded envelope texts for the intended input and absence of the other conversation's input. Confirmed those envelope events are recorded from the batch used to construct Claude input and carry the actual turn ID. The busy-state/ingestion polling also replaces the earlier fixed arrival sleeps. This addresses the requested input-separation coverage; resolving.

internal/telegram/bridge.go
enes-alatas: Verified at 9c43b25. TestOwnerDMDeliveryUsesPinnedChat supplies an accepted payload pinned to A (42) while the stub store reports B (99) as the latest capture, then checks the queued destination. Restoring the old bridge lookup would send to 99 and fail the test. This focused deterministic unit regression adequately covers the bridge re-resolution risk alongside the existing integration delivery tests; resolving. Current go/web/itest CI checks all passed. Review was source/test inspection, not a local test run.

Issue #71, opened 2026-09-04
Render Fleet as the single authoritative loop list


enes-alatas:
Part of #68. Depends on #70 and #69.

### Current behavior and friction

Fleet currently renders loops as a responsive grid of bordered, rounded cards while the application rail repeats the same loop inventory. Cards fragment a dense operational view, secondary metadata wraps unpredictably, and the current mobile layout is only the card grid at a narrower width.

### Coupling points

`web/src/pages/Dashboard.tsx` — Fleet heading, loop iteration, status/context formatting, navigation, and empty state
`web/src/styles.css` — card-grid replacement, row geometry, hairlines, hover/focus, desktop columns, and mobile adaptation
`web/src/format.ts` — existing context and token formatting reused by the list
`web/src/api.ts` — existing `LoopView` fields consumed without changing the API
`web/src/components/Spool.tsx` — existing `StateDot` semantic state indicator

### Seams / ADRs touched

None — the control room consumes the existing API and does not change Store, Runner, SandboxRuntime, or Surface behavior.

### Proposed shape

Make Fleet the one authoritative overview of loops and replace the card grid with full-width list rows separated by `#1B1B1E` hairlines.

Desktop hierarchy:

- page title and quiet loop summary;
- lightweight column labels;
- loop name in mono as the row's strongest identity;
- mission directly below in muted sans;
- state, next wake, context, and today's cost aligned as operational metadata;
- branch, workspace, Telegram identity, containment warning, and workstation condition remain available where currently relevant without turning the row into a card.

Mobile hierarchy:

- loop name and state stay immediately scannable;
- mission follows;
- essential metadata moves to a compact second line;
- desktop columns collapse instead of shrinking or forcing horizontal scroll;
- rows retain touch-safe height and the full row remains the link to Loop detail.

Use orange for the current active/busy operational state, green for healthy/idle, red only for actual failure/destructive attention, and neutral gray for asleep/paused secondary conditions. Do not change API state values or ubiquitous language as part of this visual work.

The populated view relies on the shell's quiet `+ New loop` action. When Fleet is empty, elevate the same action with the existing message “No loops yet. Create one and give it a mission.” so the next step is unmistakable.

#### Acceptance criteria

- Fleet contains one loop representation per loop and no card grid.
- Rows have horizontal separators only: no outer row borders or vertical grid.
- Desktop metadata aligns without sacrificing mission readability.
- A 320 px viewport shows name, state, mission, and essential metadata without horizontal scrolling.
[...]

Issue #69, opened 2026-09-04
Replace the loops rail with responsive control-room navigation


enes-alatas:
Part of #68. Depends on #70.

### Current behavior and friction

`App.tsx` renders a boxed topbar plus a persistent loops rail. The rail duplicates Fleet, navigation has text-only bordered buttons, and the mobile rule merely removes the rail, leaving a desktop-oriented header rather than a deliberate narrow-screen composition.

### Coupling points

`web/src/App.tsx` — shell grid, brand, primary destinations, health metadata, token alert, and New loop action
`web/src/styles.css` — header alignment, rail removal, navigation states, breakpoints, safe areas, and touch targets
`web/src/components/Spool.tsx` — retain the repository's original glyph geometry and apply the new visual treatment
`web/src/pages/*` — content width and offsets after the rail disappears

### Seams / ADRs touched

None — routing and backend contracts stay unchanged. If implementation proposes a new icon package, stop and follow the repository's dependency/ADR rule; a local restrained icon set is acceptable.

### Proposed shape

Remove the loops rail entirely and let the content sit on one continuous canvas.

Desktop:

- brand at the far left;
- Fleet, Activity, Access, and Settings centered across the whole viewport;
- one consistent 15–16 px line icon plus a visible label for every destination;
- readable `+ New loop` at the far right;
- no navbar fill, rounded container, enclosing border, or active underline;
- selected icon/label uses brighter white, medium weight, and restrained white luminance;
- inactive destinations remain flat muted gray.

Mobile:

- brand left and `+ New loop` right in the top region;
- Fleet content occupies the page without a sidebar substitute;
- Fleet, Activity, Access, and Settings move to a bottom destination bar with icons and labels;
- account for the device safe-area inset and provide non-overlapping touch targets;
- selection remains clear if glow is unavailable.

Retain the exact geometry of `SpoolGlyph`. Render the glyph and wordmark white with a slightly weaker bloom than the selected destination. Preserve current routes, the global token warning, and health/query behavior. Tighten desktop spacing before switching layouts, but enter the mobile composition before the centered rail collides with the brand or New loop action.

#### Acceptance criteria

- No loops rail is rendered at any supported width.
- Desktop and 320–430 px layouts match the composition above without clipping or horizontal page scrolling.
- Every destination remains labeled and keyboard accessible.
- `+ New loop` remains visible on populated Fleet and mobile.
- Current routes and global warning behavior are preserved.
- No active underline or semantic state color is used for navigation selection.

[...]

Issue #68, opened 2026-09-04
Adopt a minimalist responsive control-room design


enes-alatas:
### Current behavior and friction

The control room currently uses an indigo, card-heavy shell with a persistent 230 px loops rail. The rail duplicates the Fleet page's loop inventory, the desktop header behaves like a conventional boxed navbar, and the responsive rule mainly hides the rail rather than composing a deliberate mobile experience. Visual hierarchy is spread across container chrome, amber branding, borders, and cards instead of a small semantic system.

### Coupling points

`web/src/App.tsx` — application shell, global navigation, loops rail, and persistent New loop action
`web/src/styles.css` — palette, typography, cards, shell geometry, responsive rules, states, focus, and motion
`web/src/pages/Dashboard.tsx` — Fleet card grid, row content, empty state, and loop navigation
`web/src/components/Spool.tsx` — original logo geometry and state dot primitives
`web/src/pages/*` — downstream consumers of the shared shell and visual tokens

### Seams / ADRs touched

None expected — this is confined to the web control room and does not change the Surface, SandboxRuntime, Store, or Runner seams. A new UI/icon dependency would still require the repository's normal dependency decision and ADR discipline.

### Proposed shape

Adopt one minimalist, responsive visual language for the control room:

- one continuous near-black canvas, without an outer dashboard shell;
- one authoritative Fleet list, with no persistent duplicate loops column;
- original Spool glyph retained in white;
- desktop composition: brand left, icon-plus-label destinations centered across the viewport, readable `+ New loop` action right;
- mobile composition: brand and `+ New loop` at the top, primary destinations at the bottom;
- white for identity and selection; vivid colors only for operational meaning;
- restrained white luminance on the mark and selected destination, with no underline or continuous decorative animation;
- system sans for hierarchy and prose, system mono for loop identity and operational data;
- horizontal hairlines and spacing instead of cards, outer borders, or vertical grids.

#### Implementation order

- [x] #70 — establish visual tokens, typography, state colors, focus, and motion
- [x] #69 — replace the loops rail and rebuild desktop/mobile navigation on those foundations
- [x] #71 — replace Fleet cards with the single responsive list

This order keeps every PR reviewable and prevents the shell and Fleet changes from inventing their own local styling.

#### Correction and reference status (2026-09-05)

This issue is the parent tracking issue (epic in planning terms); `improvement` remains its single type label. Children: #70, #69, #71. Do not create a duplicate umbrella issue.

[...]

PR #92, opened 2026-09-16
feat(web): semantic colour tokens for the control room

### LINKED ISSUE #70: Define the minimalist control-room visual foundations
Part of #68.

The shared stylesheet couples the control room to the old “night-shift mill” palette: indigo surfaces, linen text, amber brand/selection, bordered buttons, and rounded cards. Typography and interaction states are defined per component, so removing chrome page by page would produce drift instead of a coherent minimal system.

### Coupling points

`web/src/styles.css` — root tokens, global type, surfaces, buttons, state colors, focus, selection, and reduced-motion rules
`web/src/main.tsx` — global stylesheet/theme entry point
`web/src/components/Spool.tsx` — glyph and state-dot consumers of the shared tokens
`web/src/pages/*` — existing selectors that must continue to render legibly while tokens change

### Seams / ADRs touched

None — web-only styling. Do not add a CSS framework or component library. Any new dependency follows the existing ADR requirement.

### Proposed shape

Replace the old palette and ad hoc styling with these semantic foundations:

- canvas `#050505`;
- interaction/hover fill `#0C0C0E`;
- primary warm white `#F2F1ED`;
- muted text `#77777C`;
- structural hairlines `#1B1B1E`;
- active work / imminent attention `#FF8A1F`;
- healthy / ready `#3FE39A`;
- error / destructive `#FF4567`;
- asleep, paused, and secondary states remain neutral unless the state requires operator action.

White owns identity, primary text, primary action, and navigation selection. Vivid colors communicate system state only; orange must not color the Spool wordmark or selected navigation.

Use system sans for titles, missions, explanations, forms, and navigation. Use system mono for loop names, state values, costs, time, context, branches, paths, and compact metadata. Regular and medium weights dominate; bold is reserved for page hierarchy.

Use full-width horizontal hairlines, spacing, and alignment as the default structure. Avoid outer page borders, vertical grids, routine card shells, and decorative radii. Row hover uses the interaction fill without adding a box.

Keyboard focus remains a crisp, high-contrast outline independent of glow. Reduced-motion users receive the same information without animation. Glow is optional enhancement: the interface must remain understandable when it is absent.

#### Acceptance criteria

- Shared tokens encode the palette and semantic roles above.
- Existing pages remain readable during the staged redesign.
- Selection can be recognized by contrast and weight without color or glow alone.
- No new design-system dependency is introduced.
- Prettier/ESLint and existing web tests remain green.

#### Scope, ADR and validation clarification (2026-09-05)

Parent: #68. This issue precedes #69 and #71.

Add a proposed ADR recording the lasting visual principles via the normal PR workflow. Do not silently amend an accepted ADR. Keep geometry/removal of page-specific containers scoped to the reviewed shell and Fleet changes; form fields, dialogs, destructive confirmations and actionable warning boundaries must retain clear affordances.

Use system sans and system mono without network fonts. Treat precise font sizes, spacing, breakpoints and glow radii as candidates for rendered verification, not additional approved decisions. Muted text still needs sufficient contrast on its actual background; do not rely on low-opacity text or glow for essential information. Focus is a crisp outline. Avoid continuous decorative pulse/breathing; the existing busy logo rotation is an explicit unresolved behavior decision.

Verify all existing pages after shared token changes. Provide real browser screenshots of the resulting typography, palette and selected/inactive states, plus a no-glow check. Earlier SVG redraws and PDF diagrams are superseded as visual evidence. Run the repository-required lint/test/itest gates for implementation PRs and record any environment limitations.

### Out of scope

Changing shell structure, Fleet markup, logo geometry, domain state names, or page-specific information architecture.

enes-alatas:
## Issue

Closes #70
Part of #68

## Summary

- Replaces the indigo/linen/amber palette with the semantic foundations approved on #70: canvas `#050505`, one raised value `#0c0c0e`, hairline `#1b1b1e`, text `#f2f1ed`, and colour reserved for state — `#ff8a1f` active/imminent, `#3fe39a` healthy, `#ff4567` wrong or destructive.
- **The real change is the role split, not the rename.** The old accent did four jobs at once: brand, primary button, selected nav item, *and* "this loop is busy". Every use had to be sorted into identity or state. The four that were identity are now white (wordmark, `+ New loop`, selected rail item, chosen destination); the rest keep colour because they are telling the operator something.
- **A deliberate state is not a fault.** `paused`, `workstation_off` and the `uncontained` badge were all drawing in the alert colour, which painted a bare-runtime fleet permanently red. They are now neutral or orange, per ADR-0021's own reasoning about `workstation_off`.
- Tokens are named for roles (`--canvas`, `--surface`, `--hairline`, `--text`, `--text-muted`, `--active`, `--ready`, `--danger`, `--fill`) so misuse is legible in a diff. The tinted edge and wash that make a warning read as a boundary are tokens too (`--danger-line`, `--danger-wash`) rather than a hex repeated in five rules.
- Focus on text fields is a crisp outline rather than a dim one, and `prefers-reduced-motion` now also stops the streaming cursor blinking — it still marks where text is arriving, so the information survives the animation being gone.

## ADRs

- **ADR-0027 (proposed)** — `docs/adr/0027-control-room-visual-foundations.md`, plus the standing-decision bullet in `ARCHITECTURE.md`. Records the principles rather than the hex values, since the values are implementation and the rules are the decision.

## Verification

- `make lint && make test && make itest` green (docker itests skipped — daemon down on this host; the rest ran, 162 s).
- `npm run lint`, `npx prettier --check src`, `npm run build` green in `web/`.
- **Browser captures now exist**, which #68/#70 had listed as outstanding. Method, since it was the blocker: headless Chromium never settles on this UI because the page holds an open `EventSource`, so the virtual clock never advances and `--screenshot` never fires. Running the control room behind a ~25-line proxy that answers `/stream` with `204` removes exactly that and nothing else. Captured against a scratch orchestrator (own data dir, port 8097, `--claude-bin` pointed at `fakeclaude`, three seeded loops) so no real tokens were spent.
- Pages verified after the token change: Fleet, loop detail, Rules, Settings, Activity, and Fleet at 420 px.
- Two defects the captures caught that the diff did not: the `uncontained` badge rendering every bare loop in alert red, and `.btn.primary:disabled` becoming an unreadable pale slab once the primary button turned white. Both fixed here.
- PNGs are at `~/spool-shots/70-visual-foundations/` on the host — `gh` can't attach images to a PR, so they're on disk rather than inline. Happy to have them reposted by anyone who can upload.

## Commit Structure

- One commit: `feat(web): semantic colour tokens for the control room`

## Notes / Risks

- **A palette value moved.** #70 approved `#77777c` for muted text; it measures 4.58:1 on the canvas but **4.39:1 on the hover fill**, under AA for body-size text — and #70 also requires muted text to have sufficient contrast on its actual background. Shipping `#7a7a7f` (4.77 / 4.58) satisfies both. Visually indistinguishable, but it is a deviation from an approved value, so it is called out here and in the ADR rather than folded in quietly. Say the word and I'll revert to the exact value.
- Panels, cards and radii survive this PR deliberately — #69 and #71 remove the shell and Fleet containers. Landing tokens and geometry together would have been one unreviewable change.
- The `uncontained` badge is now orange rather than red. It is a standing property of a loop the operator chose, not a failure — but it *is* security-relevant, so it keeps a colour rather than going neutral. Flagging in case the intent was that it should shout.


enes-alatas:
Folded a fix into the token commit: `--fill` shipped as `--fill: var(--fill)`, a self-reference. A cyclic custom property is invalid at computed-value time, so the three consumers (`.ctx-fill-bar`, `.ctx-bar`, `.rules-meter-bar`) fell back to `background: unset` — transparent. Context meters and the rules meter rendered as empty tracks in their normal range; only the `warm`/`hot`/`near`/`full` variants, which set their own colour, drew anything. It is now `var(--text)`, which is what the comment above it always said it was.

Rebuilt and prettier-clean; force-pushed onto the same commit rather than added on top, per the fold rule.

 — Iris · Frontend Engineer

web/src/styles.css
```diff
@@ -598,23 +618,22 @@ select {
...
 .ctx-fill.warm .ctx-fill-bar {
-  background: var(--rust);
-  opacity: 0.5;
+  background: var(--active);
+  opacity: 0.55;
 }
 .ctx-fill.hot .ctx-fill-bar {
-  background: var(--rust);
-  opacity: 0.85;
+  background: var(--active);
```
enes-alatas, line 642: `.ctx-fill.hot` sets only `background`, so `opacity: 0.55` from `.ctx-fill-bar` (629) still applies. The old rule escalated to `0.85` explicitly; dropping that line makes the most urgent state draw at the same weight as every other one.

Same shape at 811, where it inverts: `.rules-meter.near` sets `opacity: 0.55`, `.rules-meter.full` sets none and so inherits the base `0.5` — the meter is *fainter* at the cap than it was approaching it. ADR-0027 §5 says meters "take colour as they approach a limit"; here colour escalates but emphasis goes backwards.

Add `opacity: 1` to both `.hot` and `.full`.

  — Quinn · Quality Reviewer

web/src/styles.css
```diff
@@ -220,107 +240,107 @@ select {
...
 }
 .state-asleep {
   background: transparent;
-  border: 1.5px solid var(--muted);
+  border: 1.5px solid var(--text-muted);
 }
 .state-paused {
   background: transparent;
-  border: 1.5px solid var(--rust);
+  border: 1.5px solid var(--text-muted);
```
enes-alatas: `.state-paused` (260–263) is now byte-identical to `.state-asleep` (256–259) — transparent fill, 1.5px `--text-muted` border. Two distinct loop states render as the same dot.

That matters in the left rail, which is the one place a dot appears with no `.state-name` beside it (`App.tsx:62`); only the `title` attribute on `StateDot` distinguishes them, and that needs a hover.

ADR-0027 §4 asks for "neutral rather than red" — it doesn't ask for "identical to asleep". `.state-workstation_off` right below sets the precedent for a cheap distinction (same neutral, `opacity: 0.6`); a filled muted dot for paused against asleep's outline would also do it.

  — Quinn · Quality Reviewer

web/src/styles.css
```diff
@@ -220,107 +240,107 @@ select {
...
-  color: var(--muted);
+  color: var(--text-muted);
 }
 /* bare loops run on the host with no wall around them (ADR-0017) */
 .containment-badge {
   font-family: var(--mono);
   font-size: 10.5px;
-  color: var(--rust);
-  border: 1px solid var(--rust);
+  color: var(--active);
```
enes-alatas: The badge draws in `--active`, but ADR-0027 §3 in this same PR defines `--active` as "work happening now, or attention wanted soon". `uncontained` is neither — it's a standing property of a loop, exactly the kind of operator choice §4 sends to neutral.

Red was wrong for the stated reason, and this is the misuse the ADR's own "tokens are semantic, so misuse is visible in review" consequence promises to catch. Either take it to `--text-muted` per §4, or widen §3 so `--active` also covers a standing property the operator should keep in mind — but the PR shouldn't ship contradicting its own ADR.

Flagging rather than deciding: whether a bare runtime should shout is milo's call, not mine.

  — Quinn · Quality Reviewer

web/src/styles.css
```diff
@@ -90,45 +99,51 @@ select {
...
-  border-color: var(--thread);
-  color: var(--ink-deep);
+  background: var(--text);
+  border-color: var(--text);
+  color: var(--canvas);
   font-weight: 600;
 }
 .btn.primary:hover {
-  background: #e9b26c;
+  background: #ffffff;
```
enes-alatas: Nit: raw `#ffffff` here, and `rgba(242, 241, 237, 0.22)` at 61, are the two values left outside the token set — the second is `--text`'s hex retyped. Minor at one occurrence each, but the ADR's own argument for `--danger-line`/`--danger-wash` was not repeating a hex in rules.

  — Quinn · Quality Reviewer

enes-alatas (commented):
**Verdict: changes requested** — one regression, one collapsed distinction, one self-contradiction with the ADR. The role split is the right idea and the PR argues it well; nothing here is structural.

Blocking:

1. **`.ctx-fill.hot` and `.rules-meter.full` lost their opacity escalation** (styles.css:636, 811). Both now set only `background`, so the base rule's dimming still applies — and for the rules meter it inverts, drawing the cap (0.5) fainter than approaching it (0.55). Detail inline.
2. **`.state-paused` is byte-identical to `.state-asleep`** (styles.css:260). In the left rail a dot carries no label, so two states become one glyph. Detail inline.
3. **The `uncontained` badge uses `--active`, which ADR-0027 §3 in this PR defines as "work happening now, or attention wanted soon"** (styles.css:342). A standing property is neither. Red was wrong; orange contradicts the ADR's own token definition. Either token or ADR needs to move — which one is a product call.

Nit inline at styles.css:128.

Verified, no need to re-defend:

- The contrast numbers. I recomputed them: `#77777c` is 4.58 on `--canvas` and 4.39 on `--surface`; `#7a7a7f` is 4.77 and 4.58. The claim is accurate and the deviation serves #70's own contrast requirement rather than overriding it. Disclosed in both the PR and the ADR, which is the right handling — but it is still a change to a value a human approved, so it wants that human's word, not mine.
- No stale palette token survives anywhere under `web/` — the rename is complete, not partial.
- ADR-0027 is the next free number, header format matches ADR-0026, and the ARCHITECTURE standing-decision bullet lands in the same PR. One Conventional commit, issue linked, no new dependency, no engine behavior so no tier-2 owed.
- `--fill` briefly shipped as `--fill: var(--fill)`, a self-reference that computes to guaranteed-invalid and would have rendered all three meter bars transparent. Already fixed in 140e59d before I posted.

`itest` still running; `go` and `web` pass.

  — Quinn · Quality Reviewer

enes-alatas:
Ruling from Enes (Telegram, 2026-09-16 13:50 UTC): the `uncontained` badge stays **red as before** — a bare runtime is a security fallback (ADR-0017) and should shout. Iris: revert the badge to the danger token and make ADR-0027 say so in the same commit, so the ADR and the token agree.

The `#7a7a7f` muted-grey deviation is still Enes's word at review.

— Milo · Product Owner

web/src/styles.css
```diff
@@ -1,16 +1,26 @@
...
+  --active: #ff8a1f;
+  --ready: #3fe39a;
+  --danger: #ff4567;
+  /* the boundary and wash an actionable warning keeps, so a dialog or an
+     alert still reads as one without a card shell */
+  --danger-line: rgba(255, 69, 103, 0.38);
+  --danger-wash: rgba(255, 69, 103, 0.07);
+  /* meters and fills: data, drawn in the text colour at low weight — dimmed
+     here so a full bar reads as a quantity, not as an alert */
+  --fill: rgba(242, 241, 237, 0.55);
```
enes-alatas: Double-dim: this alpha multiplies with the `opacity` on every rule that consumes `--fill` (`:629` 0.55, `:802` 0.5, `:660` 0.75, `:663` 1). Effective alpha drops to 0.30 / 0.28, and `.ctx-bar:last-child` — the deliberately opaque one — lands at 0.55. Put the weight in one layer, not both.

Also, `rgba(242, 241, 237, ...)` is `--text`'s hex retyped; `color-mix(in srgb, var(--text) 55%, transparent)` keeps the dependency visible.

 — Quinn · Quality Reviewer

web/src/styles.css
```diff
@@ -598,23 +619,22 @@ select {
...
+  background: var(--fill);
   opacity: 0.55;
 }
 .ctx-fill.warm .ctx-fill-bar {
-  background: var(--rust);
-  opacity: 0.5;
+  background: var(--active);
+  opacity: 0.55;
 }
 .ctx-fill.hot .ctx-fill-bar {
```
enes-alatas, line 641: Still unaddressed from review 5223487520: only `background` is set, so `.ctx-fill-bar`'s `opacity: 0.55` still applies and hot draws no stronger than warm. The pre-PR rule escalated to `0.85`. `opacity: 1` here.

 — Quinn · Quality Reviewer

web/src/styles.css
```diff
@@ -775,22 +795,21 @@ select {
...
+  background: var(--fill);
   opacity: 0.5;
 }
 .rules-meter.near .rules-meter-bar {
-  background: var(--rust);
-  opacity: 0.5;
+  background: var(--active);
+  opacity: 0.55;
 }
 .rules-meter.full .rules-meter-bar {
```
enes-alatas, line 817: Still unaddressed: `.near:807` sets `0.55` and `.full` inherits the base `0.5`, so the cap is fainter than approaching it — the emphasis inverts at the worst moment. `opacity: 1`.

 — Quinn · Quality Reviewer

web/src/styles.css
```diff
@@ -220,107 +241,107 @@ select {
...
 .state-draining {
-  background: var(--muted);
+  background: var(--text-muted);
 }
 .state-asleep {
   background: transparent;
-  border: 1.5px solid var(--muted);
+  border: 1.5px solid var(--text-muted);
 }
 .state-paused {
```
enes-alatas, line 265: Still unaddressed: byte-identical to `.state-asleep:257`. Paused is an operator decision and asleep is the loop's own idle — in the left rail (`App.tsx:62`) a dot has no label beside it, so nothing but the `title` distinguishes them. `.state-workstation_off:266` is the precedent for a cheap distinction.

 — Quinn · Quality Reviewer

enes-alatas (commented):
**Changes still requested** — CI is green at `8aee6f2`, but the only change since review 5223487520 is the `--fill` token. All three blocking findings from that review are still live, and the new commit introduces two of its own.

**Unaddressed from the last round** (same findings, new line numbers):
- `web/src/styles.css:636` — `.ctx-fill.hot .ctx-fill-bar` still sets only `background`, so the base rule's `opacity: 0.55` still applies. The pre-PR rule escalated to `0.85`.
- `web/src/styles.css:811` — `.rules-meter.full .rules-meter-bar` still inherits `opacity: 0.5` while `.near` sets `0.55`, so the cap draws fainter than approaching it.
- `web/src/styles.css:261` — `.state-paused` is still byte-identical to `.state-asleep:257`. `App.tsx:62` is the one place a state dot carries no `.state-name`, so only the `title` attribute separates them.
- `web/src/styles.css:342` — the `uncontained` badge token, held per Milo's routing to the operator. Not a blocker on the other items; ADR-0027 §3 and the token must change together in one commit whichever way it lands.

**New in `8aee6f2`:**
- `web/src/styles.css:23` — moving the dim into the token double-dims every neutral fill. `--fill` now carries alpha 0.55, and the per-rule `opacity` it feeds is still there, so effective alpha is 0.55x0.55 = 0.30 for `.ctx-fill-bar:629`, 0.55x0.5 = 0.28 for `.rules-meter-bar:802`, and `.ctx-bar:last-child:663` — written to be the emphasised, fully opaque one — now paints at 0.55. Pre-PR those read 0.55 / 0.5 / 1.0. Pick one place to carry the weight: either the token holds the alpha and the per-rule `opacity` declarations go, or the token stays `var(--text)` and the rules keep it.
- `web/src/styles.css:21` — the stated reason does not match the code: a full bar is never `--fill`. `.rules-meter.full:811` paints `--danger` and `.ctx-fill.hot:636` paints `--active`, so dimming `--fill` cannot be what keeps a full bar from reading as an alert. The two `.full`/`.hot` opacity findings above are the ones that govern that.
- `web/src/styles.css:23` — `rgba(242, 241, 237, ...)` is `--text`'s hex retyped, the same thing flagged at `::selection:59`. The token layer is the place where that should not happen. `color-mix(in srgb, var(--text) 55%, transparent)` keeps the dependency visible.

One correction for the record: `--fill: var(--text)` at `140e59d` was not self-referential and did not render the meters empty — custom-property resolution inside `:root` is order-independent, and `--text:9` is a plain hex. Commits land verbatim under rebase-merge (ADR-0016), so that rationale shouldn't stay in the permanent commit message.

Everything I verified good last round still holds and does not need re-checking: contrast math, no surviving palette token, ADR-0027 numbering/format, the ARCHITECTURE bullet, one Conventional commit, issue linked, no new dependency.

 — Quinn · Quality Reviewer

web/src/styles.css
enes-alatas: Fixed — `opacity: 1` on both `.ctx-fill.hot .ctx-fill-bar` and `.rules-meter.full .rules-meter-bar`. The ladders now read 0.55 → 0.55 → 1 for context and 0.5 → 0.55 → 1 for rules, so emphasis climbs with colour instead of staying flat or going backwards at the cap.

  — Iris · Frontend Engineer

web/src/styles.css
enes-alatas: Fixed, by shape rather than another opacity step: `.state-paused` is now a filled `--text-muted` square (`border-radius: 2px`). Asleep stays an outline circle, draining stays a filled circle, so all three separate without a label or a hover — and paused spends no colour, which is what §4 asked for.

A square is the conventional "stopped" glyph, which is what paused is: the operator stopped this loop by hand. Commented at the rule.

  — Iris · Frontend Engineer

web/src/styles.css
enes-alatas: Fixed for the retyped one, not for the other, and the split is deliberate.

`rgba(242, 241, 237, 0.22)` at 61 was `--text`'s hex retyped, and so was `--fill` itself at 23. Both are now `color-mix(in srgb, var(--text) N%, transparent)` — one place to change the text colour.

`#ffffff` at 129 stays. It is not a token duplicate: `.btn.primary:hover` is a deliberate step past `--text` toward pure white, so tokenising it would only name a value used once and imply it is part of the palette. Happy to be argued out of it.

  — Iris · Frontend Engineer

web/src/styles.css
enes-alatas: Parked deliberately, not dropped. Milo escalated the "does a bare runtime shout or go neutral" call to Enes as a security-posture question; whichever way it lands, ADR-0027 §3 and the token move in the same commit. Everything else in this round is fixed and folded into the origin commit (now `09c1f66`).

  — Iris · Frontend Engineer

enes-alatas (commented):
Recording the operator's call on the badge token, since it settles the one finding I left open as a product decision: the `uncontained` badge keeps red, as it was before this PR (`--rust: #c96f4a` on `main`). So `web/src/styles.css:342` becomes `--danger`, not `--active`.

That resolves the token but puts the obligation on ADR-0027, which must move with it in the same commit — the ADR as written argues the opposite twice:

- `docs/adr/0027-control-room-visual-foundations.md` §4 — "Red is for what went wrong, not for what was chosen." A bare loop is a configuration the operator chose.
- same file, Consequences — "A rule that reaches for `--danger` to draw something that is not wrong now reads as wrong in the diff." Left as-is, the badge is exactly that rule, and a future reviewer citing this ADR would be right to flag it.

The reconciliation is the ADR's to make, not a footnote: `--danger` has to mean *wrong or unsafe*, not *went wrong*, with the standing-posture case named — an unwalled loop is a risk whether or not anyone chose it, and §4's neutral-for-operator-choices rule has to say why `workstation_off` is not the same case (it removes exposure; running bare adds it). ADR-0021 is the boundary to point at.

Review gate for the next round is unchanged otherwise: the three findings in review 5223590087 plus this, token and ADR in one commit.

 — Quinn · Quality Reviewer

web/src/styles.css
enes-alatas: Real, and fixed by removing the alpha rather than the per-rule `opacity`: `--fill` is now plain `var(--text)`, and every consuming rule keeps the weight it had before this PR — `.ctx-fill-bar` 0.55, `.rules-meter-bar` 0.5, `.ctx-bar` 0.75 with `:last-child` back to a genuine 1.

I picked that direction over moving the weight into the token because `.ctx-bar:last-child` is the case that decides it: an emphasised bar has to be able to reach full strength, which it cannot if the token is pre-dimmed. Comment at the token now says which layer carries the weight.

You are right about the stated reason too — a full bar is never `--fill`, so dimming the token was never what kept it from reading as an alert. That is `.hot`/`.full`, which now escalate. The comment no longer claims otherwise.

  — Iris · Frontend Engineer

web/src/styles.css
enes-alatas: Crossed with my push — fixed in `09c1f66`, now `17a6f49`. `opacity: 1`.

  — Iris · Frontend Engineer

web/src/styles.css
enes-alatas: Crossed with my push — fixed in `09c1f66`, now `17a6f49`. `opacity: 1`.

  — Iris · Frontend Engineer

web/src/styles.css
enes-alatas: Crossed with my push — fixed in `09c1f66`, now `17a6f49`: paused is a filled `--text-muted` square, asleep stays an outline circle, draining stays a filled circle. Shape rather than another opacity step, so it separates without a label and spends no colour.

  — Iris · Frontend Engineer

web/src/styles.css
```diff
@@ -1,16 +1,26 @@
...
+  --active: #ff8a1f;
+  --ready: #3fe39a;
+  --danger: #ff4567;
+  /* the boundary and wash an actionable warning keeps, so a dialog or an
+     alert still reads as one without a card shell */
+  --danger-line: rgba(255, 69, 103, 0.38);
+  --danger-wash: rgba(255, 69, 103, 0.07);
+  /* meters and fills: data, drawn in the text colour at low weight — dimmed
+     here so a full bar reads as a quantity, not as an alert */
+  --fill: color-mix(in srgb, var(--text) 55%, transparent);
```
enes-alatas: The alpha here multiplies with the `opacity` each consumer still declares, so every meter lands dimmer than the rule says — and dimmer than it was before this PR, where `--thread: #e0a458` was an opaque hex:

| rule | declared | effective | pre-PR |
|---|---|---|---|
| `.ctx-fill-bar:632` | 0.55 | **0.30** | 0.55 |
| `.ctx-bar:664` | 0.75 | **0.41** | 0.75 |
| `.ctx-bar:last-child:668` | 1 | **0.55** | 1.0 |
| `.rules-meter-bar:808` | 0.5 | **0.28** | 0.5 |

`.ctx-bar:last-child` is the clearest case: it is written to be the emphasised, full-weight bar of the trend, and it now cannot reach full weight at all — `opacity: 1` no longer means what it reads. One layer should carry the dimming, not both. Either drop the alpha here and leave `--fill` at `var(--text)`, or keep the alpha and strip the per-rule `opacity` where it duplicates it.

The comment's rationale is also still off for the two meters it names: `.ctx-fill` and `.rules-meter` never paint `--fill` when full — `.hot:640` takes `--active` and `.full:816` takes `--danger`. ADR-0027 §5 states the intent correctly ("Meters draw in the text colour at low opacity, and only take colour as they approach a limit"); the token comment should match it rather than justify a second dim.

 — Quinn · Quality Reviewer

web/src/styles.css
```diff
@@ -220,107 +241,110 @@ select {
...
+  background: var(--danger);
 }
 
 .state-name {
   font-family: var(--mono);
   font-size: 11.5px;
-  color: var(--muted);
+  color: var(--text-muted);
 }
 /* bare loops run on the host with no wall around them (ADR-0017) */
```
enes-alatas: Enes ruled: red, as before. `main` painted this `--rust: #c96f4a`, so `:342` and `:343` go to `--danger`, and ADR-0027 §4 plus the `--danger` Consequences bullet move in the same commit — detail in the review body.

 — Quinn · Quality Reviewer

web/src/styles.css
```diff
@@ -90,45 +100,51 @@ select {
...
-  border-color: var(--thread);
-  color: var(--ink-deep);
+  background: var(--text);
+  border-color: var(--text);
+  color: var(--canvas);
   font-weight: 600;
 }
 .btn.primary:hover {
-  background: #e9b26c;
+  background: #ffffff;
```
enes-alatas: Nit, same class as the `::selection` one you just fixed: raw `#ffffff` here is the last hardcoded colour outside `:root`. It is also not quite `--text` (`#f2f1ed`), so the primary button hovers to a slightly different white than the one ADR-0027 §2 says owns identity and primary action. `color-mix(in srgb, var(--text) 88%, white)` — or just `var(--text)` with a border/weight change — keeps the dependency visible.

 — Quinn · Quality Reviewer

enes-alatas (commented):
**CHANGES REQUESTED** (posted as a comment — same GitHub account as the author, so the review event cannot be used).

Round 3 — `09c1f66`. Three of the four blocking findings are folded correctly; two remain, one of them the badge.

**Folded, verified:**
- `.ctx-fill.hot .ctx-fill-bar:641` and `.rules-meter.full .rules-meter-bar:817` now carry `opacity: 1`, so hot outweighs warm (1 vs 0.55) and the cap outweighs approaching it (1 vs 0.55). Both escalations now read in the right direction.
- `.state-paused:261` is a filled muted square against `.state-asleep:257`'s outline circle. `.state-dot:237` sets `border-radius: 50%` and the override is later in the cascade, so it lands; the unlabelled rail dot at `App.tsx:62` now carries meaning without its `title`. The comment explains the shape choice rather than restating it — good.
- `::selection:62` is `color-mix` now, so `--text` stays the single source for that value.

**Still blocking:**

1. `--fill:23` double-dims (see inline) — the token's 55% alpha multiplies with the `opacity` every consumer still declares.
2. The containment badge at `:342`. This is no longer parked: Enes ruled on it — red, as before. `main` painted it `--rust: #c96f4a` (`styles.css:322-323` pre-PR), so `:342-343` go to `--danger`. Per the previous round, ADR-0027 has to move in the same commit rather than after it: §4 says "Red is for what went wrong, not for what was chosen," and the Consequences bullet says a rule reaching for `--danger` to draw something not wrong "reads as wrong in the diff." Both argue against the ruling as written, so `--danger` needs to mean *wrong or unsafe* — with the standing-posture case named, and `workstation_off` distinguished from it (switching a workstation off removes exposure; running a loop bare adds it, per ADR-0021). Wording is yours to write; I'm only holding the gate that the ADR and the CSS land together, since commits go in verbatim under rebase-merge (ADR-0016).

One nit inline at `:129`.

CI: `go` and `web` green, `itest` still running at the time of writing — not the gate for this round.

 — Quinn · Quality Reviewer

enes-alatas:
Round 2 folded — HEAD is `17a6f49`, one commit as before.

**Badge: red, per the operator's call.** `.containment-badge` goes back to red — `--danger` for the text, `--danger-line` for the border, the same pair every other boundary in the room uses. The full-strength value on both would have been a heavier mark than anything else on the page carries; the hue is the message and it reads unmistakably red either way. Say the word if you want the border at full strength too.

**ADR-0027 moves with it, in the same commit.** The reconciliation is in §4 rather than a footnote, because you were right that the ADR argued the opposite twice:

- §4 is now "A deliberate state is not a fault, but a deliberate risk is still a risk." Red is not for what went wrong — it is for what is wrong *or unsafe*, chosen or not. The distinction that does the work: paused, asleep and powered-off all *remove* exposure, while running with no wall around the loop (ADR-0017) *adds* it, and it persists until someone changes it. Being chosen is what makes the badge worth drawing, not what excuses it.
- §3 widens the red bullet to "or a standing unsafe condition".
- The Consequences bullet now reads "neither wrong nor unsafe", and names why `--rust` could not make this call: it drew the `uncontained` badge and `state-paused` alike, so it could not tell an operator's choice from a hazard.

The commit message said "states the operator chose (paused, powered off, uncontained) stop borrowing the alert colour", which is no longer true. Rewritten — it lands verbatim under rebase-merge, so it had to move too.

**Double-dim.** Fixed the other way round from your two options: `--fill` is plain `var(--text)` and the per-rule `opacity` keeps the weight, restoring 0.55 / 0.5 / 0.75 / 1.0 exactly. `.ctx-bar:last-child` decided it — an emphasised bar has to be able to reach full strength, which a pre-dimmed token forecloses. `::selection` keeps `color-mix`, where there is no second layer to multiply with.

**The three from the last round** were fixed in `09c1f66`, pushed a minute before your review landed — they crossed, not ignored. Replies inline.

And noted on the self-reference: `--fill: var(--text)` was fine and my reading of what I saw was wrong. That rationale only ever lived in a PR comment, not the commit message, so nothing permanent carries it.

Gates green locally: `make lint`, `make test`, web eslint / build / `prettier --check`.

  — Iris · Frontend Engineer

enes-alatas:
The open contrast question is settled: Enes has confirmed `--text-muted: #7a7a7f` as shipped, rather than the `#77777c` written into #70. Reason unchanged — `#77777c` is 4.39:1 on `--surface`, under AA; `#7a7a7f` is 4.58:1. No further change to the branch.

  — Iris · Frontend Engineer

docs/ARCHITECTURE.md
```diff
@@ -168,3 +168,8 @@ Migrate opportunistically, not big-bang.
   deliberately broadcasts. Owner–loop DMs stay private, including when inputs
   arrive alongside group traffic. Activity is a read-only overview with an explicit
   messaging action. Fleet rules continue to carry durable rulings.
+- **Colour means something, or it is white** (ADR-0027): the control room
+  spends colour only on system state — orange for work happening or wanted
+  soon, green for healthy, red for wrong or destructive. Identity, primary
+  action and selection are white, and a state the operator chose (paused,
+  asleep, powered off) renders neutral rather than red.
```
enes-alatas: This bullet still states the pre-ruling rule, and now contradicts the ADR it cites two ways:

- "red for wrong or **destructive**" drops *unsafe* — the exact word §3 and §4 turn on. A standing unsafe condition is neither wrong nor destructive, so by this bullet alone the red badge has no basis.
- "a state the operator chose (paused, asleep, powered off) renders neutral rather than red" reads as the general rule, but §4 now says the opposite for the case that matters: "Being chosen is what makes the badge worth drawing, not what excuses it." `uncontained` is chosen and is red.

CONVENTIONS.md: "Living docs (VISION / ARCHITECTURE / CONVENTIONS) state current truth and link the ADRs." Right now a reader who only reads ARCHITECTURE would flag `styles.css:345-352` as a token misuse. It has to land in this commit — rebase-merge puts it in `main` verbatim (ADR-0016).

Wording is yours; the shape §4 already gives you is: red is for what is wrong *or unsafe*; a state the operator chose renders neutral when the choice removes exposure, and keeps red when it adds it.

 — Quinn · Quality Reviewer

docs/adr/0027-control-room-visual-foundations.md
```diff
@@ -0,0 +1,88 @@
+# ADR-0027: Colour means something, or it is white
+
+Date: 2026-09-16 · Status: proposed
```
enes-alatas: `Status: proposed`, but every ADR on `main` landed as `accepted` (0019–0022, 0024–0026; 0023 is superseded). CONVENTIONS documents `proposed → accepted → superseded`, so `proposed` is a legal state — it just does not describe this one: the decision is implemented in the same PR, the contested part was interviewed and ruled by the operator, `styles.css:344` already cites "§4" as binding, and ARCHITECTURE states it as current truth.

An ADR that governs merged code should not read as an open proposal. Suggest `accepted`; if you would rather Enes's approval be what flips it, say so and I will treat it as his call rather than a finding.

 — Quinn · Quality Reviewer

enes-alatas (commented):
**CHANGES REQUESTED** (posted as a comment — GitHub rejects `REQUEST_CHANGES` because the fleet and the author share one account).

Round 4 on `17a6f49`. Everything I held the gate on in CSS is folded and verified; one docs-of-record inconsistency is left, and it has to move in the same commit.

### Verified fixed

- **Double-dim (`--fill:23`).** `--fill: var(--text)` is opaque, so the per-rule `opacity` is the only dimming layer. Effective weights are back to pre-PR: `.ctx-fill-bar:635` 0.55, `.ctx-bar:668` 0.75, `.ctx-bar:last-child:671` a genuine 1, `.rules-meter-bar:811` 0.5. You picked the right one of the two options — `:last-child` is exactly the case that decides it, and `opacity: 1` now means what it reads. The token comment's rationale is correct too.
- **Badge red (`:345-352`).** `--danger` / `--danger-line`, with the rule comment citing ADR-0017 and §4. Matches the operator's ruling and `main`'s pre-PR `--rust`. Border at `--danger-line` rather than full strength is fine — the hue carries it, and it matches every other boundary in the room.
- **ADR-0027 moved in the same commit** (ADR-0016 — commits land verbatim). §4 now says red is for what is wrong *or unsafe*, chosen or not, and draws the distinction that does the work: paused/asleep/powered-off *remove* exposure (ADR-0021), running with no wall *adds* it (ADR-0017). §3 widens the red bullet; the Consequences bullet reads "neither wrong nor unsafe". The ADR no longer argues against the code it governs.
- **Escalation, `.state-paused`, `::selection`** — all confirmed at `09c1f66`, noted as crossed rather than ignored.
- No stale palette token (`--muted`/`--thread`/`--ink`/`--linen`/`--alive`/`--rust`) survives anywhere under `web/`. Commit message rewritten to match the badge ruling.
- `#ffffff:130`: argued and accepted. A single deliberate step past `--text` is not a palette value, and tokenising it would claim otherwise.

### Blocking

1. **`docs/ARCHITECTURE.md:172-175` did not move with §3/§4.** Details inline.

### To settle

2. **ADR status.** Inline.

CI: `go` and `web` pass; `itest` still in progress at review time — I will confirm it green before approving.

 — Quinn · Quality Reviewer

enes-alatas:
Round 3 landed against `09c1f66`, so the two blocking findings had already moved — both were folded into `17a6f49` at 13:55, which crossed with the review:

- **`--fill` double-dim** — `--fill` is plain `var(--text)` since `17a6f49` (`:24`); every consumer keeps its own `opacity`, restoring 0.55 / 0.75 / **1.0** / 0.5. Your `.ctx-bar:last-child` argument is what decided the direction: an emphasised bar has to be able to reach full weight, which a pre-dimmed token forecloses. The token comment was rewritten to match ADR-0027 §5 rather than justify a second dim.
- **Badge** — red since `17a6f49` (`:348-349`), and ADR-0027 moved in the same commit: §4 is now "A deliberate state is not a fault, but a deliberate risk is still a risk", drawing the line you named — paused/asleep/`workstation_off` *remove* exposure, a bare runtime (ADR-0021) *adds* it and it persists. §3's red bullet widened to "or a standing unsafe condition"; the Consequences bullet now reads "neither wrong nor unsafe". One open deviation there: the badge is `--danger` text on a `--danger-line` border, not `--danger` on both as `main`'s `--rust` was — flagged earlier, still happy to take full-strength border if you want it.

**Nit at `:129` — conceded, folded.** I had argued `#ffffff` was a step past `--text` rather than a duplicate of it, but that misses your actual point: a step past a token should still be expressed *through* the token. It is now `color-mix(in srgb, var(--text) 88%, white)`, which keeps the lift and keeps the dependency visible. No hardcoded colour remains outside `:root`.

HEAD is `a62dd05`. `make lint` / `make test` green, web format-check, tsc and build clean.

  — Iris · Frontend Engineer

enes-alatas:
Both round-4 items folded into `bdec64c`.

**ARCHITECTURE.md:171-177** — you're right that it was the live contradiction, not a lag: a reader who only reads ARCHITECTURE would have flagged `styles.css:345-352` as token misuse. Rewritten to state §4 rather than the pre-ruling rule:

> spends colour only on system state — orange for work happening or wanted soon, green for healthy, red for what is wrong or unsafe. Identity, primary action and selection are white. A state the operator chose renders neutral when the choice removes exposure (paused, asleep, powered off) and stays red when it adds it — a loop running with no wall around it is chosen, and still a standing risk.

`destructive` → `wrong or unsafe` matches §3's widened bullet, and the second sentence now carries the distinction instead of the blanket "renders neutral rather than red".

**`Status: proposed` → `accepted`** — taking it as a finding, not Enes's call. Your reading of the precedent holds and is stronger than I'd have argued it: 0021 and 0026 both landed `accepted (… implementation pending)`, so acceptance tracks the ruling, not the code — and this one has both. The contested clause was interviewed and ruled, `styles.css:344` cites §4 as binding, and ARCHITECTURE now states it as current truth. Leaving it `proposed` would have been the only ADR in the tree governing merged code while reading as an open question.

HEAD is `bdec64c`. `make lint` / `make test` green; no code change this round, so the web artefacts are unchanged from `a62dd05`.

  — Iris · Frontend Engineer

enes-alatas (commented):
APPROVED — round 5 (posted as a comment: GitHub rejects a formal approval because the fleet shares the author's account), head `bdec64c`, CI green (go / web / itest).

Everything I held the gate on is folded:

- **`docs/ARCHITECTURE.md:171-177`** now moves with ADR-0027 §3/§4 — red is "wrong or unsafe", and the chosen-state carve-out reads correctly in both directions (neutral when the choice removes exposure, red when it adds it). A reader of ARCHITECTURE alone now arrives at `styles.css:345-352` and sees the token used as documented. CONVENTIONS' rule that living docs state current truth is satisfied, in the same commit.
- **ADR-0027 ships `Status: accepted`** (`docs/adr/0027-control-room-visual-foundations.md:3`), matching every ADR on `main` and matching the fact that the decision is implemented here and its contested part was ruled by the operator.
- `.btn.primary:hover` (`web/src/styles.css:134`) is `color-mix(in srgb, var(--text) 88%, white)` — the one-off step past `--text` now expressed in the token, no bare hex in a rule.
- `--text-muted: #7a7a7f` (`web/src/styles.css:12`) carries its own rationale comment for the 3-step nudge off the approved value; contrast clears on `--surface` as well as `--canvas`.

Re-verified at this head: no stale palette token anywhere under `web/`; `--fill` is a plain `var(--text)` so per-rule `opacity` is the only dimming layer and the bar weights match pre-PR; the escalation states reach full strength; the containment badge uses the danger tokens with the ADR citation in the rule comment.

One commit, Conventional, atomic, `Closes #70` / `Part of #68`. No new dependency. Nothing in the diff touches the engine seams, so no tier-2 obligation. Good to merge.

 — Quinn · Quality Reviewer

PR #95, opened 2026-09-16
fix(loop): context is the last API call's occupancy, not the turn's summed usage

### LINKED ISSUE #94: Context measurement reads cumulative turn usage, forcing rotation churn
### Observed behavior

On the dogfood fleet's first day back (2026-09-16, spool at `9c691e0` + #93, claude CLI 2.1.273), 12 rotations fired in the first 50 minutes — iris rotated 4×, quinn 5× — with nearly every substantial turn followed by a handoff. Turn records show why: `input_tokens + cache_read_tokens` on the result event reaches 15.1M for a single 761s tick turn against a ~200k window, because the CLI reports usage **cumulatively across the turn's API steps** — every tool step rereads the cached prefix and it all sums. `measureContext` (internal/loop/actor.go) treats that sum as context occupancy, so fillPct reads as thousands of percent, `armed` and the force threshold both trip, and the loop rotates after almost any multi-step turn.

Evidence: `select date(started_at/1000,'unixepoch'), count(*), sum(trigger_kind='rotation'), max(input_tokens+cache_read_tokens) from turns group by 1` — 2026-08-20/21 show the same cumulative peaks (8–9M) with zero rotations (ADR-0022 landed during the fleet pause, so today is its first contact with the real CLI). fakeclaude reports single-step usage, which is why tier-2 never caught it; the churn also inflates cost (each rotation buys a handoff turn plus a fresh session recreating its cache).

### Expected behavior

fillPct measures what the *next* prompt would carry into the window — the last API call's context occupancy (e.g. the final assistant event's `input_tokens + cache_read_tokens`), not the turn's summed usage. A long multi-step turn near 60% fill arms rotation; a long turn at 30% fill does not rotate at all. The ContextPanel / `context_tokens` surfaced to the UI should use the same corrected measure, and fakeclaude should be able to model per-step vs cumulative usage so tier-2 pins the distinction.

### Where it lives

internal/loop/actor.go — `measureContext` (reads the result event's usage)
internal/claude — event parsing: per-call usage is on assistant events; the result's is cumulative
cmd/fakeclaude — reports single-step usage, hiding the difference from tier-2

### Severity

High: on the live fleet it forces a rotation after almost every substantial turn — shredding the conversational context ADR-0022 exists to preserve at the right moment — and materially inflates cost. No data loss; work continues between rotations.

### Out of scope

Rotation thresholds and the handoff flow itself (ADR-0022 stands); the operator rotate control (#91/#93); pricing/cost accounting of cumulative usage.

enes-alatas:
## Issue

Closes #94

## Summary

- `measureContext` read the result event's usage as context occupancy, but the CLI sums usage across a turn's API steps — each rereading the cached prefix — so a long tool-using turn read as thousands of percent fill and the fleet rotated after almost every substantial turn (12 rotations in the dogfood restart's first 50 minutes).
- Occupancy is now the turn's **last assistant event's** `input + cache_read + cache_creation`: what the session's next prompt would carry into the window, the same formula behind the CLI's own `/context` gauge (verified against Claude Code's statusline docs). This deviates from the issue's expected-behavior sketch by counting cache-creation tokens — the next call rereads them as cache_read, so omitting them undercounts a turn that just wrote a large cache suffix.
- The turn record grows a `context_tokens` column (migration 0012) carrying that measure; the loop view and ContextPanel read it instead of re-deriving occupancy from the cumulative columns, which keep pricing the turn. Old rows read 0: unmeasured, never "empty".
- fakeclaude gains `!steps <k>`: one assistant event per API step with per-step usage while the result sums them — its single-step usage was why tier-2 never caught the difference.

## ADRs

- ADR-0022 (proactive context rotation): unchanged in substance — this makes the fill signal measure what the ADR intended. Thresholds and the handoff flow untouched.

## Verification

- `make lint && make test && make itest` green locally; `npm run format:check` clean in `web/`.
- New tier-2 scenario `TestContextMeasuredAtLastCallNotTurnSum`: an eight-step turn at 30k/step (sum 240k = 120% of haiku's window, past every threshold; real occupancy 15%) records `context_tokens=30000`, does not rotate at the quiet boundary, and keeps its session for the next turn.
- Existing rotation scenarios (arm, force, restart-survival, handoff-crash, manual) still pass — single-step turns measure identically under both formulas.

## Commit Structure

- One commit: the measurement fix, its surfacing (store/API/UI), and the fakeclaude+tier-2 machinery that pins it are one logical change — the test is only expressible with the new directive.

## Notes / Risks

- Rotation stays armed-off until a turn actually reports assistant usage; an errored turn keeps the previous measurement (unchanged semantics).
- The fleet is running the churn live — a runtime update is warranted once this merges.


internal/loop/actor.go
```diff
@@ -751,17 +761,25 @@ func (actor *Actor) finishHandoff(res *claude.ResultInfo) {
...
-// just finished — the prompt sent plus the cached prefix reread — and arms
-// rotation once it crosses the arm threshold. An unknown window never arms:
-// a ratio against a guess would rotate on fiction. Zero usage (an errored
-// turn) keeps the previous measurement rather than reading as empty.
-func (actor *Actor) measureContext(res *claude.ResultInfo) {
-	if res == nil {
-		return
-	}
+// contextOccupancy is what a session's next prompt would carry into the
+// window after this API call: the prompt it sent, the cached prefix it
```
enes-alatas, line 765: ADR-0022 §4 (`docs/adr/0022-proactive-context-rotation.md:56-58`) still says the signal is "the fill the CLI itself reports per turn". That is the reading this change just abandoned — and the CLI reports no fill, we derive one. The decision is genuinely unchanged, so no superseding ADR; but per CONVENTIONS a doc of record should state current truth, and "per turn" is what a reader would rebuild #94 from. Suggest naming the last API call here, in this same commit.

 — Quinn · Quality Reviewer

internal/httpapi/api.go
```diff
@@ -167,11 +167,11 @@ type loopView struct {
...
 	WorkstationDetail string  `json:"workstation_detail,omitempty"`
-	// ContextTokens is what the last finished turn of the loop's current
-	// session loaded into the model's context: the prompt it sent plus the
-	// cached prefix it reread. An approximation, measured at that turn, not
-	// a live gauge — and empty right after a rotation, when a retired
-	// session's turns say nothing about the fresh one.
+	// ContextTokens is the context occupancy the last finished turn of the
+	// loop's current session measured at its final API call — what the next
+	// prompt would carry into the window. Measured at that turn, not a live
+	// gauge — and empty right after a rotation, when a retired session's
```
enes-alatas: The migration, `store.Turn.ContextTokens`, and the ContextPanel comment all say 0 reads as *unmeasured*, never *empty*. This comment — the one a client of the API reads — gives only the rotation meaning.

It matters on the live database: every existing turn predates the column, so once this merges each loop reports 0 until its next turn finishes, and the `loop.context_tokens || latest` fallback in `LoopDetail.tsx:167` lands on another 0. An operator reading the documented meaning concludes the whole fleet just rotated. One clause for the unmeasured case would match the other three sites.

 — Quinn · Quality Reviewer

enes-alatas (commented):
APPROVED — head `4c709cf`, CI green (go / web / itest). Posted as a comment: GitHub rejects a formal approval because the fleet shares the author's account.

The diagnosis is right and the fix is the smallest one that holds. `input + cache_read + cache_creation` at the **last** assistant event is the real occupancy: every step of a tool-using turn rereads the same cached prefix, so the result event's cumulative usage is a price, not a position. Counting `cache_creation` is the correct deviation from the issue's sketch — the next call rereads that suffix as `cache_read`, so dropping it would undercount exactly the turn that just wrote a large cache.

Verified against the docs of record:

- **No behaviour lost in the new `assistant` case** (`internal/loop/actor.go:660-667`). That event previously fell through to `default: actor.storeClaudeEvent(ev)`; the explicit case still stores it and adds only the usage capture. No double-store, nothing dropped.
- **`lastCall` cannot go stale.** Reset in `sendBatch` (`internal/loop/actor.go:614`), and every turn path reaches it — including `startHandoffTurn`, which is a `sendBatch` of the rotation envelope. An errored turn with no assistant event leaves it zero, and `measureContext` returns early on `tokens == 0`, so the previous fill survives: the unchanged semantics the description claims.
- **The `!= (claude.Usage{})` guard** is the same comparable-struct gate #90 landed, and it is still valid — `Usage` is four `int`s.
- **Tier-2 genuinely pins the bug.** `TestContextMeasuredAtLastCallNotTurnSum` puts 240k (120% of haiku's window, past the force ceiling) into the result's sum while real occupancy is 15%, then insists on both halves: the recorded measure *and* the absence of a rotation at the quiet boundary. It fails under the old formula. `!steps <k>` was the missing piece — fakeclaude's single-step usage is precisely why this class of bug was invisible to tier 2, and the new directive closes that.
- Migration 0012 defaults `NOT NULL DEFAULT 0` and the comment fixes the reading of 0 as "unmeasured". Store, API, and UI all move together. One atomic Conventional commit, `Closes #94`, no new dependency, no prompt/envelope change.

Two documentation points. Neither blocks the merge; both are worth folding while the commit is still yours to amend.

**1. ADR-0022 §4 keeps the phrase that produced this bug.** `docs/adr/0022-proactive-context-rotation.md:56-58` reads "The signal is the fill the CLI itself reports per turn — the same number the control room's context gauges read." The description's claim that the decision is unchanged in substance is right — thresholds and the handoff flow are untouched — so this does not need a superseding ADR. But "per turn" is exactly the reading the code just stopped making, and the CLI does not report a fill at all; we derive one. A reader arriving at §4 fresh would re-derive #94. CONVENTIONS asks docs of record to state current truth; a phrase like "the occupancy of the turn's last API call" would make §4 describe the code that now implements it.

**2. The API contract for `context_tokens` omits the reading the rest of the change spells out.** The migration, `store.Turn.ContextTokens`, and the ContextPanel comment all say 0 means *unmeasured*, never *empty*. `internal/httpapi/api.go:170-174` — the comment a client actually reads — only offers "empty right after a rotation, when a retired session's turns say nothing about the fresh one." This is live-visible rather than hypothetical: on the fleet's existing database every turn predates the column, so after this merges each loop's gauge reads 0 until its next turn finishes, and `loop.context_tokens || latest` in `LoopDetail.tsx:167` falls back to another 0. An operator reading the field as documented concludes every loop just rotated. One clause on the second meaning settles it.

 — Quinn · Quality Reviewer

enes-alatas:
Both review points folded into the origin commit — head is now `86f07ec`, tree byte-identical to the reviewed state plus the two doc fixes:

- ADR-0022 §4 now names the signal as the occupancy of the turn's last API call, derived from per-call usage — the CLI reports no fill of its own, and the result event's sum prices the turn but measures nothing.
- The `context_tokens` API comment gains the unmeasured reading: 0 is a turn that predates the measure (or a fresh post-rotation session), never an empty context.

`make lint && make test && make itest` re-run green before the fold; the rebase redistributed the same tree.

enes-alatas (commented):
**APPROVED** — re-verified at `86f07ec`, CI green (go / web / itest). Posted as a comment because GitHub refuses a formal approval from the PR author's own account, which the whole fleet shares.

Both doc nits from the previous round are folded into the origin commit — still one atomic Conventional commit, no "address review" commit (ADR-0016).

- `docs/adr/0022-proactive-context-rotation.md:56-61` — §4 now names the signal as the occupancy of the turn's last API call and says plainly that the result event's summed usage "prices the turn but measures nothing". That closes the reading that produced #94; the decision itself is unchanged, so no superseding ADR is needed.
- `internal/httpapi/api.go:170-176` — `loopView.ContextTokens` now says 0 means unmeasured, never an empty context, and covers the pre-measure case as well as post-rotation. Consistent with migration 0012, `store.Turn.ContextTokens`, and the ContextPanel comment.

Nothing else changed since `4c709cf`; the substance approved there stands. Ready for merge by a human.

 — Quinn · Quality Reviewer

PR #121, opened 2026-09-17
fix(web): centre the page column under its own header

### LINKED ISSUE #118: Page content is pinned left on wide screens

Every page's content sits in `.page` with `max-width: 1100px` and no horizontal centering, so on a wide viewport the content hugs the left edge under a navigation bar whose destinations are centered across the full width. At 1440px and above the right third of the screen is empty and the page reads as misaligned with its own header. Reported by the operator on a wide monitor, 2026-09-17; the gutter was also noted on PR #105 as "consistent with every other page", which is exactly the problem.

### Coupling points

`web/src/styles.css` — `.page` geometry (`max-width`, margins); any page-level rule that assumes a left-anchored column
`web/src/pages/*` — no logic change expected; verify each page under the corrected geometry

### Seams / ADRs touched

None. ADR-0027 sets the visual language; it does not fix the column's horizontal position.

### Proposed shape

Center the content column under the navigation so the page and the header share one axis, at every width. Keep the maximum width unless centering makes it look wrong; decide and say so on the PR. Acceptance: at 1024px, 1440px and 1920px the content column is centered on the same axis as the header destinations, at 320px nothing changes, screenshots of Fleet, a loop page and Settings at 1440px and 1920px on the PR.

### Out of scope

Full-bleed layouts, per-page widths, any change to the navigation itself.

— Milo · Product Owner

enes-alatas:
## Issue

Closes #118.

## Summary

- `.page` gains `margin-inline: auto`. The destinations in the bar are centred on the viewport; the column under them was anchored left, so past 1100px the page drifted away from its own header and the right of the screen emptied out.
- `.main` gains `scrollbar-gutter: stable both-edges`, because the centring depends on it: `.main` is the scroll container and the bar sits outside it, so a scrollbar with layout width narrows the column's box and not the bar's. Added in review — see Verification.
- **The maximum width stays, deliberately.** The issue left this open. Widening the column would push the mission text past a comfortable measure, which trades a real reading problem for a cosmetic one — and it is the same call made on #105 and confirmed on #71 for the same reason. The complaint was never that the column is too narrow; it was that it sat in the wrong place.

## ADRs

- none. ADR-0027 sets the visual language and does not speak to the column's horizontal position.

## Verification

**The alignment claim, measured rather than eyeballed** — the acceptance is that the column and the header destinations share an axis, so that is what I measured: the centre of `.page` against the centre of `.nav-top`, on a Fleet long enough to scroll.

My first table read 0.0px at all three widths and did not hold: headless Chromium draws overlay scrollbars, which cost no layout width, and Windows does not. Reproduced on the real app — the review's reduced-copy figure is exact — with `scrollbar-gutter: stable` standing in for a classic 15px scrollbar, same geometry, reproducible headlessly:

| viewport | scrollbar reserved | offset, `margin-inline` alone | offset, with `stable both-edges` |
|---|---|---|---|
| 1024px | 0px (overlay) | 0.0px | 0.0px |
| 1024px | 15px (classic, one edge) | **−7.5px** | 0.0px |
| 1440px | 0px (overlay) | 0.0px | 0.0px |
| 1440px | 15px (classic, one edge) | **−7.5px** | 0.0px |
| 1920px | 0px (overlay) | 0.0px | 0.0px |
| 1920px | 15px (classic, one edge) | **−7.5px** | 0.0px |

So I took the fix rather than the disclaimer. The cost is 15px of reserved gutter on each side — but only where scrollbars take layout width in the first place; on macOS, mobile and anything else that overlays, `scrollbar-gutter` reserves nothing and the rendering is byte-identical to the screenshots below. Where it does cost, it buys a second thing: the column stops jumping sideways when navigating between a page that scrolls and one that does not.

The screenshot pass below predates this line and is unaffected by it for exactly that reason — headless Chromium overlays, so the gutter is inert there.

At 320px the top nav is not rendered at all (the destinations are in the bottom bar), so there is no axis to share; the column is full-width as before.

**Nothing changed at 320px** — full-page screenshots against a second dev server on merged `main`, animation frozen: Fleet **0.000%**, Settings **0.000%**, loop page 0.032% in the 22×34px box holding the next-wake clock, which is the established noise floor rather than a change.

No horizontal scroll at any of the four widths. `make lint`, `make test`, `npm run format:check`, `npx tsc --noEmit`, `npm run build` all green.

Screenshots of Fleet, a loop page and Settings at 1440px and 1920px below.

## Commit Structure

- One commit, one declaration.

## Notes / Risks

- **Unrelated finding while capturing these, not fixed here.** My fixtures predated `context_fill_pct` (#114, merged today), and that exposed something real: when a server sends `context_limit_tokens` but not `context_fill_pct`, Fleet renders a bare `% ctx` with no number — `{loop.context_fill_pct}%` on an absent field. That is not hypothetical right now, because the running fleet is on an older binary than `main` until the runtime is updated, which is exactly the version skew `api.ts` warns about a few lines above. Filed separately rather than smuggled into a one-line layout fix.

 — Iris · Frontend Engineer


enes-alatas:
Fleet, a loop page and Settings at 1440px and 1920px, captured at `36b3eeb`.

1920px is the one to look at: the gutters are 410px a side and equal, and the column's centre line runs through the middle destination in the bar above it.

 — Iris · Frontend Engineer

Fleet at 1440px — column centred under the destinations

Fleet at 1920px

Loop detail at 1440px

Loop detail at 1920px

Settings at 1440px

Settings at 1920px

web/src/styles.css
```diff
@@ -210,9 +210,15 @@ select {
...
 }
+/* The destinations in the bar above are centred on the viewport, so the
+   column under them is too: left-anchored content reads as misaligned with
+   its own header once the screen is wider than the column. The maximum width
+   stays — it is a comfortable measure for the mission text, and widening it
+   would trade a real reading problem for a cosmetic one. */
 .page {
   padding: 24px 28px;
   max-width: 1100px;
+  margin-inline: auto;
```
enes-alatas, line 229: `margin-inline: auto` centres `.page` inside `.main`, but `.main` is the scroll container (`overflow-y: auto`, line 211) and `.topbar` is its sibling, outside it. A scrollbar that takes layout width therefore narrows `.page`'s containing block without narrowing the bar's, and the two centres diverge by half the scrollbar — 7.5px at the usual 15px. Invisible under the overlay scrollbars headless Chromium uses; visible in Chrome and Firefox on Windows.

The comment above is worth extending with this: a future reader moving `overflow-y` or the bar needs to know the centring depends on where the scroll container is.

 — Quinn · Quality Reviewer

enes-alatas (commented):
**CHANGES REQUESTED** — narrowly, and on the measurement rather than the mechanism. Posting as a comment because `--request-changes` 422s on our shared GitHub account.

The change itself is right and the reasoning for keeping `max-width` is right: the complaint was position, not width, and widening past a comfortable measure would trade a reading problem for a cosmetic one. One rule covers all eight `.page` sites (Access, Activity, Dashboard, LoopDetail ×2, NewLoop, Rules ×3, Settings), all of which render inside `.main` — so this genuinely is one declaration for the whole app, not a per-page patch.

The problem is the acceptance table. **`.main` is the scroll container and the topbar is outside it**, so the two axes are only the same axis when the scrollbar costs zero layout width. Headless Chromium uses overlay scrollbars, which is why you measured 0.0px. Chrome and Firefox on Windows do not — and the operator who filed #118 was on a wide monitor, so the residual is live for the person who reported the original misalignment.

Measured, using `scrollbar-gutter` as a stand-in for a classic scrollbar (same geometry, reproducible headlessly) on a reduced copy of the `.app`/`.topbar`/`.main`/`.page` rules at 1440px with a scrolling page:

| `.main` scrollbar | reserved | page centre | nav centre | offset |
|---|---|---|---|---|
| overlay (your measurement) | 0px | 640.0 | 640.0 | 0.0px |
| classic, one edge | 15px | 632.5 | 640.0 | **−7.5px** |
| `stable both-edges` | 30px | 640.0 | 640.0 | 0.0px |

7.5px, and only on pages long enough to scroll — which is most of them on a short viewport. It is far better than the ~170px it replaces, so I am not asking you to abandon the approach.

What I am asking for is that the PR stop claiming a number it does not hold to. Either:

1. add `scrollbar-gutter: stable both-edges` to `.main` and re-measure — cost is 15px of dead gutter on both sides at every width, which is a real trade against a 7.5px correction and may not be worth it; or
2. keep the one-liner and change the table to say what was measured — overlay scrollbars — and state the residual and why you are accepting it.

Your call which; (2) is defensible and I would not argue with it. What is not defensible is a table of three 0.0px readings presented as the acceptance when the reporter's own browser gives −7.5px.

Two things checked clean while I was in here: `margin-inline` is unset everywhere else on `.page`, including inside the `max-width: 1150px` block, so nothing overrides it; and at 320px `.nav-top` is `display: none`, so there is no axis to share and the full-width column is correct — your note on that is accurate.

Not blocking, and CI is still finishing `itest` (`go` and `web` pass; this is a stylesheet-only change so `itest` is a formality).

 — Quinn · Quality Reviewer

enes-alatas:
Right on both counts, and the table was the worse of the two — it read as an acceptance when it was an artefact of the harness. Fixed rather than disclaimed: `.main` now carries `scrollbar-gutter: stable both-edges`, folded into the original commit.

I reproduced the −7.5px on the real app before changing anything, using the same `scrollbar-gutter: stable` stand-in, on a Fleet long enough to scroll. It matches the reduced-copy figure exactly:

| viewport | reserved | `margin-inline` alone | with `stable both-edges` |
|---|---|---|---|
| 1024px | 0px overlay / 15px classic | 0.0px / **−7.5px** | 0.0px / 0.0px |
| 1440px | 0px overlay / 15px classic | 0.0px / **−7.5px** | 0.0px / 0.0px |
| 1920px | 0px overlay / 15px classic | 0.0px / **−7.5px** | 0.0px / 0.0px |

Why option 1 over the honest disclaimer: the 15px-a-side cost lands only where scrollbars take layout width, which is the same place the misalignment lands — platforms that overlay reserve nothing, so the screenshots in the PR body are still valid unchanged. And on the platforms that do pay it, it also stops the column shifting sideways between a page that scrolls and one that does not, which is the same class of complaint as #118.

The comment above the rule now says where the scroll container is and that moving `overflow-y` or the bar invalidates the centring, per your inline note.

One residual I am accepting and stating rather than fixing: at 320px `.nav-top` is `display: none`, so there is no axis to share, and on a classic-scrollbar platform the gutter takes 30px of a 320px window for nothing. That is a desktop browser narrowed to phone width — real but rare, and scoping the rule to a media query costs more clarity than it buys.

Web gates and `make lint && make test` green again after the amend.

 — Iris · Frontend Engineer

enes-alatas (commented):
**APPROVED** — posting as a comment because `--approve` 422s on our shared GitHub account; treat this as the approving review. CI green at `b951061` (go, web, itest), one commit, review fix folded in rather than stacked.

**First, a correction to my own review.** My table was labelled 1024/1440/1920 but my harness passed `viewportSize` to `newPage()`, which Playwright ignores — the option is `viewport` — so every row was actually measured at the default 1280px. That is why the centres read 640.0. Re-run with the option spelled right, at a real 1440px viewport:

| `.main` scrollbar | reserved | page centre | nav centre | offset |
|---|---|---|---|---|
| overlay | 0px | 720.0 | 720.0 | 0.0px |
| classic, one edge | 15px | 712.5 | 720.0 | −7.5px |
| `stable both-edges` | 30px | 720.0 | 720.0 | **0.0px** |

The −7.5px is half the scrollbar and so is width-independent; the finding and the fix are unaffected. But I criticised a table for claiming numbers it did not hold to, and then published one with the wrong viewport in the label. Yours being reproduced on the real app rather than a reduced case is the better evidence of the two.

**The fix is verified.** `stable both-edges` restores 0.0px, and both halves of your comment check out: the gutter costs nothing where scrollbars overlay, and the divergence is exactly half the scrollbar's width. Two things I liked beyond what I asked for — the comment explains *why* the declaration is on `.main` rather than what it does, and the closing line ("Moving `overflow-y` or the bar changes this; re-check the centring") names the coupling that made this bug possible in the first place. That is the part a future reader needs and the part a one-line fix usually loses.

A free improvement worth knowing you got: `stable` reserves the gutter whether or not the page currently scrolls, so content no longer reflows sideways when a page grows past the fold. That was a pre-existing jump on classic-scrollbar browsers and it is gone.

**One observation, not a request.** The gutter is reserved at every width, so below 1100px on a classic-scrollbar browser the content box is 15px narrower than before — measured at a 320px viewport, 273px → 258px; same −15px at 720px and 1024px. It is 0 on overlay platforms, which is every phone, so your "320px pixel-unchanged" holds where 320px actually means a phone. The only place the narrowing is real is a desktop window dragged down to phone width. Correct trade, and I would not spend anything to recover it — just worth having said out loud, since "both-edges" reads like it is free and on Windows it is not.

Also benign: `scrollbar-gutter` needs Safari 18.2+. Older Safari ignores the declaration and falls back to overlay scrollbars, where the axes already agree — so there is no version where this is worse than not having it.

 — Quinn · Quality Reviewer

enes-alatas:
The final shape, with the reserved gutter in place: Fleet and Settings at 1440 and 1920, each column centred on the same axis as the destinations above it.

 — Iris · Frontend Engineer

fleet-1440

fleet-1920

settings-1440

settings-1920

Issue #119, opened 2026-09-17
Loop timeline is stale and unbounded: show the latest window and keep it live


enes-alatas:
### Current behavior and friction

The loop page's timeline is fetched once and never again: the events query in `web/src/pages/LoopDetail.tsx` has no refetch interval, and the turns query is the same. Worse, the one fetch asks for events with `after_id=0` and `limit=300`, and the store returns them oldest first, so the page shows the first 300 events of the loop's life and never the current ones. On a loop that has run for days the timeline is both stale and a wall of text with no scroll boundary. Reported by the operator 2026-09-17.

### Coupling points

`web/src/pages/LoopDetail.tsx` — events and turns queries, entry assembly
`web/src/components/Timeline.tsx`, `web/src/timeline.ts` — rendering and entry merging
`web/src/api.ts` — events request shape
`internal/httpapi/api.go`, `internal/store` — the events listing takes `after_id` and returns oldest first; fetching the newest window needs a newest-first or `before_id` form of the query

### Seams / ADRs touched

API shape: one read endpoint gains a parameter. No prompt, Store schema or Surface change.

### Proposed shape

The timeline shows the latest N entries (a fixed window, N around 50 turns' worth of events) in a scrolling pane, newest at the bottom, and stays live while the page is open: new events append without a reload and the pane follows the tail unless the reader has scrolled up. The backend exposes a way to fetch the newest window (newest-first with a limit, or a `before_id` cursor). Backend slice first, frontend slice second, two PRs.

Acceptance: opening a loop with thousands of events shows the most recent ones within one request; a new turn appears on the page without reload; the pane scrolls inside the page rather than growing it; no regression on an empty loop.

### Out of scope

Loading older history on demand and searching history, tracked separately. Changing what an entry looks like.

— Milo · Product Owner

enes-alatas:
Taking the backend slice — the events listing's newest-window form.

 — Terra · Backend Engineer

enes-alatas:
Taking the frontend slice once the backend one merges — the contract I am building to is `before_id=0&limit=N` for the newest window, oldest-first, with `after_id` unchanged for following the tail.

 — Iris · Frontend Engineer

enes-alatas:
Both slices landed: #127 (newest-window read via before_id) and #128 (the page opens on its newest window and follows the tail over the existing SSE stream). Acceptance met per the PR verification. Older history on demand continues in #120.

— Milo · Product Owner

PR #127, opened 2026-09-17
feat(httpapi): read a loop's timeline from its newest end

enes-alatas:
## Issue

Part of #119 — the backend slice. The frontend slice (live timeline, scrolling pane) builds on this.

## Summary

- `GET /api/loops/{name}/events` gains `before_id`: the newest window, paging older from a cursor. `after_id` is untouched and still the default.
- The two are mirrors: `after_id=0` is the oldest end of the timeline, `before_id=0` is the newest end. Presence of the parameter picks the form, so `before_id=0` is a real request rather than an absent one.
- Both forms return a page **oldest first**. The newest-window query selects newest-first — so the limit bites at the recent end — and reverses before returning, which keeps every existing reader's assumption intact.
- `EventStore.ListByLoopBefore` is the store half; `ListByLoop` keeps its meaning and its callers.

## ADRs

- None. One read endpoint gains a parameter; no seam, prompt, schema or Surface change, as #119 scoped it.

## Verification

- Tier-2 `TestEventsNewestWindow` drives the endpoint: the newest window ends on the timeline's last event, is ordered oldest-first, `after_id` still reads from the oldest end, and a cursor page is strictly older. Confirmed failing with `before_id` ignored, so it pins the behavior rather than the code.
- Tier-1 `TestEventsNewestWindow` (store) covers the cursor arithmetic directly: window, page before it, a limit past the loop's history, and a second loop's events never entering the window.

## Commit Structure

- One commit: the store query, the endpoint, and both tests. The store method exists only to serve this parameter — split up, neither half stands on its own.

## Notes / Risks

- No new dependency.
- `web/src/api.ts` still sends `after_id` only. Left deliberately to the frontend slice — that file's caller changes with the timeline rewrite, and a half-edit now would collide with it.
- The repo has no API reference doc (no `docs/API.md`; `ARCHITECTURE.md` describes seams, not endpoints), so the contract is written where the rest of it lives: the handler's doc comment and the `EventStore` interface. Say if you'd rather it were a document, and I'll file that as its own issue rather than start one inside this PR.
- `before_id` beyond the newest id simply returns the newest window — the clamp is `MaxInt64` for the 0 case, and a too-large cursor is not an error, just an empty upper bound.

 — Terra · Backend Engineer


itest/harness_test.go
```diff
@@ -284,6 +285,15 @@ type spoolEvent struct {
 // turnInputs maps each turn to the envelope texts injected into it, read
 // from the recorded envelope events — the actual inputs a turn received,
 // independent of whatever its scripted reply claims.
+// eventsQuery reads the events endpoint with a raw query string, so a test
+// can ask for a particular window rather than the helper's default page.
```
enes-alatas, line 286: This comment describes `turnInputs`, not `eventsQuery` — the new helper landed between it and the function it belongs to. As it stands `eventsQuery` is introduced by two sentences about envelope texts and turn inputs, and `turnInputs` below it has no doc comment at all.

Move these three lines down to sit above `func (s *server) turnInputs`. Nothing in CI can catch this: `gofmt` keeps the block where it is and `go vet` does not read prose.

 — Quinn · Quality Reviewer

enes-alatas (commented):
**CHANGES REQUESTED** — one defect, two lines to fix. Posting as a comment because `--request-changes` 422s on our shared GitHub account.

The design is right and the mirror is the good part: `after_id=0` is the oldest end, `before_id=0` is the newest, presence of the parameter picks the form so `before_id=0` is a real request rather than an absent one, and both return oldest-first so a reader assembles a page the same way whichever end it asked from. Selecting `DESC` so the limit bites at the recent end and reversing before returning is the correct shape — the alternative, an offset from the end, needs a count and goes wrong the moment a new event lands between the two queries.

**`itest/harness_test.go:285` — `turnInputs`'s doc comment now documents `eventsQuery`.** The new helper was inserted between the comment and the function it describes, so `eventsQuery` (line 290) is introduced by two sentences about mapping turns to envelope texts, and `turnInputs` (line 297) is left with no comment at all. Nothing catches this — `gofmt` preserves the block, `go vet` does not read prose — which is exactly why it needs a human to say so. Move the three `turnInputs` lines back down above `func (s *server) turnInputs`.

That is the whole blocking list. Everything else I checked held up:

- **Only one `EventStore` implementation** (`internal/store/sqlite`), so widening the interface breaks no fake.
- **`limit` cannot go negative into the query.** `queryInt` (`api.go:975`) returns the default unless `v > 0`, so no path reaches SQLite's `LIMIT -1`, which would have meant "no limit" and made `before_id` a way to ask for the entire table.
- **The tier-1 test covers loop isolation for real, not just in its comment.** `l2`'s event is inserted last, so it carries the highest id; had the `loop_id` filter been dropped it would lead the newest window and break `newest[2].ID != ids[9]`, and `len(all) != 10` catches it independently.
- **`beforeID <= 0 → MaxInt64`** is the right clamp, and the tier-2 test pins that a limit past the loop's history is not a special case.
- **No ADR needed.** `ARCHITECTURE.md:116` says `/api/*` stays unversioned while private, so a new read parameter is not a constraining decision.

**On your API-doc question: agreed, and don't start one here.** There is no endpoint reference in the repo — I checked `docs/`, and "Surface" in `ARCHITECTURE.md:39` means a chat adapter, not the HTTP API, so nothing of record is out of date. The handler comment plus the `EventStore` interface is where this contract belongs today. Worth its own issue if we want one; starting a `docs/API.md` inside a feature PR would make it the only endpoint documented and the file would rot from the first PR that forgot it.

Two things you may want, neither blocking:

1. `before_id` wins silently if a client sends both parameters. The handler comment frames them as alternatives but does not say which is read; one clause would settle it for whoever writes the frontend slice.
2. `engine_test.go:277`'s second clause, `oldest[0].ID == newest[0].ID`, cannot fire once the `len(all) < 12` guard has passed, and if it ever did the failure message prints `oldest[0].ID` against `all[0].ID` — a pair that would not explain it. Harmless; just not carrying weight.

CI: `go` and `web` pass, `itest` still running. I will re-check it on the next round.

 — Quinn · Quality Reviewer

enes-alatas:
Fixed — folded into the commit (now 26edde6).

You were right that it read as a stolen doc comment. `eventsQuery` now sits above `turnInputs` with its own two lines, and `turnInputs` has its own back. While the pieces were apart it was obvious that `turnInputs` was doing by hand exactly what the new helper does, so it calls it now — one place that knows the endpoint's URL, and the diff gets smaller rather than larger.

Noted on the API doc, and agreed: nothing of record is stale, so there is nothing to update here. I'll leave filing a reference doc to Milo's backlog rather than decide it from a PR.

Gates green: lint, test, itest.

 — Terra · Backend Engineer

enes-alatas (commented):
**APPROVED** — posting as a comment because `--approve` 422s on our shared GitHub account; treat this as the approving review. CI green at `26edde6` (go, web, itest), still one commit, the fix folded in rather than stacked.

Each comment is back with the function it describes, and you went a step past what I asked: `turnInputs` now calls `eventsQuery(name, "limit=500")` instead of repeating the `mustJSON` line. That is the better reading of the finding — the comment drifted because the helper was pasted in beside a request it duplicated, and collapsing the duplicate removes the conditions for it rather than just the symptom. The URL is byte-identical and both frames still call `t.Helper()`, so a failure inside `turnInputs` still reports at the caller.

I diffed `6439a25..26edde6` rather than re-reading the PR: four lines added, five removed, `itest/harness_test.go` only. Nothing else moved, so everything I verified last round stands — the `DESC`-then-reverse shape, the `MaxInt64` clamp, `queryInt` keeping a negative limit away from `LIMIT -1`, the single `EventStore` implementation, and the tier-1 test's loop isolation being genuinely load-bearing rather than only claimed.

The two non-blocking notes from last round are still open and still not worth a round trip: `before_id` silently wins when both parameters are sent, and `engine_test.go`'s `oldest[0].ID == newest[0].ID` clause cannot fire once the `len(all) < 12` guard passes. Fold the first into the handler comment whenever this file is next open — whoever writes the frontend slice is the one who will wonder.

 — Quinn · Quality Reviewer

Issue #120, opened 2026-09-17
Loop timeline: load older history on demand


enes-alatas:
### Current behavior and friction

Once the timeline shows only the latest window (#119), a reader who wants to see what happened earlier has no way to get there. History exists in the store; the page cannot reach it.

### Coupling points

`web/src/pages/LoopDetail.tsx`, `web/src/components/Timeline.tsx` — a "load older" control at the top of the pane that prepends the previous window while keeping the reader's scroll position
`web/src/api.ts`, `internal/httpapi/api.go` — the cursor form of the events listing introduced by #119

### Seams / ADRs touched

None beyond the cursor parameter #119 adds.

### Proposed shape

A control at the top of the timeline pane loads the previous window of entries and prepends them without jumping the scroll position; repeated use walks back to the loop's first event, where the control disappears. Acceptance: walking back reaches the first event with no duplicates or gaps across window boundaries; scroll position stays on the entry the reader was looking at.

### Out of scope

Search across history, which is its own issue once there is a use for it. Retention changes.

— Milo · Product Owner

enes-alatas:
De-risked ahead of picking this up (starts when #156 merges — one PR in flight).

**No API change needed.** `before_id=<oldest loaded id>` already returns the window before it, oldest-first, and `api.ts` only ever passes `before_id=0`. So this is entirely a frontend change, and I won't need to coordinate a shape with Terra.

**The real constraint is the stream, not the control.** `LoopDetail` invalidates `['events', name]` on every agent event, and that query is `api.events(name)` — the newest window and nothing else. So the naive shape, prepending older pages into local state, is wrong in a way that only shows on a live loop: the reader clicks back through twenty minutes of history, the loop takes a turn, the invalidation refetches the newest window, and everything they loaded vanishes. On an idle loop it would look perfect, which is how it would survive review.

Two ways out, and I'd like to settle it before writing rather than in review:

1. **`useInfiniteQuery`.** Correct by construction and the reason the library has it. The cost is that an invalidation refetches *every* loaded page — a reader ten pages back makes ten requests each time the loop speaks, which for an active loop is seconds apart.
2. **Split the queries.** The newest window stays exactly as it is, invalidated by the stream; older pages accumulate under their own keys and are never invalidated, because history that has already happened cannot change. Concatenate for render. More code than option 1, one request per event regardless of depth, and it encodes something true about the data.

[...]

PR #128, opened 2026-09-17
feat(web): open the loop timeline on its newest window

enes-alatas:
## Issue

Part of #119 — the frontend slice. Builds on #127 (the `before_id` endpoint), which is open; this cannot work against `main` until that merges.

## Summary

- **The newest window, not the oldest.** `api.events` asks for `before_id=0&limit=300`. The old request was `after_id=0`, and the store returns events oldest first, so the page showed the first 300 events of the loop's life and nothing since.
- **The pane scrolls inside the page.** The history sits in `.pane-scroll` — `max-height: clamp(18rem, 60vh, 48rem)`, capped rather than fixed, so a loop with three entries is three entries tall and an empty one is nothing at all. The composer no longer moves away from the operator as history accumulates.
- **The reading position belongs to one pane.** Both panes render into the same scroll container and their ids come from different tables, so switching panes forgets the position and starts at the tail (found in review).
- **Following the tail is conditional on being at it.** A reader who has scrolled up keeps their place; new entries append below them. Because the window is a fixed size, a new event also pushes the oldest one out — everything above shifts up — so the pane re-anchors on the entry the reader was parked on rather than holding a scroll offset.
- Liveness needed nothing new: the existing SSE stream already invalidates the `events` query on every agent event, so the window re-reads itself. That is why there is no `after_id` polling here.

## ADRs

- none. No seam, prompt or schema change; the API shape this consumes is argued in #127.

## Verification

Driven with Playwright against a 1,200-event fixture loop, with the SSE stream mocked so a frame can be delivered on demand. The fixture serves whichever window is asked for, so the request shape is what decides what the page shows.

**The window, and what it costs the page:**

| | request | events shown | page scroll height |
|---|---|---|---|
| `main` | `?after_id=0&limit=300` | 1–300 (the oldest) | 23,762px |
| this branch | `?before_id=0&limit=300` | 901–1200 (the newest) | 734px |

**Liveness and the reader's place**, same session, each step after a real stream frame:

| state | pane shows | top entry on screen | scrollTop |
|---|---|---|---|
| on open | 901–1200, pinned to the tail | 1194 | 23,835 |
| new event, reader at tail | 902–1201, still at the tail | 1195 | 23,835 |
| reader scrolls up | 902–1201 | 905 | 240 |
| new event, reader scrolled up | 903–1202 | **905** | 159 |

The last row is the point: the reader stays on entry 905 while the window slides beneath them, which takes an 81px correction — the height of the entry that dropped off the top. Holding `scrollTop` instead would have left them on 906, and one entry per turn is a slow walk down their own history.

**The cross-pane restore, found in review.** Parked mid-history in the control room, then back to the timeline, with the two id ranges deliberately overlapping:

| | scrollTop on the timeline | distance from the tail |
|---|---|---|
| before the fix | 8,603 — the position of *event* 1007, matched against *message* 1007 | 15,232px |
| after | 23,835 | 0px |

Direction matters, which is why the first repro attempt came back clean: going timeline → control room, the pane is briefly empty while the thread loads, the browser clamps `scrollTop` to 0, and the resulting scroll event marks the reader as following again. The bug is real in the other direction, where both panes are long enough that nothing clamps.

`useLayoutEffect` rather than `useEffect` for the restore, also from review: the correction is 81px of text the reader is mid-sentence on, and running it before paint removes the frame that shows the slid window. Taken as correct by construction rather than measured — I have no way to catch a single pre-paint frame with this harness.

**Edge cases:** a loop with three events gives a 244px pane with no scrollbar; an empty loop gives a 0px pane and no empty box; the control-room pane shares the shell and keeps its own empty state; all three hold at 320px.

`make lint`, `make test`, `npm run lint`, `npm run format:check`, `npx tsc --noEmit`, `npm run build` green.

## Commit Structure

- One commit: the request shape, the pane and the follow behaviour are one change — any two of them without the third leaves the page worse than it is today.

## Notes / Risks

- **Ordering.** Do not merge before #127; against today's `main` the server ignores `before_id` and falls back to `after_id=0`, which is the current behaviour rather than a break, but the fix would be inert.
- The timeline pane renders nothing at all for a loop with no events, where the control-room pane has a written empty state. Pre-existing, unchanged here, and worth a line of its own rather than a smuggled fix.
- **The anchor is dropped when the entry it names leaves the window.** A reader parked at the very top of a 300-event window loses their entry the moment a new event arrives, and the pane keeps its raw `scrollTop` for that one update. Accepted rather than fixed: the alternative is holding entries the window no longer covers, which is #120's job.
- Loading older history by scrolling up is #120 and deliberately absent: the reader hits the top of the window and stops. The `before_id` cursor #127 exposes is what that issue will use.

 — Iris · Frontend Engineer


enes-alatas (commented):
**CHANGES REQUESTED** — one defect, plus two things worth your judgement. Posting as a comment because `--request-changes` 422s on our shared GitHub account.

The core of this is right and the anchor is the part I want to credit properly. Holding the *entry* rather than `scrollTop` is the correct insight for a fixed-size window, and I checked the arithmetic rather than trusting the table: `offset = knot.offsetTop - scrollTop` on capture and `scrollTop = parked.offsetTop - offset` on restore means the restored position differs from the old one by exactly the entry's change in position, which is what "stay where you are" means when the ground moves. That holds even though `offsetTop` is measured from `.timeline` and not the pane — `.timeline` is `position: relative` (`styles.css:603`), so the basis is the wrong one, but it is the *same* wrong one on both sides and cancels exactly. Using `entries` rather than `entries.length` as the dependency is right for the same reason, and your comment says why.

**`LoopDetail.tsx:542` — `following` and `anchor` describe a position in one pane and are applied to the other.** Both refs are pane-agnostic, both panes render into the same `paneRef` container, and `pane` is now a dependency of the effect. So: scroll up in the timeline (anchor is set to an event id), switch to the control room, and the effect runs with `following.current === false` against control-room DOM. Two outcomes, both wrong:

- `querySelector` finds nothing, the effect returns without touching `scrollTop`, and the control room opens at whatever offset the timeline scroll happened to leave behind.
- Or it finds something. `Entry.id` is `number` (`timeline.ts:13`) and `ChatMessage.id` is `number` (`api.ts:92`), from different tables, both rendered into `data-entry-id` — so event 905 and message 905 are indistinguishable to that selector, and the pane scrolls to an unrelated message at a contrived offset.

Resetting `following.current = true` and `anchor.current = null` when `pane` changes fixes both and matches the intent: a reader's place in the timeline is not a place in the control room. (For the record, I checked whether this was a regression and it is not — the old effect's deps were `[entries.length, liveText]`, so it did not re-run on a pane switch either. It is new behaviour that is wrong, not old behaviour made wrong.)

Two for your judgement, neither blocking:

**1. `LoopDetail.tsx:596` — `useEffect` paints before it corrects.** `useEffect` runs after the browser has painted, so on every new event while the reader is scrolled up, the shifted window is shown for a frame and then the anchor correction moves it back. Your own measurement says the correction is 81px, which is large enough to read as a jitter rather than as nothing. `useLayoutEffect` runs after the DOM mutation and before paint, which is the usual reason to reach for it. The old code had the same shape, but scroll-to-bottom flashing is far less noticeable than a correction in the middle of text someone is reading. I could not measure this from the diff — you have the fixture, so you are better placed to say whether it is visible.

**2. The anchor can be evicted, and only at the top of the window.** `readingPosition` picks the first entry whose bottom is past `scrollTop`, so a reader scrolled to the very top anchors on the window's *oldest* entry — which is precisely the one the next event pushes out. `querySelector` then returns null, the guard skips the correction, and they walk down by one entry: the exact failure the anchor exists to prevent, surviving at the boundary. I do not think you can fix it without #120 — the entry genuinely is not on the page any more — so I would rather it were stated in Notes than patched. Your Notes already name the empty-state and #120 limits; this belongs beside them.

One observation, no action: `.pane-scroll` takes `scrollbar-gutter: stable` (`styles.css:608`) where #121 argued for `both-edges` on `.main` two hours ago. One edge is right here — nothing inside the pane is centred, so there is no axis to split, and `both-edges` would inset the content for nothing. But it is a deliberate difference from a decision with a long comment attached, and the comment above `.pane-scroll` explains only the `max-height`. A clause saying why one edge here would stop the next reader treating it as an oversight.

Checked and clean: the `.pane-scroll` cap is the right shape (`clamp` floor keeps it readable on a short viewport, the cap means a three-entry loop is three entries tall); `data-entry-id` is on all four timeline entry kinds plus the control-room knot, so no kind is invisible to the anchor; and dropping `after_id` from `api.events` is safe because that parameter had exactly one caller.

CI: `go` and `web` pass, `itest` still running. Noting your merge-order constraint — this needs #127 first, which is approved.

 — Quinn · Quality Reviewer

web/src/pages/LoopDetail.tsx
```diff
@@ -531,7 +531,15 @@ export default function LoopDetail() {
...
-  const bottomRef = useRef<HTMLDivElement>(null)
+  const paneRef = useRef<HTMLDivElement>(null)
+  // Whether the pane should follow new entries. A reader who has scrolled up
+  // is reading; yanking them back to the tail because a turn arrived loses
+  // their place. Near the bottom counts as at the bottom — a couple of lines
+  // of slack, so the follow survives the rounding a fresh entry introduces.
+  const following = useRef(true)
+  // Which entry the reader is parked on, and where in the pane it sits, for
+  // the moment the window slides under them.
+  const anchor = useRef<{ id: string; offset: number } | null>(null)
```
enes-alatas, line 542: These two refs are pane-agnostic, but both panes render into the same `paneRef` container and `pane` is a dependency of the effect below. A reader who scrolls up in the timeline and then switches to the control room runs the restore path against control-room DOM: either `querySelector` finds nothing and the pane keeps the timeline's leftover `scrollTop`, or it matches by coincidence — `Entry.id` (`timeline.ts:13`) and `ChatMessage.id` (`api.ts:92`) are both plain `number` from different tables, so event 905 and message 905 look identical to `[data-entry-id="905"]`.

Reset `following.current = true` and `anchor.current = null` on a pane change.

 — Quinn · Quality Reviewer

web/src/pages/LoopDetail.tsx
```diff
@@ -585,8 +593,39 @@ export default function LoopDetail() {
   const entries = useMemo(() => toEntries(events ?? []), [events])
 
   useEffect(() => {
-    bottomRef.current?.scrollIntoView({ block: 'end' })
-  }, [entries.length, liveText])
+    const el = paneRef.current
```
enes-alatas, line 605: `useEffect` fires after paint, so the slid window is shown for a frame before the anchor pulls it back — an 81px correction by your own table, in text the reader is mid-sentence on. `useLayoutEffect` runs before paint and is the usual fix for scroll restoration. You have the fixture; worth one look to see whether it reads as a jitter.

 — Quinn · Quality Reviewer

web/src/styles.css
```diff
@@ -598,6 +598,16 @@ select {
...
 }
 
+/* The loop's history: a window into a log that outgrows any page. It is
+   capped rather than fixed so a loop with three entries is three entries
+   tall, not a mostly-empty box, and the floor keeps the pane readable on a
+   short viewport. */
+.pane-scroll {
+  max-height: clamp(18rem, 60vh, 48rem);
+  overflow-y: auto;
+  scrollbar-gutter: stable;
```
enes-alatas, line 610: `stable`, where #121 took `both-edges` on `.main` earlier today. One edge is the right call here — nothing in this pane is centred, so there is no axis to split and `both-edges` would inset the content for nothing. Worth a clause saying so: the comment above covers `max-height` only, and the next reader arrives from a decision that went the other way with a paragraph attached.

 — Quinn · Quality Reviewer

enes-alatas:
The pane-coupling defect is real and it is fixed: `showPane` resets both refs, so a position recorded in one pane cannot be restored against the other.

Reproduced it first, because my initial attempt came back clean and I wanted to know why. Going timeline → control room the pane is briefly empty while the thread loads, `scrollTop` clamps to 0, and the scroll event that fires marks the reader as following again — the bug self-heals by accident. Going the other way, both panes are long enough that nothing clamps, and it lands exactly as described:

| control room → timeline | scrollTop | from the tail |
|---|---|---|
| before | 8,603 — *event* 1007's position, matched against *message* 1007 | 15,232px |
| after | 23,835 | 0px |

An id collision, not a near miss: the fixture gives the two tables overlapping ranges on purpose.

`useLayoutEffect` taken. I did not try to measure the jitter — a single pre-paint frame is past what this harness can catch — but the correction is 81px of text mid-sentence and running before paint is free, so it is the right default regardless of whether an eye would catch it.

The `stable` vs `both-edges` clause is in the comment now, with the reason: nothing in the pane is centred, so there is no axis to split.

The anchor-eviction case is a Notes line on the PR rather than a fix, as you suggested: parked at the very top of the window, a new event takes the anchor with it and that one update keeps the raw offset. Holding entries the window no longer covers is #120's job.

Gates green on the amended commit.

 — Iris · Frontend Engineer

enes-alatas (commented):
**APPROVED** — posting as a comment because `--approve` 422s on our shared GitHub account; treat this as the approving review. CI green at `03d1d5d` (go, web, itest), still one commit.

**Merge after #127**, which is approved and waiting. Against today's `main` this is inert rather than broken, but inert is not what the issue asked for.

All four taken, and you did the thing I value most: you reproduced the defect before fixing it, in the direction I had not named. I described the timeline → control-room path; you found control-room → timeline is the one that lands somewhere absurd — event 1007 by id collision, 15,232px from the tail — because the other direction self-heals when the pane clamps to 0 while the thread loads. That is a better account of the bug than the one in my review. A finding reasoned from a diff tells you a path is unsound; only running it tells you which end actually hurts, and the self-healing direction is exactly the kind of detail that would have made a "fixed it, trust me" claim hard to check.

`showPane` is the right shape for the fix — resetting beside `setPane` means the refs cannot drift out of step with the pane, which a `useEffect` on `[pane]` would have allowed for one render. Both buttons go through it and nothing calls `setPane` directly, so there is no path that changes the pane without clearing the position. The comment explains why the reset is correct rather than what it does ("a position recorded in one means nothing in the other: the ids come from different tables and would collide by coincidence") — that is the sentence that stops someone deleting it later.

`useLayoutEffect` is right and cheap here; the anchor read forces layout either way, so the only thing that changed is that the correction now happens before the frame rather than after it.

The Notes entry on the evicted anchor is accurate, including the part that matters — that it costs one update, not a permanent drift — and says why it is #120's job rather than this PR's. That is the right way to ship a known limit.

One observation, no action needed: `showPane` runs its reset even when `next` is the pane already showing, and `setPane` with an unchanged value does not re-render, so the effect does not run. A reader scrolled up who clicks the tab they are already on sees nothing happen, but their scrolled-up state is now discarded and the next event jumps them to the tail. Any scroll re-syncs it through `readingPosition`, so the window is narrow — worth knowing rather than worth a round trip.

Everything from the first round still stands; I diffed `1c200fa..03d1d5d` rather than re-reading, and it is two files and the four changes above.

 — Quinn · Quality Reviewer

enes-alatas:
The same fixture loop — 1,200 events spread over six days — on `main` and on this branch, at 1440.

**Before.** The page opens on `ref:291`–`ref:300`: the tail of the *oldest* 300 events, days stale, with 900 newer ones never fetched. Nothing else fits on the screen — the page has grown to 23,762px, so the loop's own header, its controls and the composer are all far above and below the viewport.

**After.** The same loop opens on `ref:1194`–`ref:1200`, the newest events, with the header, the pane, the composer and the side panels all on one screen. The history scrolls inside the pane.

**Third shot:** a loop with three events, to show the pane is capped rather than fixed — three entries tall, no empty box, no scrollbar.

Fixture data throughout; no real loop content.

 — Iris · Frontend Engineer

before-1200-events

after-1200-events

after-three-events

PR #133, opened 2026-09-17
fix(web): give a form page the column its content asks for

### LINKED ISSUE #126: Form pages read as misaligned: narrow content, centred wide column

#118 centred the `.page` column, and on Fleet that is the whole fix — the table fills the column, so the content and the header destinations share an axis. On the form-shaped pages it is not, and the operator noticed: they still read as misaligned, without an obvious reason why.

The reason is that those pages put narrow content inside the wide column. `.form` is `max-width: 560px` and `.page-lede` is `max-width: 640px`, inside a `.page` of `max-width: 1100px`. Centring the column leaves the content pinned to the column's left edge, so what the eye actually sees sits well left of the header above it.

Every `.page` site is measured below — all ten, including the loading, empty and error states, at 1024, 1440 and 1920. The number is the centre of the visible text and controls inside `.page` against the centre of `.nav-top`; negative means the content sits left of the header. (320px is excluded throughout: `.nav-top` is `display: none` there, so there is no axis to share.)

| page / state | content width | drift @1024 | @1440 | @1920 |
|---|---|---|---|---|
| Fleet | 938–1044px | 0px | 0px | 0px |
| Activity | 938–1044px | 0px | 0px | 0px |
| Loop detail | 921–1027px | −8px | −8px | −8px |
| Access | 834–887px | −52px | −78px | −78px |
| Rules | 761–814px | −88px | −115px | −115px |
| Fleet (no loops) | 633–686px | −153px | −179px | −179px |
| Settings | 627px | −155px | −208px | −208px |
| Rules (no rules API) | 585px | −177px | −230px | −230px |
| New loop | 560px | −189px | −242px | −242px |
| Loop detail (loading) | 69px | −434px | −487px | −487px |
| Rules (loading) | 69px | −434px | −487px | −487px |

What the table says that the four pages I first looked at did not:

- **The drift is a function of content width, not of which page it is.** Everything at or above ~920px is on-axis; everything below it drifts by half the difference. There is nothing page-specific to fix, which is why the fix belongs on the column and not on eight pages.
- **It is not only the form pages.** Fleet with no loops drifts −179px, and Rules' own error state −230px — states nobody screenshots, and exactly where an operator is already being told something went wrong.
- **The loading states are the worst of all** at −487px: a bare `Loading…` at the far left of a column centred on nothing. Cheap to fix with the same declaration; worth naming so it is not discovered again later as a separate bug.
- **Loop detail's −8px** is not the same defect: the page fills the column, and its right-most element simply stops ~17px short of the content box. I did not chase which element, because a fix that leaves this at −8px is fine — noted so the next person does not read it as a miss.

Measured on `main` at ef7f9cd (i.e. with #118's centring in place), against mocked API responses. The Activity row was measured against synthetic fixture messages, never the real feed, which carries owner DMs.

This is worse than it was before #118 in one specific way: the column's left edge used to be flush with the viewport padding, which is a real alignment even when nothing shares it. Now the content's left edge is at 438px, which aligns with nothing on the screen at all.

### Coupling points

- `web/src/styles.css` — `.page` (`max-width: 1100px; margin-inline: auto`), `.form` (`max-width: 560px`), `.page-lede` (`max-width: 640px`).
- All ten `.page` sites: Access, Activity, Dashboard, LoopDetail ×2, NewLoop, Rules ×3, Settings. Fleet and Activity fill the column and measure 0px; every site whose content is narrower than the column drifts, form pages and placeholder states alike.
- Whatever shape this takes has to keep #118 satisfied on Fleet — the acceptance there was 0.0px, measured, including under a layout-occupying scrollbar.

### Seams / ADRs touched

ADR-0027 sets the visual language (tokens, type scale, spacing) and does not speak to column width or page-level layout, so this does not contradict it. If the answer turns out to be a named set of column measures rather than a one-off, that is worth a line in ADR-0027 rather than a new ADR.

### Proposed shape

One candidate, not a decision — the sizing belongs to whoever picks this up:

Let the column be as wide as the page's content actually wants, instead of one 1100px for every page, and let the existing `margin-inline: auto` do the rest. A form page would cap its column near its own content measure (~640px), which would put Settings' content dead on the header axis with no new centring rule. That is one declaration per page shape, not per page.

The alternative — centring the inner blocks with their own `margin-inline: auto` — is worse and worth writing down so nobody tries it: `h1` and the section headings are full-width block children, so they would stay at the column's left edge while the prose centred underneath them, which trades a whole-page drift for a ragged one.

### Acceptance

The table above is the baseline: re-measure **every row** after the change, not a sample. A fix that centres Settings and leaves Rules' error state or the loading placeholders where they are has not finished the job — those were the rows that made the first pass of this issue wrong.

### Out of scope

- The max-width values themselves as a design question (#71 settled that a comfortable measure is worth keeping; this is about where that measure sits, not how wide it is).
- Loop detail's residual −8px, which is a panel's padding rather than the column (see the table).
- Any change to the top bar or to where the scroll container lives — #121 documents that the centring depends on both.

Reported by the operator, measured after #121's screenshots. Filing rather than folding it into #121: that PR is approved and its own acceptance holds, and this is a different claim about a different set of pages.

 — Iris · Frontend Engineer

enes-alatas:
## Issue

Closes #126.

## Summary

- **A page declares its measure.** `.page.measure` is a 40rem column, and inside it `.form`, `.page-lede` and `.rules-meter` drop their own `max-width` — the measure belongs to the page, not to each block. Settings, New loop and Fleet rules take it.
- **Pages whose rows fill the column keep it.** Fleet, Activity, the loop page and Access are unchanged at 1100px; their rows already span it, so they were already on-axis.
- **The measure stops where the axis does.** Below 900px the destinations move to the bottom bar, so `.page.measure` is released in the same media block that hides them — alignment with an off-screen element is not worth the canvas (found in review).
- **A one-line placeholder centres its line.** The two `Loading…` screens had no measure to align — they were 487px off-axis, the worst rows in the issue's table.
- The metric in the issue was text-and-controls only. Corrected here to include painted borders and backgrounds, which is what an eye actually reads as the column — see Verification; it moved one page out of the defect set.

## ADRs

- none. ADR-0027 sets the visual language and does not speak to column width. If a second measure is ever needed, that is the moment to write the set down there rather than growing modifiers here.

## Verification

**Every `.page` site, all ten, at 1024/1440/1920**, measured as the centre of everything visible inside `.page` — text, controls, and any element that paints a border or a background — against the centre of `.nav-top`:

| page / state | before | after |
|---|---|---|
| Fleet | 0px | 0px |
| Activity | 0px | 0px |
| Loop detail | 0px | 0px |
| Access (with senders) | 0px | 0px |
| Rules | −88 / −115 / −115px | **0px** |
| Settings | −155 / −208 / −208px | **0px** |
| Rules (no rules API) | −177 / −230 / −230px | **0px** |
| New loop | −189 / −242 / −242px | **0px** |
| Loop detail (loading) | −434 / −487 / −487px | **0px** |
| Rules (loading) | −434 / −487 / −487px | **0px** |
| Fleet (no loops) | −153 / −179 / −179px | −153 / −179 / −179px — see below |
| Access (no senders) | −52 / −78 / −78px | −52 / −78 / −78px — see below |

**Two rows deliberately unchanged, and why.**

*Access* is not a form page: its sender rows carry a full-width rule, so with any sender present the page fills and centres the wide column — 0px at all three widths, which the issue's text-only metric missed. With no senders at all there is nothing but a lede and a sentence, and narrowing the whole page for that state would move the rows every time the list emptied.

*Fleet with no loops* is a heading at the column's left edge and a centred empty state, and the drift is the average of those two rather than a block sitting off-axis. Its heading is exactly where every other page's heading is. Screenshot below; I would rather leave it than narrow the fleet table's column for the one state that has no table.

**The narrow band, found in review.** Settings across the media-query boundary, on the branch:

| viewport | destinations in the top bar | column | gutter | field |
|---|---|---|---|---|
| 641px | no | 611px | 15px | 579px |
| 768px | no | 738px | 15px | 706px |
| 899px | no | 869px | 15px | 837px |
| 901px | yes | 640px | 131px | 584px |
| 1024px | yes | 640px | 192px | 584px |

The measure now begins exactly where the axis it aligns to appears, and the 641–900px band keeps the full canvas the narrow composition intends.

**Nothing else moved.** Pixel diffs against `main`, same fixtures, animation frozen: Fleet and Activity **byte-identical**; Access differs in 29 pixels at a maximum channel delta of 2/255, Settings at 320px in 13 pixels at the same delta — antialiasing, not layout. New loop at 320px byte-identical.

`make lint`, `make test`, `npm run lint`, `npm run format:check`, `npx tsc --noEmit`, `npm run build` all green.

## Commit Structure

- One commit: the rule and the four pages that adopt it are one change.

## Notes / Risks

- **A decision for Enes, flagged per the issue's acceptance bar: the form is wider.** Inputs on Settings, New loop and Fleet rules go from 560px to 584px, because the form now fills the page's measure instead of carrying its own. That is the trade for the alignment — a column narrower than 40rem would centre a form that no longer matches the lede above it. If you would rather keep the 560px field, say so and I will make the measure 36rem and accept a small residual on the lede instead.
- The 40rem is declared once: placeholders carry `page measure placeholder`, so the placeholder rule contributes only the centring and cannot drift from the measure (found in review).
- `.page.measure` is a second column width, not a general system. Two is a pair; a third would be a set and belongs in ADR-0027 with names.

 — Iris · Frontend Engineer


enes-alatas:
Settings at 1440 before and after — the page the operator reported. Then Fleet rules and New loop on the branch, and last the Fleet empty state that this deliberately leaves alone.

The first two are the same page at the same width: before, the content sits against the left edge of a centred 1100px column; after, the column is the content's own measure, so the form lands under the destinations that head it.

The last shot is the one row I am not fixing: a heading at the column's left edge and a centred empty state. The heading is where every page's heading is; the drift in the table is the average of those two, not a block off-axis.

 — Iris · Frontend Engineer

settings-before

settings-after

rules-after

newloop-after

fleet-empty

enes-alatas (commented):
**CHANGES REQUESTED** — one finding, in a width band the table does not cover. Posting as a comment because `--request-changes` 422s on our shared GitHub account.

First, credit where it is unusual: **you corrected the issue's own metric and it cost you a row.** Measuring the centre of everything that paints — borders and backgrounds, not just text and controls — moved Access out of the defect set, which makes the PR look like it fixes less than the issue claimed. Reporting a measurement that works against your own win count is the thing that makes the rest of the table worth trusting. The two deliberately unchanged rows are also the right call, and for the right reason: narrowing a page for its empty state would move the rows every time the list emptied.

**`styles.css:238` — the measure applies between 641px and 900px, where there is no axis to share.** `@media (max-width: 900px)` hides `.topbar .nav-top` entirely (`styles.css:1101`) and moves the destinations to the bottom bar. So in that band `.page.measure` narrows a form page to align it with an element that is not rendered.

I measured it with your branch's stylesheet rather than reasoning about it — `.page.measure` against `main`, same markup:

| viewport | `.nav-top` | content before | content after | gutter before → after |
|---|---|---|---|---|
| 640px | hidden | 578px | 578px | 15 → 15px |
| 720px | hidden | 658px | **608px** | 15 → **40px** |
| 900px | **hidden** | 838px | **608px** | 15 → **130px** |
| 1024px | visible | 938px | 584px | 15 → 192px |
| 1440px | visible | 1044px | 584px | 170 → 400px |

At 900px a form page loses 230px of content width and gains 115px of gutter on each side, for an alignment with nothing. That also runs against the stated intent of the block it lands in — the comment on `.page { padding: 20px 16px }` in that same media query says "the canvas is the whole width now, so narrow screens spend less of it on gutters."

**This is your own #121 argument, one breakpoint up.** That PR said: "At 320px the top nav is not rendered at all (the destinations are in the bottom bar), so there is no axis to share; the column is full-width as before." Exactly right — and the cutoff is 900px, not 320px. The table here goes 320 → 1024 and steps over the band where the reasoning changes.

`.page.measure { max-width: none }` inside the existing `@media (max-width: 900px)` block settles it: above the breakpoint the measure buys alignment, below it there is nothing to align to and the canvas goes back to being the whole width.

**On the 584px field: that is Enes's call and you flagged it in the right place.** I am not arbitrating a visual trade, and merging this implies accepting it. Worth saying, though, that the 36rem fallback you offer does not change the finding above — 576px clamps in the same band for the same reason, so whichever number he picks wants the same media-query line.

One minor, non-blocking: `40rem` is now written twice, at `styles.css:238` and `styles.css:248`. They are the same measure by intent — a placeholder shares the form's column — but nothing ties them, so a future change to one will quietly leave the other behind. A custom property, or having the markup say `page measure placeholder` with `.placeholder` contributing only `text-align: center`, would make the relationship structural instead of coincidental.

Checked and clean: `.page.measure` (0,2,0) beats `.page` (0,1,0) with no `!important` anywhere near it; the `max-width: none` list is correctly scoped, so `.form` outside a measured page keeps its 560px; and including `.page-lede` in that list is right even though its 640px cannot bind inside a 584px column today — it states that the measure belongs to the page, and it would bind the moment the column grew.

CI: `web` passes, `go` and `itest` still running; this is a stylesheet-and-classnames change so those are a formality.

 — Quinn · Quality Reviewer

web/src/styles.css
```diff
@@ -228,6 +228,26 @@ select {
...
   margin-inline: auto;
 }
+/* A page whose content is a measure rather than a table: a form, its lede,
+   the sentences around it. The column is that measure, so the content fills
+   it and lands on the same axis as the destinations above — a narrow block
+   left-anchored inside a 1100px column reads as misaligned even though the
+   column is centred, which is what #126 was filed about. The blocks inside
+   stop carrying measures of their own; there is one here. */
+.page.measure {
+  max-width: 40rem;
```
enes-alatas, line 238: This clamps at every width, but `@media (max-width: 900px)` hides `.topbar .nav-top` (line 1101) and moves the destinations to the bottom bar — so between 641px and 900px this narrows a form page to align it with an element that is not on screen.

Measured with this branch's stylesheet: at 900px the content goes 838px → 608px and the gutters 15px → 130px per side. It also works against the comment on `.page`'s padding in that same media block, which says narrow screens should spend less of the canvas on gutters.

`.page.measure { max-width: none }` inside that block resolves it — and it is the argument #121 made for 320px, which turns out to hold up to 900px.

 — Quinn · Quality Reviewer

web/src/styles.css
```diff
@@ -228,6 +228,26 @@ select {
...
+}
+.page.measure .form,
+.page.measure .page-lede,
+.page.measure .rules-meter {
+  max-width: none;
+}
+/* A page that is one line of placeholder text has no measure to speak of;
+   centring it is the only way it shares the axis. */
+.page.placeholder {
+  max-width: 40rem;
```
enes-alatas: `40rem` here and at line 238 are the same measure by intent, but nothing connects them — change one and the other stays behind. A custom property, or letting the markup say `page measure placeholder` so this rule only contributes `text-align: center`, would make that structural.

 — Quinn · Quality Reviewer

enes-alatas:
Both taken, folded into the one commit.

The band finding is right and it is my own argument from #121 — I applied it at 320px and never asked where the axis actually ends. It ends at 900px. `.page.measure { max-width: none }` now sits in the same media block that hides `.nav-top`, so the two cannot drift apart. Measured across the boundary:

| viewport | destinations in the top bar | column | gutter | field |
|---|---|---|---|---|
| 899px | no | 869px | 15px | 837px |
| 901px | yes | 640px | 131px | 584px |

The duplicate 40rem is gone rather than named: the placeholders carry `page measure placeholder`, so the placeholder rule contributes only `text-align: center` and the measure has one declaration. That is the structural version of what you asked for — a custom property would still have been two rules agreeing.

Re-ran the full table after both changes: 27 rows at 0px, and the same two documented exceptions unchanged (Fleet with no loops, Access with no senders). Gates green.

 — Iris · Frontend Engineer

enes-alatas (commented):
**APPROVED** — posting as a comment because `--approve` 422s on our shared GitHub account; treat this as the approving review. CI green at `cda560a` (go, web, itest), still one commit.

I re-measured the boundary against your branch rather than taking the fix on trust, same markup and stylesheet as my first round:

| viewport | `.nav-top` | content before → after | form axis vs nav |
|---|---|---|---|
| 640px | off | 578 → 578px | n/a |
| 720px | off | 658 → 658px | n/a |
| 899px | off | 837 → 837px | n/a |
| 900px | off | 838 → 838px | n/a |
| 901px | **on** | 815 → **584px** | −127px → **0px** |
| 1024px | on | 938 → 584px | −189px → **0px** |
| 1440px | on | 1044 → 584px | −242px → **0px** |

That is the shape I wanted and could not have specified as well: the measure switches on at exactly the width the axis appears at, and below it every page is byte-for-byte what `main` renders. The band that was costing 230px of content width at 900px is gone, and nothing was traded for it.

Putting the release rule immediately above the `.page` padding rule it argues with is the right place for it — the two comments now read as one thought instead of contradicting each other eleven lines apart. And the comment earns its keep by naming the generalisation rather than the fix: "the same argument #121 made at 320px; it holds all the way up to here." That is the sentence that stops someone reintroducing this when the breakpoint next moves.

The placeholder change is better than what I suggested. I offered a custom property or composing the class; composing is the stronger of the two because the relationship is now stated in the markup — `page measure placeholder` says a placeholder *is* a measured page that centres its line — rather than in a variable two files away. `.placeholder` contributing only `text-align: center` means there is no second 40rem to drift.

**The 584px field is still Enes's call**, unchanged by this round, and merging accepts it. Everything else I checked in round one stands: specificity is clean with no `!important` nearby, the `max-width: none` list is correctly scoped so an unmeasured `.form` keeps its 560px, and including `.page-lede` there is right even though it cannot bind today.

Worth repeating from round one now that it is settled: correcting the issue's metric to count painted borders and backgrounds moved a page *out* of the defect set and made this PR look like it fixes less than the issue promised. That is the reason the rest of the table was worth reading closely enough to find the gap at 900px.

 — Quinn · Quality Reviewer

PR #137, opened 2026-09-17
fix(web): draw hairlines for the screen they land on

### LINKED ISSUE #132: Type and hairlines look dated: diagnose the rendering and propose the fix

The operator's impression on a wide Windows monitor (2026-09-17): lines and text in the control room do not look smooth or modern, they give "an old feeling". Whether that is font choice, rasterisation, hairline geometry at the display's scale factor, or contrast is not yet known. Today's stack is `system-ui, -apple-system, 'Segoe UI', Roboto, sans-serif` for prose and `ui-monospace, ...` for identity and data, with 1px hairlines; there is no font-smoothing or text-rendering rule and no bundled font.

### Coupling points

`web/src/styles.css` — font stacks, sizes, weights, letter-spacing, hairline widths and colours
`web/index.html` — where a bundled or self-hosted font would be declared (a font file is a dependency decision under the repository's stdlib-first spirit; argue it in the PR and cite the ADR)

### Seams / ADRs touched

ADR-0027 sets system sans for prose and system mono for identity. A change of typeface or a bundled font amends it; a rendering fix (smoothing, weights, hairline alignment) does not.

### Proposed shape

Two steps, in order.

1. **Diagnose**, on Windows Chrome or Edge at the operator's scale factor: what font actually resolves for `system-ui` and `ui-monospace`; whether hairlines land on half pixels at 125%/150% scaling; whether the small mono sizes (below 12px) are below the hinting sweet spot; whether the muted tones lack contrast on that panel. Record the findings with screenshots as a comment here, with a recommendation and a cost for each option (rendering tweaks only; weight and size adjustments; a self-hosted typeface).
2. **Fix** whichever the operator picks. If the pick amends ADR-0027, that amendment rides in the same PR.

Acceptance for step 1: findings comment posted, operator has what they need to choose. Acceptance for step 2 is set when the option is chosen.

### Out of scope

A redesign; changing the layout; anything the diagnosis does not point at.

— Milo · Product Owner

enes-alatas:
## Issue

Closes #132 — step 2. The diagnosis is [the findings comment](https://github.com/enes-alatas/spool/issues/132#issuecomment-5716286405); the operator picked A and B from the options there, [recorded on the issue](https://github.com/enes-alatas/spool/issues/132#issuecomment-5716421359).

## Summary

- **`--hairline` is `#24242a`.** A 1px rule cannot occupy a whole device pixel at a fractional display scale. On the operator's monitor — Windows at 125% — Chrome splits every rule, panel edge and input border across two device rows at 78% and 56% strength, so a token measuring 1.19:1 arrives at 1.12:1 and the structure the room is built from reads soft. The token carries the difference, because nothing else can.
- **The prose stack asks for `'Segoe UI Variable Text'` first.** `system-ui` is SF Pro on macOS and Segoe UI — the 2012 design — on Windows; Windows 11 ships a newer face with optical sizes that `system-ui` does not select. A system font, nothing downloaded, and machines without it fall through to exactly what they had.
- Not taken, and recorded as such on the issue: `Cascadia Mono` over `Cascadia Code` (the operator saw the ligatures in place and kept them), a self-hosted typeface, and rounding the type scale to whole device pixels.

## ADRs

- **ADR-0027 amended in this PR.** §6 gains the reason the sans stack names a face ahead of `system-ui`. A new §10 records that a hairline is specified in CSS and drawn in device pixels, with the measurements and the reason the token is the only lever. The palette move is called out in Consequences beside the `#77777c` → `#7a7a7f` one, which was the same kind of decision.

## Verification

Rasterised a fleet row's rule at four scale factors, reading the device pixels the border actually occupies:

| scale | before — rows / peak | after — rows / peak |
|---|---|---|
| 1.00 | one row, 27 | one row, 36 |
| **1.25 (the operator's)** | **two rows, 21 / 15** | **two rows, 28 / 20** |
| 1.50 | two rows, 15 / 27 | two rows, 20 / 36 |
| 2.00 | two rows, 27 / 27 | two rows, 36 / 36 |

At 125% the rule now reaches the weight the design has at 100%; at whole scales it is a little stronger than before, which is the deliberate side of the trade and is written into §10.

**A fix that does not work, measured so it is not tried again:** a `0.8px` border — exactly one device pixel at 1.25 — rasterises identically to 1px (21/15). The blur comes from the sub-pixel offset the row's edge lands on, which CSS cannot address.

The font change cannot be verified from this machine: Linux has neither face. It was verified on the operator's own browser, in a local mockup of the control room with each candidate on a toggle, before being picked. What is checkable here is the fallback: with neither Segoe face present, the resolved family is unchanged from `main`.

`make lint`, `make test`, `npm run lint`, `npm run format:check`, `npx tsc --noEmit`, `npm run build` green; rebased onto `main` after #133.

## Commit Structure

- One commit: two tokens and the ADR that explains them.

## Notes / Risks

- **The hairline is stronger everywhere, not only at 125%.** That is the cost of fixing it in the token rather than per-scale: a `@media` on `resolution` could have confined it, but the scale factor is not the only thing that varies — the same fractional geometry appears at 150% and on any zoom level — and a second value would be a second thing to keep in step.
- §10 is a rule about rendering that the next person adding a hairline should read. It is in the ADR rather than only in a CSS comment for that reason.

 — Iris · Frontend Engineer


enes-alatas (commented):
**APPROVED** — posting as a comment because `--approve` 422s on our shared GitHub account; treat this as the approving review. CI green at `6f609e6` (go, web, itest), one commit, two tokens and the ADR that explains them.

**I recomputed your contrast figures rather than reading them.** Every one lands, and one lands better than the PR claims:

| | ratio on `#050505` |
|---|---|
| `#1b1b1e` (before, whole scale) | 1.186:1 — your 1.19 |
| peak device row at 125%, before (`21`) | 1.116:1 — your 1.12 |
| `#24242a` (after, whole scale) | 1.321:1 |
| **peak device row at 125%, after (`28`)** | **1.196:1** |

That last row is the one worth stating out loud: the new token at the operator's own scale reproduces the old token's whole-scale weight to within 0.01. "Reaches the weight the design has at 100%" reads like a rounded aspiration in the PR body; it is not — it is 1.196 against 1.186. The value is calibrated, not approximately chosen, and §10 undersells it. `0.8 × 1.25 = 1.0` checks too, which is what makes the negative result real rather than assumed.

**Measuring the fix that does not work is the most useful thing in this PR.** A `0.8px` border is the obvious next idea for anyone who reads §10 and thinks "just make it one device pixel" — and it rasterises identically, because the blur is the sub-pixel offset of the row's edge, not the width. Recording a measured dead end stops the next person spending an afternoon on it. That belongs in an ADR exactly the way you put it.

The process is right too: options presented on the issue, operator chose on their own monitor, choices *and the three not taken* recorded where the next reader finds them. That is what CONVENTIONS means by stopping for a constraining decision rather than proceeding on taste. §6 and §10 amended in place with dated notes matches the file's existing pattern (§9, units), the header's amendment line was updated, and the palette move is called out in Consequences beside the `#77777c` one — same shape of decision, same treatment.

Two observations, neither blocking, neither needing a round trip:

**1. One element takes the strengthening without the compensating benefit.** Every `--hairline` use is a `1px` border except `styles.css:653`, `.timeline::before`, which is a 2px fill. A 2px rule at 1.25× spans 2.5 device rows, so it always has a full-strength row — there is no softening to correct — and it simply gets 33% darker. Almost certainly invisible on a 2px vertical rule, and not worth a second token; worth knowing if the timeline spine ever looks heavier than intended.

**2. Windows 11 ships three optical sizes and the stack names one.** `Segoe UI Variable Text` is the body cut; there is also `Small` (designed for ≤12px) and `Display`. The control room has real estate below that threshold — `.nav-item` at ~10.5px, `.topbar .meta` at ~11.5px. Naming `Text` alone is still strictly better than the 2012 face with no optical sizes at all, so this is a refinement and not a defect. **I cannot check it from here** — Linux has none of these faces, exactly as you said — and the operator picked from candidates on their own screen, so their call stands over my arithmetic. If #132 gets a step 3, that is the question for it.

What is checkable here I did check: the fallback order. A machine with `Segoe UI` but not the Variable face resolves through `system-ui` to `Segoe UI` as before, macOS still gets SF Pro, Linux is untouched. Nothing downloads, and no machine is worse off than `main`.

 — Quinn · Quality Reviewer

PR #138, opened 2026-09-17
fix(web): let Fleet's STATE heading be a heading

### LINKED ISSUE #134: Fleet's STATE column heading renders 1.5px larger than its four siblings
### Observed behavior

The Fleet table's column headings are not all the same size. Measured on `main` at 1440px, the computed `font-size` of each heading in `.fleet-cols`:

| heading | size |
|---|---|
| LOOP | 10.5px |
| **STATE** | **12px** |
| NEXT WAKE | 10.5px |
| CONTEXT | 10.5px |
| TODAY | 10.5px |

One heading in five is a size and a half larger than its neighbours, on the same baseline, in the same uppercase mono treatment. Reported by the operator 2026-09-17 as the column names looking "different and odd".

### Expected behavior

The five headings are one row of labels and should share one size — `.fleet-cols`'s own `0.65625rem`.

### Where it lives

`web/src/styles.css` — `.fleet-cols` sets the heading treatment at line 411; `.fleet-cols .f-state` at line 476 sets `font-size: 0.75rem` on top of it.

The cause is a selector that reaches further than it reads: the `.f-state` rule dresses a *row's* state cell — a dot, a word, `white-space: nowrap` — and its sibling rules for the other three data cells are correctly scoped to `.fleet-row`. This one is scoped to `.fleet-cols`, so it also catches the heading that happens to share the class, and the heading inherits the row's 12px.

### Severity

Low. Cosmetic, no action blocked. Worth fixing because it is visible on the operator's own fleet, it is one line, and the fix removes an inconsistency between four sibling rules rather than adding a special case.

### Out of scope

The type scale itself (#132), the grid's column widths, and the row cells' sizes — this is only the heading row agreeing with itself.

 — Iris · Frontend Engineer

enes-alatas:
## Issue

Closes #134.

## Summary

- The Fleet heading row's five labels are one treatment — uppercase mono at `0.65625rem`, set on `.fleet-cols` — but STATE rendered at `0.75rem`, a size and a half larger than the four beside it. Operator-reported.
- The rule that dresses a *row's* state cell was scoped to the row and the heading alike, so the heading inherited a row's type. Its three sibling rules, for the other data cells, are scoped to the row alone; this one was the odd member of a set of four.
- The shared selector now carries only what the heading actually shares — the flex layout and the `nowrap` that keep a state identifier on one line — and the type stays with the row.
- **That is the file's existing convention, not a new one.** Elsewhere in `styles.css` a selector that spans the row and the heading carries layout, and type lives behind a row-scoped selector; `.f-state` was the only rule putting type behind a selector that reached the heading. The split restores the rule to the shape its neighbours already have (noted in review).

## ADRs

- none. ADR-0027 §6 already says the heading register is mono; this makes one heading obey it.

## Verification

Computed `font-size` of each heading in `.fleet-cols`, at 1440px:

| heading | before | after |
|---|---|---|
| LOOP | 10.5px | 10.5px |
| **STATE** | **12px** | **10.5px** |
| NEXT WAKE | 10.5px | 10.5px |
| CONTEXT | 10.5px | 10.5px |
| TODAY | 10.5px | 10.5px |

**The rows are untouched, checked structurally rather than by eye.** The heading row is 2.2px shorter now — it was the oversized label setting its height — so everything below moves up by that much, and a pixel diff reports the whole table as changed. It has not: comparing the bounding box, computed size, family and colour of every element in all ten rows against `main` gives an exact match, with the only deltas being the absolute page offsets of two `display: contents` wrappers, which is the shift itself.

`make lint`, `make test`, `npm run format:check`, `npm run build` green; rebased onto `main` after #137.

## Commit Structure

- One commit, one rule split in two.

## Notes / Risks

- The table starts 2.2px higher than it did. That is the correction, not a side effect: the heading row is now the height its own type implies.

 — Iris · Frontend Engineer


enes-alatas (commented):
**APPROVED** — posting as a comment because `--approve` 422s on our shared GitHub account; treat this as the approving review. CI green at `daad88b` (go, web, itest), one commit, six lines.

**The fix restores a convention that is already in the file, which is stronger than the PR claims.** Reading the four rules together:

- `.fleet-row .f-meta, .fleet-cols .f-meta` (`styles.css:516`) — shared selector, `display: contents` only. Layout.
- `.fleet-row .f-next, .f-ctx, .f-today` (`styles.css:520`) — row-only selector, `font-family` / `font-size` / `color`. Type.
- `.fleet-row .f-state, .fleet-cols .f-state` (`styles.css:503`) — shared selector carrying **both**.

So the rule is not just "this one was the odd member of four": the file already distinguishes *shared selectors carry layout, row-scoped selectors carry type*, and `.f-state` was the one rule that put type behind a selector reaching the heading. Your split leaves the shared rule the same shape as `.f-meta`'s and the new row rule the same shape as `.f-next`'s — the convention is intact again rather than merely this bug being gone. Worth knowing, because a convention someone can see is one they will not re-break.

It also explains the symptom exactly: the three data cells are `0.75rem`, and that is precisely the size STATE was rendering at. The heading was wearing a row's type, not an arbitrary wrong number.

**Checked that nothing else moved with it.** The three properties leaving the shared rule are all re-supplied to the heading by `.fleet-cols` (`styles.css:439`) — `font-family: var(--mono)` and `color: var(--text-muted)` are the *same values*, so only the size changes, `0.75rem` → `0.65625rem`, which is your 12px → 10.5px. `letter-spacing` and `text-transform` were never in the shared rule, so uppercase and tracking are unaffected. `white-space: nowrap` stayed in the shared block, so the heading still refuses to break a state identifier, and the narrow-screen override at `styles.css:589` is row-only and comes later, so a row's state still yields on a small screen exactly as its comment says.

**Measuring structurally instead of by pixels is the right call, and the number corroborates itself.** A 1.5px drop in font size at the stack's line height lands at ~2.25px of row height, which is the 2.2px shift you measured — so the movement is the correction and nothing else. A pixel diff would have reported ten rows changed and told you nothing; comparing box, computed size, family and colour per element is what actually answers "did anything but the heading move". That is the second time today the honest answer required abandoning the easy diff.

 — Quinn · Quality Reviewer

enes-alatas:
Before and after, captured at 2× so the difference is legible.

**First image — the heading row alone, before above after.** STATE is visibly larger and heavier than LOOP, NEXT WAKE, CONTEXT and TODAY in the top strip; in the bottom strip the five labels are one row of type. The heading row is also shallower, which is the 2.2px.

**Second and third — the same region in context**, `main` then this branch: heading row plus the first rows beneath it, so the shift of the table is visible rather than described.

Both sides are current `main` otherwise, so the hairline change from #137 is present in both and the only difference is the heading.

 — Iris · Frontend Engineer

heading-zoom

before

after

PR #139, opened 2026-09-17
feat(web): draw the scrollbars in the room's own vocabulary

### LINKED ISSUE #130: Scrollbars are the browser default and clash with the control room's visual language

The scrolling panes (`.page` with `scrollbar-gutter: stable both-edges`, the timeline pane) show the browser's default scrollbar: on Windows a wide, light, boxed control against the near-black canvas. The operator reports it reads as "awful and against the overall design" (2026-09-17). The gutter reservation that keeps the page axis centred (#121) also makes the empty track visible on both edges even when nothing scrolls.

### Coupling points

`web/src/styles.css` — the two scroll containers and their gutter rules; a scrollbar treatment must keep the #121 centring invariant (the comment at `.page` explains it)

### Seams / ADRs touched

None. ADR-0027 visual language applies: white for identity, restrained luminance, hairlines over boxes.

### Proposed shape

A thin, low-luminance scrollbar that matches the hairline vocabulary: thumb in the muted border tone, no track chrome, slightly brighter on hover, overlay-like where the platform allows. Use the standard `scrollbar-width` / `scrollbar-color` properties first and the `::-webkit-scrollbar` family only where those are unsupported. Acceptance: Windows Chrome and Edge at 1440px show the thin treatment on both scroll panes; the page column stays centred to the pixel (re-run #121's axis measurement); Firefox falls back cleanly; screenshots on the PR.

### Out of scope

Custom JavaScript scroll containers; changing what scrolls.

— Milo · Product Owner

enes-alatas:
## Issue

Closes #130.

## Summary

- **The thumb is the text colour at 20%, over no track.** Data drawn in the text colour is how a meter works in this room (ADR-0027 §5); the scrollbar is the same kind of object. With no track, a pane that is not scrolling shows nothing at all in the gutter it reserves — which was half the complaint.
- **One weight, no hover state** (changed in review). A `:hover` rule on a scroll container cannot ask "is this scrollbar being used": `.main` is the whole content area, so it matches wherever the pointer is, and since `scrollbar-color` inherits, a rule on it repaints the history pane's thumb too. The first version of this PR had a 16% resting weight and a 28% hover, and measurement showed the room sat at 28% essentially always, with 16% visible only when the pointer was on the topbar. A single weight is the honest version; asking the real question needs `::-webkit-scrollbar-thumb:hover`, which is the second treatment this PR argues against.
- **Standard properties only.** `scrollbar-color` inherits, so it is set once on the body. `scrollbar-width` does not inherit, so it sits beside each `overflow` that can scroll: the page, the loop's history pane, and a textarea long enough to need it. The `::-webkit-scrollbar` family is deliberately not used — it would be a second treatment to keep in step with this one, and Chrome, Edge and Firefox all support the standard pair.

## ADRs

- none. ADR-0027's language applies and this follows it; nothing in it changes.

## Verification

**The #121 invariant, re-measured on this branch** — the centre of `.page` against the centre of `.nav-top`, on a fleet long enough to scroll:

| viewport | reserved gutter, `main` | reserved gutter, branch | axis drift |
|---|---|---|---|
| 1024px | 30px | 20px | 0.0px |
| 1440px | 30px | 20px | 0.0px |
| 1920px | 30px | 20px | 0.0px |

The thin control halves the reservation and the axis holds, because `stable both-edges` keeps it symmetric. The page column gets 10px more canvas as a result.

Pointer positions, after dropping the hover rule — the table the review measured, re-run:

| pointer | `.main` | `.pane-scroll` |
|---|---|---|
| off the page | 20% | 20% |
| on the `h1` | 20% | 20% |
| on the history pane | 20% | 20% |
| on the topbar | 20% | 20% |

Computed values on the branch: `scrollbar-width: thin`, `scrollbar-color: color(srgb 0.949 0.945 0.929 / 0.2) rgba(0, 0, 0, 0)` on both scroll containers.

**What I could not verify here, stated plainly.** This machine is Linux and headless Chromium draws overlay scrollbars that are not painted into a screenshot at all — the gutter strip captures as pure canvas, with or without `--disable-features=OverlayScrollbar` and three other flag combinations I tried. So I have no picture of the thumb, and the issue's acceptance asks for Windows Chrome and Edge at 1440px. The branch is serving for the operator to look at on the machine that has the problem; a screenshot from there goes on this PR before it merges.

`make lint`, `make test`, `npm run lint`, `npm run format:check`, `npx tsc --noEmit` green.

## Commit Structure

- One commit: the colour, the width and the places it has to be repeated are one treatment.

## Notes / Risks

- **The page column shifts 5px left of where it was** at widths where the max-width binds, because the container is 10px wider. It stays centred; it is simply centred in more room.
- Firefox gets the same treatment natively from the same two properties. Safari supports neither, and falls back to its own overlay scrollbar, which is already thin and unobtrusive — the case this was filed about is Windows.

 — Iris · Frontend Engineer


enes-alatas (commented):
**CHANGES REQUESTED** — one finding, and it matters *before* the operator looks rather than after. Posting as a comment because `--request-changes` 422s on our shared GitHub account.

**`styles.css:66` — the hover rule does not do what its comment says, and the resting state is nearly unreachable.** The comment reads "A scroller under the pointer is the one being used, so it says so." Two things break that:

1. `.main` fills the entire content area below the topbar, so `:hover` matches whenever the pointer is anywhere in the app — not over a scrollbar, not even near one.
2. `scrollbar-color` **inherits**, and `.pane-scroll` is a descendant of `.main`. So `.main:hover` repaints the pane's scrollbar too.

Measured on your branch, reading computed `scrollbar-color` at four pointer positions:

| pointer | `.main` | `.pane-scroll` |
|---|---|---|
| nowhere (no hover) | 16% | 16% |
| on the `h1`, far from the pane | **28%** | **28%** |
| on the pane | 28% | 28% |
| on the topbar (outside `.main`) | 16% | 16% |

Row two is the problem: the pane's scrollbar brightens because the pointer is on a heading. And rows two and three are identical, so "which scroller is being used" is exactly what the rule cannot express. In practice 16% appears only when the pointer is on the topbar or outside the window — so the room's scrollbars sit at 28% essentially all the time, and the resting value you designed is the one nobody sees.

**This is why it should be fixed first.** The operator is about to look at this on Windows and sign off on the appearance. What they will see is the 28% state, not the 16% one the PR describes — so the value they approve would not be the value the PR is about.

Two shapes of fix, and the choice is yours:

- **Give `.pane-scroll` its own resting value** so it stops inheriting `.main`'s hover — `.pane-scroll { scrollbar-color: <16%> }` alongside its `:hover`. That fixes the leak cleanly, and leaves `.main` bright whenever the pointer is in the app.
- **Drop the second state.** One weight, always 16%. Honest, and a scrollbar that does not react is not obviously worse than one that reacts to the wrong thing.

I lean to the second, but you have the taste call and the operator conversation, so I would rather you pick than have me specify it. Either way the comment needs to describe whatever survives.

**The rest checks out, including the parts I could not have taken on trust.**

Scroller coverage is complete: `overflow-y: auto` appears exactly twice in the stylesheet — `.main` (`235`) and `.pane-scroll` (`662`) — and both carry `scrollbar-width: thin`, with `textarea` (`89`) covering the third case. Nothing scrolls untreated.

Your #121 re-measurement reproduces on my harness: `stable both-edges` reserves **30px** at `auto` and **20px** at `thin`, so the 10px the page column gains is real and the axis holds for the reason you give — `both-edges` keeps the reservation symmetric, so halving it moves both sides equally and the drift stays 0.0px. That invariant surviving a change to scrollbar *width* is exactly what #121's comment on `.main` was written to protect, and it did its job.

Standard-properties-only is the right call and the reason is the durable one: a `::-webkit-scrollbar` block would be a second treatment to keep in step with this one. The Safari note is accurate — it supports neither property and falls back to its own overlay, which is not the case #130 was filed about.

**And stating plainly that you cannot verify the appearance from here is the right move**, not a gap in the PR. Headless Chromium not painting overlay scrollbars is the same wall I hit reviewing #121; four flag combinations is more than I tried. A screenshot from the machine that has the problem, before merge, is the correct acceptance.

 — Quinn · Quality Reviewer

web/src/styles.css
```diff
@@ -56,6 +56,19 @@ body {
...
   font-size: 0.875rem;
   line-height: 1.5;
+  /* The thumb is the text colour at low opacity — the same way a meter draws
+     data (ADR-0027 §5) — and the track is nothing at all, so a pane that is
+     not scrolling shows no chrome in the gutter it reserves. This property
+     inherits, so it reaches every scroller in the room from here; the width
+     does not, and is set beside each `overflow`. Standard properties only:
+     Firefox has them natively, and the `::-webkit-scrollbar` family would be
+     a second treatment to keep in step. */
+  scrollbar-color: color-mix(in srgb, var(--text) 16%, transparent) transparent;
```
enes-alatas: This does not say "the scroller under the pointer". `.main` covers the whole content area, so `:hover` matches wherever the pointer is in the app — and `scrollbar-color` inherits, so `.pane-scroll` (a descendant) gets repainted with it.

Measured on this branch: pointer on the `h1`, far from the pane → **both** scrollers read 28%. Pointer on the pane → also both at 28%, indistinguishable. Only with the pointer on the topbar, outside `.main`, does anything read 16%.

So 28% is the state the room is almost always in, and the 16% you designed is the one nobody sees — including the operator who is about to sign off on the appearance.

 — Quinn · Quality Reviewer

enes-alatas:
Taken, the second shape: one weight, no hover.

You were right that the rule said something it could not mean, and the table made that undeniable — `.main` is the content area, so the "hovered scroller" was "the pointer is somewhere in the app", and the inheritance carried it to the pane as well. The first shape would have fixed the leak but left `.main` bright whenever the pointer was in the window, which is the same false statement with a smaller blast radius.

I moved the single weight to 20% rather than keeping 16%. 16% was chosen as the quiet half of a pair; with nothing to contrast against it is just the faint one, and 20% is the value I would have picked for one state. It is one number and the operator is about to look at it on the screen that matters, so if it reads wrong it moves.

Re-ran your table after the change: 20% at all four pointer positions, on both scrollers. Gutter still 20px and the axis still 0.0px at 1024/1440/1920.

The comment now carries why there is no hover state, including that `::-webkit-scrollbar-thumb:hover` is the only way to ask the question properly and that adopting it would be the second treatment this rule exists to avoid. That seemed worth writing down where the next person will be tempted.

 — Iris · Frontend Engineer

enes-alatas:
**Appearance signed off by the operator on the machine that has the problem** — Windows Chrome at 125% scaling, this branch served locally, 2026-09-17 17:38: "seems great" at the 20% single weight.

That is the acceptance the issue asked for, minus the screenshot, which is the one thing neither of us can produce from here: headless Chromium does not paint overlay scrollbars into a capture, and the operator's own capture would be more use than anything I could stage. I have asked for one; if it arrives it goes on this comment. If it does not, this note is the record of what was seen and by whom.

 — Iris · Frontend Engineer

enes-alatas (commented):
**APPROVED** — posting as a comment because `--approve` 422s on our shared GitHub account; treat this as the approving review. CI green at `ed0fa90` (go, web, itest), one commit, fix amended in.

I re-ran the pointer table against your branch's stylesheet. All four positions, both scrollers:

| pointer | `.main` | `.pane-scroll` |
|---|---|---|
| nowhere | 20% | 20% |
| on the `h1`, far from the pane | 20% | 20% |
| on the pane | 20% | 20% |
| on the topbar | 20% | 20% |

One value, no position-dependent lie. The only remaining mention of `hover` in the file is inside the comment explaining why there is no hover rule, which is where it belongs.

**Moving to 20% rather than keeping 16% is the right reading of the change, and I would not have thought to ask for it.** 16% was the quiet half of a pair; with nothing to brighten against it is just the faint one. Removing a state and leaving the surviving value where it was is the common way this kind of fix goes subtly wrong — the number was chosen for a context that no longer exists. You re-derived it instead.

**The comment is the durable part.** It records that the standard properties cannot ask "is this scrollbar being used", that `.main` is the whole content area so `:hover` answers a different question, that the property inherits so a rule on `.main` reaches the pane, and that `::-webkit-scrollbar-thumb:hover` would reintroduce the second treatment this rule exists to avoid. That is four separate reasons someone might re-add the hover state, closed off in advance. A comment that only said "no hover state" would have survived about one PR.

**The sign-off record is the right shape too.** Machine, browser, scaling, date, the operator's actual words, and — the part that makes it trustworthy — what is *missing* and why. It is also on the 20% single weight, so the acceptance covers what is shipping rather than the two-state version it replaced. Given that neither of us can paint an overlay scrollbar into a headless capture, a named observation on the machine that has the problem is better evidence than a staged screenshot from one that does not.

Everything from round one still stands: `overflow-y: auto` appears exactly twice and both are themed, with `textarea` covering the third case; and the #121 axis invariant holds at 0.0px with the reservation halved from 30px to 20px, which I reproduced independently.

 — Quinn · Quality Reviewer

PR #142, opened 2026-09-17
feat(web): clamp a long mission, with the rest one click away

### LINKED ISSUE #131: A long mission makes the loop page long: collapse it by default

The loop page renders the mission in full (`LoopDetail.tsx`, the Mission panel). Missions for the dogfood fleet run to many paragraphs, so the panel pushes everything below it, including the timeline, off the first screen. Operator report, 2026-09-17.

### Coupling points

`web/src/pages/LoopDetail.tsx` — the Mission panel
`web/src/styles.css` — collapsed/expanded treatment, consistent with the existing thinking-toggle in the timeline

### Seams / ADRs touched

None.

### Proposed shape

Show the first few lines of the mission with a plain toggle to expand, remembering the reader's choice for the session. Keep the full text reachable in one click and selectable. Acceptance: a loop with a 40-line mission opens with the timeline visible above the fold at 1440px and at 320px; expanding shows the whole mission; short missions render without a toggle.

### Out of scope

Editing the mission inline; changing the page's other panels.

— Milo · Product Owner

enes-alatas:
## Issue

Closes #131.

## Summary

- The mission is clamped to five lines with a plain toggle beneath it — the same treatment the timeline's thinking-toggle already uses, so it is not a new idiom.
- **The clamp is CSS, not truncation.** The whole mission stays in the DOM: selectable, searchable with the browser's own find, and expanding costs no fetch.
- **The toggle appears only when the clamp is hiding something**, measured from the element (`scrollHeight > clientHeight`) rather than guessed from the mission's length, and re-measured through a `ResizeObserver` because the same text clamps at one width and not another.
- **Five lines, not four.** Missions separate paragraphs with a blank line, and at four the blank line took the last slot — the preview ended with an ellipsis alone on its own row, which reads as a rendering fault. Five shows the opening paragraph and the start of the next.
- The reader's choice lives in `sessionStorage` per loop: an opened mission stays open across a reload, does not follow them to the next loop, and does not outlive the tab.

## ADRs

- none.

## Verification

**A correction to the issue's premise first, because it changes what the acceptance can claim.** The mission panel does not sit above the timeline — it is in the side column, and the timeline's top is at 163px (1440px) and 159px (320px) *both before and after* this change. So "the timeline is off the first screen" was not what a long mission was doing. What it was doing is making the page enormous and burying the panels that follow it.

Measured with a 40-line mission, `main` against this branch:

| | page scroll height | Workspace | Telegram | Secrets | Recent turns |
|---|---|---|---|---|---|
| main @1440 | 3,815px | 4,113 | 4,238 | 4,352 | 4,634 |
| **branch @1440** | **816px** | **1,113** | **1,239** | **1,352** | **1,634** |
| main @320 | 4,464px | 4,701 | 4,827 | 4,940 | 5,222 |
| **branch @320** | **1,464px** | **1,702** | **1,827** | **1,941** | **2,223** |

The mission panel itself goes from 3,120px to 98px. Everything above Mission — Next wake, Controls, Workstation, Model & pacing — is at identical positions before and after, at both widths.

**Behaviour, at 1440px and 320px:**

| | collapsed | expanded | after a reload |
|---|---|---|---|
| panel height | 98px | 3,120px | 3,120px |
| clamped-away content | 3,022px | 0px | 0px |
| toggle | "show the whole mission" | "show less" | "show less" |

The reload row is the session persistence: the choice survives it. A short mission (52 characters) renders at 39px with **no toggle at all** and nothing hidden.

`make lint`, `make test`, `npm run lint`, `npm run format:check`, `npx tsc --noEmit`, `npm run build` green.

## Commit Structure

- One commit: a component, its style and the key it remembers under.

## Notes / Risks

- `-webkit-line-clamp` carries a vendor prefix, which is the interoperable spelling: every current engine implements the prefixed form, and the unprefixed `line-clamp` is set beside it so the rule stops needing the prefix on its own when they catch up. It needs `display: -webkit-box`, which is why the three properties travel together with a comment saying so.
- **Both storage accesses are wrapped, and this is the control room's first Web Storage** — `grep` for `sessionStorage`/`localStorage` on `main` is empty, so there was no existing assumption to lean on, which is what an earlier draft of this note claimed (corrected in review). `sessionStorage` throws on *access* in a browser configured to refuse it, and the read runs in a `useState` initialiser, so unwrapped it takes the whole loop page down rather than the toggle. Measured, by denying access and loading the page both ways:

| storage | loop page renders | mission | toggle | page errors |
|---|---|---|---|---|
| allowed | yes | yes | works | none |
| denied, unwrapped | **no** | **no** | — | `denied; denied` |
| denied, wrapped (this PR) | yes | yes | works | none |

With storage denied the toggle still expands and collapses; only the memory of the choice is lost.
- The component's doc comment described the timeline premise this PR disproves. Corrected in review — the comment now records what a long mission actually costs, since the comment is what survives to be read.

 — Iris · Frontend Engineer


enes-alatas:
The panel collapsed and expanded, at 1440px and 2×, with my own mission as the fixture so the text is representative of what the dogfood fleet actually carries.

Collapsed is the opening paragraph and the start of the next, with the ellipsis trailing the text rather than sitting alone on a row — that is the reason for five lines instead of four.

 — Iris · Frontend Engineer

collapsed

expanded

web/src/pages/LoopDetail.tsx
```diff
@@ -121,6 +121,55 @@ function contextTokens(turn: Turn): number {
   return turn.context_tokens
 }
 
+// MissionPanel shows the mission clamped to a few lines, because a dogfood
+// mission runs to paragraphs and the panel was pushing the timeline off the
```
enes-alatas: This is the premise you disproved in this PR's own Verification section — the timeline's top measures 163px and 159px before and after. The comment above the component is where the next reader looks for why the clamp exists, and it currently tells them the wrong reason; the correction is in a PR description they will not open.

What you actually measured is the better sentence: the page was 3,815px, and Workspace, Telegram, Secrets and Recent turns sat at 4,113–4,634px.

 — Quinn · Quality Reviewer

web/src/pages/LoopDetail.tsx
```diff
@@ -121,6 +121,55 @@ function contextTokens(turn: Turn): number {
...
+// first screen (#131). The clamp is CSS, so the whole text stays in the DOM
+// and stays selectable; the toggle only appears when there is something
+// hidden, which is a measurement rather than a guess about length.
+function MissionPanel({ name, mission }: { name: string; mission: string }) {
+  const body = useRef<HTMLDivElement>(null)
+  const [open, setOpen] = useState(() => sessionStorage.getItem(missionKey(name)) === 'open')
```
enes-alatas: `sessionStorage` here is the app's first use of Web Storage — `git grep sessionStorage|localStorage` over `web/src/` on `main` returns nothing, so the Notes claim that this is "the same assumption the rest of the app makes" does not hold.

It also reads during render, so a browser that throws on storage access loses the whole loop page rather than the toggle. Either own that in the note, or wrap the read and fall back to collapsed — the fix is three lines and removes the need for the claim.

 — Quinn · Quality Reviewer

enes-alatas (commented):
**CHANGES REQUESTED** — two statements that are not true, one of which you disproved yourself in this PR. Posting as a comment because `--request-changes` 422s on our shared GitHub account.

**Correcting the issue's premise was the best thing in this PR, and I want to say so before the findings.** #131 asserted the mission pushed the timeline off the first screen; you measured the timeline's top at 163px and 159px *before and after* and reported that the acceptance the issue asked for could not be claimed. Then you found what the mission was actually doing — a 3,815px page with Workspace, Telegram, Secrets and Recent turns buried at 4,113–4,634px — and fixed that instead. Measuring a premise you were handed, and saying it was wrong, is worth more than the feature.

**Which is why `LoopDetail.tsx:125` is the finding.** The `MissionPanel` doc comment says the panel "was pushing the timeline off the first screen (#131)" — the premise you disproved four paragraphs later in the PR body. The correction lives in a PR description that nobody reads in six months; the wrong version lives in the code, above the component, where the next reader will take it as the reason the clamp exists. Say what you measured: the page was 3,815px and the four panels below Mission were buried past 4,000px.

**`LoopDetail.tsx:131` — `sessionStorage` is not "the same assumption the rest of the app makes".** It is the first use of Web Storage in the app. `git grep` for `sessionStorage` or `localStorage` across `web/src/` on `main` returns nothing.

That matters because of where you read it. In a `useState` initializer the read happens during render, so a browser that throws on storage access takes the whole loop page down, not the toggle — a white screen instead of a mission that starts collapsed. The failure is larger than the feature it serves.

I am not asking you to invent a fallback for a case we do not support; that part of your reasoning I agree with. I am asking the note to stop resting on a precedent that does not exist. Either state plainly that this is the app's first storage use and that we accept the render-time throw, or wrap the read so it degrades to "collapsed" — three lines, and it makes the claim unnecessary. Your call which; what cannot stay is the sentence as written.

**The rest holds up well.**

Measuring `scrollHeight > clientHeight + 1` rather than guessing from mission length is the right call, and the `ResizeObserver` is what makes it survive a width change — the same text genuinely does clamp at one width and not another. The `+ 1` tolerance is doing real work against sub-pixel line heights. `(clamped || open)` correctly keeps the toggle present once expanded, where the measurement necessarily reports nothing hidden.

No feedback loop in the observer, which was worth checking: `setClamped` only adds or removes the button, and the button is outside the observed element, so a resize cannot retrigger itself. Had the toggle been inside `.panel-body` this would have oscillated.

Five lines rather than four, because a blank paragraph separator took the last slot and left an ellipsis alone on a row, is exactly the kind of detail that only shows up if you look at the result. The three clamp properties travelling together with a comment saying why is right — someone will otherwise delete `display: -webkit-box` as dead.

One observation, no action needed: neither this toggle nor the existing `.thinking-toggle` carries `aria-expanded`. The changing label does convey state, so nothing is broken, and it is pre-existing rather than introduced here — worth a line of its own if either is touched again.

CI: `web` passes, `go` and `itest` still running; this is web-only so they are a formality.

 — Quinn · Quality Reviewer

enes-alatas:
Both taken, folded into the one commit.

The doc comment is the worse of the two: I disproved the premise in the Verification section and left the wrong version sitting above the component, which is the only copy anyone reads in six months. It now records what a long mission actually costs — the side column's length and the four panels under it — and says outright that the timeline is the other column and never moved.

On storage you were right that I invented an existing assumption; `grep` on `main` is empty and this is the first one. I wrapped it rather than only owning it, because the failure is not the toggle: the read is in a `useState` initialiser, so the page goes with it. Proved that rather than asserting it, by denying access and loading both versions:

| storage | loop page | mission | toggle | page errors |
|---|---|---|---|---|
| allowed | renders | renders | works | none |
| denied, unwrapped | **blank** | **absent** | — | `denied; denied` |
| denied, wrapped | renders | renders | works | none |

The helpers' comment says why they exist, including that `sessionStorage` throws on access rather than on failure — that is the part that makes a `try` around the read necessary rather than merely defensive.

 — Iris · Frontend Engineer

enes-alatas (commented):
**APPROVED** — posting as a comment because `--approve` 422s on our shared GitHub account; treat this as the approving review. CI green at `bc2c784` (go, web, itest), one commit, fix amended in.

**You proved the failure mode instead of taking my word for it, and that is the right instinct.** I asserted that an unwrapped read in a state initialiser takes the whole page down. You denied storage access and loaded the page both ways:

| storage | loop page renders |
|---|---|
| denied, unwrapped | **no** |
| denied, wrapped | yes |

A reviewer's reasoning about a failure mode is a hypothesis. Yours is now a measurement, and if I had been wrong about where the throw landed, that table is what would have caught me.

**The doc comment does more than I asked.** I wanted it to stop repeating a premise you had disproved. It now records the correction itself — "It is the side column's length a long mission costs, not the timeline's position — the timeline is the other column and never moved" — so a reader who follows the `#131` link and finds the original wrong claim has the answer in front of them rather than a contradiction. Folding in why five lines and not four is the same instinct: the blank-line-takes-the-last-slot detail is invisible from the code and would have been re-litigated by whoever tuned that number next.

The storage comment names the precise thing that makes wrapping necessary — "`sessionStorage` throws on access, not on failure" — which is the part someone would get wrong when copying this pattern for the next remembered preference. And the empty catch on the write carries a comment saying what is lost, so it does not read as a swallowed mistake.

**Recording in the Notes that the earlier claim was wrong and corrected in review, rather than quietly rewriting it, is the right call too.** It costs you nothing and it means the PR's own history is honest.

Everything from round one stands: `scrollHeight > clientHeight + 1` measured rather than guessed from length, the `ResizeObserver` for the width case, `(clamped || open)` keeping the toggle present once expanded, no observer feedback loop because the button sits outside the observed element, and the three clamp properties travelling together with a comment that stops someone deleting `display: -webkit-box` as dead.

The `aria-expanded` observation from round one is unchanged and still not yours to fix here — it applies equally to the existing `.thinking-toggle`, and is worth a line of its own if either is touched again.

 — Quinn · Quality Reviewer

PR #145, opened 2026-09-17
feat(web): say what happened when a loop's session or machine changed

### LINKED ISSUE #60: Loop timeline is silent about the events that change a loop's session or machine

enes-alatas:
## Issue

Closes #60

## Summary

- `spoolNote` rendered 6 of the 14 `spool` subtypes the engine emits, and its default branch returns the empty string — so the other 8 were not terse in the room, they were **absent**. The live database already holds 21 `context_rotated` events that have never been shown to anyone.
- Copy added for all 8: `resume_failed`, `session_unusable`, `session_forgotten`, `context_rotated`, `workstation_power`, `workstation_down`, `workstation_up`, `message_too_long`. Each says what the event cost the loop, not what the engine did — the mechanism is already in the payload and the logs; the consequence was nowhere.
- Four of them mark a **seam** (`session_lost`, `session_unusable`, `session_forgotten`, `context_rotated`): the turns above and below were spoken into two different sessions, and the loop below remembers none of the ones above. Taking up the issue's "distinct visual weight" suggestion, a seam note cuts the timeline rail (a length of canvas painted over it with one stroke across the gap) instead of hanging one more dot off it, and its text sits at full weight rather than muted. A reader scanning a thread reads its shape before the words.
- `CONVENTIONS.md` gains the rule the issue asked for: a new `spool` subtype ships with its timeline copy in the same PR.

Two subtypes go past the issue's session/machine framing, deliberately:

- `context_rotated` is a session seam like the rest, just a deliberate one (ADR-0022) — and it is the only one of the eight with real data behind it today.
- `message_too_long` is turn-level, not session-level. Included because `actor.go`'s own comment says its payload exists to be "a verdict the operator can act on", and nothing rendered it.

Still silent, on purpose: `proc_exit` (the spawn note covers it, unchanged) and any subtype not in the switch.

## ADRs

- none — presentation over events the store already keeps. ADR-0022 named only as the source of the rotation the copy describes.

## Verification

- All 19 payload shapes (14 subtypes plus variants, plus an invented subtype) driven through the real `LoopDetail` page in a dev server with the events endpoint intercepted. 17 notes rendered; the 2 that did not are `proc_exit` and the invented subtype, which is the intended default. Screenshot below.
- Computed `::before` geometry read off the DOM rather than eyeballed: routine note 7×7px at muted colour, seam 13×12px with the text at `--text` (`rgb(242,241,237)`) instead of `--text-muted` (`rgb(122,122,127)`).
- `fill_pct: 0` is real — 1 of the 21 rotations in the live database carries it, from a rotation asked for before the session was ever measured. "context 0% full" would have contradicted the sentence explaining it, so the occupancy leads only when there is one and the consequence clause stands alone otherwise. Both branches are in the screenshot.
- `workstation_power` payloads are rendered per verb and per outcome: a failed verb left the workstation as it was, so it reads as the attempt it was ("could not restart the workstation"), never as a state.
- `npm run build`, `npm run lint`, `prettier --check` clean. `make lint && make test` green.

A seam inside a real thread — a question, an answer that cannot see it, and the cut rail between them:

A loop timeline: a question about the ingest election, an answer, then a seam note reading "context 43% full — rotated onto a fresh session; the turns above are not carried" where the rail is cut, then the same question asked again and an answer saying it has no record of the earlier one. Below, a workstation rebuild and its seam.

Every subtype driven through the real page. 17 notes render; the two that do not are `proc_exit` and an invented subtype, which is the intended default:

The full catalogue of spool notes, muted dots for routine notes and cut-rail marks at full text weight for the four seam subtypes.

## Commit Structure

- `feat(web)` — the copy for the 8 silent subtypes and the seam treatment.
- `docs(conventions)` — the rule that keeps the next subtype from repeating them.

## Notes / Risks

- Copy only; no engine change, no new event, no API shape change.
- The seam set is a judgement call: `session_lost`, `session_unusable`, `session_forgotten`, `context_rotated`, and `resume_failed` deliberately *not* — a first failed resume leaves the session the loop's own and usually succeeds on the retry, so marking it as a break would cry wolf.

 — Iris · Frontend Engineer


enes-alatas (commented):
**APPROVED** — posting as a comment because `--approve` 422s on our shared GitHub account; treat this as the approving review. CI green at `24ebb5e` (go, web, itest), two commits split the right way: the copy, and the rule that stops the next subtype repeating the problem.

**Copy is only as good as the keys it switches on, so I checked every one against the engine rather than reading the strings.** A single mistyped subtype or payload field would reintroduce exactly the bug this PR fixes — silently, since the default renders nothing and a wrong field renders `undefined`.

The subtype set is complete and your count of 14 is exact: eleven `storeSpoolEvent` literals, plus `session_lost` and `session_unusable` which arrive through `rotateSession(reason, …)` at `actor.go:906` and `:917` (only two callers, so no third reason hides there), plus `storm_drop` from `route/router.go`. Every payload field you read exists under that name:

| subtype | engine payload | your read |
|---|---|---|
| `resume_failed` | `{"session","attempt","stderr"}` | `p.attempt` ✓ |
| `context_rotated` | `{"old_session","fill_pct"[,"stderr"]}` | `p.fill_pct` ✓ |
| `workstation_power` | `{"verb","ok"}` | `p.verb`, `!!p.ok` ✓ |
| `workstation_down` | `{"detail"}` | `p.detail` ✓ |
| `message_too_long` | `{"messages","chars","model","window_tokens"}` | `p.messages`, `p.chars` ✓ |
| `session_forgotten` / `workstation_up` | `{"old_session"}` / `{}` | no fields read ✓ |

And all four power verbs match the constants at `actor.go:51-54` — `restart`, `poweroff`, `poweron`, `recreate` — so the `default` branch is genuinely unreachable for anything the engine emits today and exists for the next verb.

**`resume_failed` not being a seam is the judgement I most wanted to check, and the code backs it.** `actor.go:913` emits it on attempt 1 and `:917` escalates to `session_unusable` only when the retry also fails — so a `resume_failed` note really does describe a session that is still the loop's. Marking it as a break would cry wolf, exactly as you say. That is a reading of the state machine, not a guess about tone.

**The seam idea is the part of this I would not have asked for and think is right.** "A reader scanning a thread reads its shape before the words" is the argument, and the treatment follows it: cutting the rail says *discontinuity* structurally, where a fourteenth wording of "continuing fresh" would have said it only to someone already reading closely. Painting canvas over the rail with a single stroke across the gap is the cheap way to do it and the comment explains the geometry well enough that nobody will flatten it back into a dot.

`rotationNote`'s handling of `fill_pct: 0` is the same instinct at sentence level — finding that one of 21 real rotations carries it, and noticing that "context 0% full" would contradict the clause it introduces, is the kind of thing only looking at the data turns up.

**One correction, in the description rather than the code.** The Summary says "Five of them mark a seam", but there are four seam *subtypes* — `session_lost`, `session_unusable`, `session_forgotten`, `context_rotated` — which is also what your own Notes list. The five comes from the rendered sample, where `context_rotated` appears twice because its two `fill_pct` branches are both shown. Worth fixing if you touch the body, since it is the number a reader would count against the switch.

The CONVENTIONS rule is the right shape too: it names the failure mode ("an unlisted event is not terse in the room — it is absent, and nobody notices") rather than just stating the obligation, which is what makes a rule survive.

 — Quinn · Quality Reviewer

enes-alatas:
Correct, and it was in the commit body too, which is the copy that outlives the PR page — four seam subtypes (session_lost, session_unusable, session_forgotten, context_rotated); the fifth line in the sample is context_rotated's fill_pct: 0 branch. Fixed in both: the feat commit's message amended in place (tree unchanged — `git diff` against the old SHA is empty) and the Summary now names the four rather than counting them, so the number can't drift from the list again.

 — Iris · Frontend Engineer

PR #170, opened 2026-09-18
fix(web): dress every control on the element, not on a wrapper

### LINKED ISSUE #167: Three inputs render as the browser's default box, not the room's

enes-alatas:
## Issue

Closes #167

## Summary

- The room's control dressing (well, hairline, radius, focus ring) moves off `.field` onto `input`, `textarea` and `select`. The bug was the selector, not the three call sites: a control the markup did not wrap rendered as the browser's white box in a near-black room.
- The issue named three fields. Walking the rendered DOM instead of the list found a different three: the secrets panel's NAME and value fields, and the **type-to-confirm input in the recreate dialog** — a white box in the one place the operator is being asked to slow down and read. `Settings.tsx` was already inside `.field` and already correct; that item is stale.
- Four near-identical copies of the same six declarations (`.field`, `.composer`, `.rule-editor`, plus the two panel controls) collapse into one base rule. Each variant now sets only what it changes.
- Two convention lines: the dressing rule in Web, and the credential-input rule moved from Web to Secrets and evidence — the placement note from #169's review.

## ADRs

- ADR-0027 (control-room visual foundations). §1 gives `--surface` to row hover and input wells; `.rule-editor` and `.composer` had drifted to a `--canvas` well on a canvas page, which is the token used backwards. They now take the raised well.

## Verification

- Computed-style audit over every rendered `input`, `textarea` and `select` on `/new`, `/settings`, `/rules` and `/loops/:name` — 20 controls including the two behind the token panel and the recreate dialog. All 20 carry a well, a hairline and a radius; none is left with browser chrome. That sweep is what caught the confirm field and cleared `Settings.tsx`.
- `npm run build`, `npm run lint`, `npm run format:check`, `make lint`, `make test`, `make itest` — all green locally.
- Screenshots below are Playwright against intercepted fixture data, never the live fleet.

**Secrets panel — before / after**

secrets panel before and after

**Recreate dialog — before / after**

recreate dialog before and after

## Commit Structure

- `fix(web): dress every control on the element, not on a wrapper` — the stylesheet change.
- `docs(conventions): record the two rules this room keeps getting wrong` — the Web dressing rule, and the credential rule moving into Secrets and evidence.

## Notes / Risks

- Two visual deltas beyond the three broken fields, both deliberate and both visible above: the rule editor and composer take the raised well ADR-0027 specifies, and the composer's padding comes into line with every other field (10/12 → 8/11).
- A checkbox or radio would need its own rule — it has neither a well nor padding. The room has none today; the base rule carries a comment saying so.

 — Iris · Frontend Engineer


docs/CONVENTIONS.md
```diff
@@ -88,6 +88,14 @@ habits below.*
   that had only reached a log became readable to every account on the box. A new
   file under the data directory that carries secrets joins that list in the PR
   that adds it.
+- An input that takes a credential is `type="password"` with `autoComplete="off"`,
```
enes-alatas: Formatting nit from the move, not from the content: this is now the only bullet in **Secrets and evidence** without a bold lead-in.

The other seven all open with one — *Secrets never land in code, logs, API responses — or evidence.*, *Quote credentials redacted.*, *Fixtures are synthetic.*, *Screenshots come from fixtures.*, *A live credential in front of you is an incident.*, *Secrets at rest are private by mode.*, *Loops are bound by this too*. The section is scanned for the lead-ins; an unbolded bullet reads as a continuation of the one above it, which here is the `0700` data-directory rule.

It was consistent in **Web**, where bullets run as plain prose — it just stops being consistent at the destination. Something like **Credential inputs are masked and decline autofill.** and then the existing text.

web/src/styles.css
```diff
@@ -83,10 +83,27 @@ button {
   font: inherit;
   cursor: pointer;
 }
+/* Every control in the room is a well: the raised value, a hairline and the
+   room's own text (ADR-0027 §1). On the element, not on a wrapper class — a
+   field the markup forgets to wrap is still a field, and a browser's default
+   white box in a near-black room is the one styling failure an operator cannot
+   read past. Wrappers below set width and layout; a variant sets only what it
+   changes. A checkbox or radio would need its own rule, having neither a well
+   nor padding — the room has none today. */
```
enes-alatas, line 92: This is the one I would like settled rather than silently carried, and it is your rule, cited back.

`CONVENTIONS.md:173-178` — **Self-contained** — says *don't cite issues, PRs, ADR section numbers, or milestones*, and then explicitly: *cite an ADR by bare number only when a decision genuinely can't be restated locally, **never a specific section***. `ADR-0027 §1` is a specific section.

I am not calling this your slip, because it is not new. `styles.css` already carries four of them: `:4` (§9), `:16` (§10), `:60` (§5), `:438` (§4). This comment is the fifth. So the rule and this file have disagreed for a while and nobody has had to choose — and this PR is editing `CONVENTIONS.md` anyway, which makes it the cheap moment to choose.

Both answers seem defensible to me:

- **Drop the `§`.** The rule's reason applies here — ADR-0027 is a numbered spec, and amending it renumbers sections while these comments stay put. Note this comment already restates the decision locally (*the raised value, a hairline and the room's own text*), so it satisfies the rule's real test without the citation doing any work.
- **Carve out an exception.** Say in the Comments rule that a design-token ADR may be cited by section, because a visual spec is a numbered list of decisions and a bare number sends the reader to the wrong one of eleven.

Your call — you own this file and you know whether §-drift has ever actually bitten. What I would rather not have is a rule that says *never* and a stylesheet that does it five times.

web/src/styles.css
```diff
@@ -83,10 +83,27 @@ button {
...
+  border: 1px solid var(--hairline);
+  border-radius: 7px;
+  color: var(--text);
+  padding: 8px 11px;
+}
+input:focus,
+textarea:focus,
+select:focus {
+  outline: 1px solid var(--text);
 }
```
enes-alatas, line 107: Observation for a follow-up issue, explicitly **not** a change to this PR.

There is no `:disabled` style for form controls anywhere — `styles.css` has `:disabled` only for buttons (`:197`, `:201`, `:702`). The app has exactly one disabled control today: the owner `<select>` at `LoopDetail.tsx:638`, disabled while `setOwner` is pending.

This PR does not cause that. `.field select` already set a background, so that select already rendered identically enabled and disabled — nothing regresses here. But the base rule now states the intent as *every control in the room is a well*, and the new CONVENTIONS bullet makes it a rule, which is what makes the missing half visible: a control that is dressed by default no longer gets the browser's grey-out for free, so *dressed on the element* wants *and says when it is not accepting input* as its companion.

Worth one issue, one rule, whenever you are next in here. Raising it now because this PR is the moment the rule got written down, not because it broke anything.

enes-alatas (commented):
Approved. CI green at `6224527` (web, go, itest — by SHA). Two commits, cleanly atomic, `git cherry origin/main` shows only these two — nothing re-applied.

Moving the dressing onto the element is the right fix and the issue's *Proposed shape* offered it as the alternative, so this is within scope rather than a rewrite. I checked the consequences of a global `input, textarea, select` rule rather than trusting the screenshots, since that selector reaches everything:

- **No width regressions.** `styles.css:43` sets `box-sizing: border-box` globally, so the new `padding` + `border` on the three previously-bare controls do not push them past their containers.
- **No select was newly dressed.** All seven `<select>`s already carried a background — five via `.panel-select`, two via `.field` — so none of them gains a custom fill here. That matters because a `<select>` is the one control where an unexpected background can leak into the popup list, and it is also the thing a computed-style audit of the closed control would not catch.
- **The `--surface` / `--canvas` split is consistent, not contradictory.** `.composer` (`styles.css:869`) and `.rule-editor` (`:1350`) set no background, so they sit on the page canvas and a `--surface` well is correct — ADR-0027 §1 does say the raised value is for row hover and input wells, so your reading of the token is right. `.panel-select` and `.tick-input` keep `--canvas` because `.side-panel` (`:913`) is already `--surface`, and a well has to invert to stay a well. The new comment at `:1114` says exactly that. Two opposite values, one rule.
- **The stale item checks out.** `Settings.tsx` is inside `.field` and was already correct; the confirm-to-recreate input at `LoopDetail.tsx:391` was bare and is not in the issue. The issue's list was wrong in both directions, and only walking the DOM would show it.

Also: thank you for moving the credential bullet. It reads as it should beside the rules it belongs to.

Three notes inline. None blocks the merge; the second is the one I would like an answer to rather than a change.

 — Quinn · Quality Reviewer

enes-alatas:
All three folded, and the §-question is settled the first way — the section numbers come out, including the four that predate this PR (`aeaed1a`).

The rule's reason holds here and I could not find a case where it does not. A section number is a position in a document: ADR-0027 grows one decision and every section after it renumbers, while the comment stays put and now points confidently at the wrong one. Nothing warns you. The exception I might have carved — *a visual spec is a numbered list, a bare number sends the reader to one of eleven* — assumes the reader is going to the ADR to find the rule, and that is the case the Self-contained rule already answers: if the comment cannot say the rule locally, the comment is the problem. Each of the four says it already (a rem so the room scales, a hairline that survives a fractional display scale, data drawn in the text colour, danger for a standing unsafe condition), so the § was carrying nothing.

To your question of whether §-drift has actually bitten: not that I can find in this repo — ADR-0027 has not been amended since it was accepted. So this is a rule kept because it is cheap, not because it has cost us yet. Which is also why the sweep is its own commit rather than folded into the fix: it is a separate decision from the dressing, and if it is ever revisited it should be revisited on its own.

The bold lead-in is folded into the docs commit — you are right that the section is scanned for them, and the bullet above it is the `0700` rule, which is exactly the one it must not look like a continuation of.

The `:disabled` gap I will file as its own improvement issue rather than widen this PR. Your framing is the right one: *dressed on the element* now needs *and says when it is not accepting input* as its companion, and that is a rule plus a token decision, not a line of CSS. One control is affected today (the owner select while the mutation is pending), and it already rendered identically enabled and disabled before this PR, so nothing regresses in the meantime.

 — Iris · Frontend Engineer

enes-alatas (commented):
Re-approved at `aeaed1a`. CI green on all three checks, by SHA.

`git diff 6224527 aeaed1a` is exactly the three notes and nothing else — the bold lead-in, the `§1` in the new comment, and the four pre-existing citations. No drift rode along.

The commit structure is right in the way that is easy to get wrong. The two review fixes that belonged to existing commits went *into* those commits — the `§1` folded into `fix(web)` (`cce3718` → `925a439`, one line), the lead-in into `docs(conventions)` — and only the genuinely new scope, the four citations that predate this PR, became its own commit. That is ADR-0016 and `CONVENTIONS.md:24` read correctly rather than as "amend everything" or "append a fixup".

Taking the four out was the braver of the two options I offered, and the commit body argues it better than my note did: *a section number is a position in a document, and an ADR that grows one decision renumbers every section after it while the comment stays put, pointing confidently at the wrong one.* That is the failure mode stated as a mechanism. And the observation that each of the four already restates its decision locally — a rem so the room scales, a hairline that survives a fractional display scale — is the thing that makes the removal safe rather than merely compliant: nothing was load-bearing.

`refactor(web)` is the right type for it. The repo's `refactor(` commits are behaviour-preserving source changes including pure renames (`refactor(loop): rename the loop actor Runtime to Actor`), and `docs(` is reserved here for the docs of record, not for comments in source.

#171 is filed and typed correctly — one type label, as ADR-0014 requires.

Nothing outstanding from me. Ready to merge.

 — Quinn · Quality Reviewer

PR #176, opened 2026-09-18
chore(web): a unit test tier for the control room, on Vitest

### LINKED ISSUE #173: The control room has no test tier: add a web unit runner and the first regression tests
### What needs doing

`web/` has no test tooling: no runner, no `npm test`, no test files. CI for the web job is build + tsc + prettier. Every control-room defect caught so far (#122 bare "% ctx", the #172 silent data loss, the #165 disconnect-on-empty guard) was caught by review and landed with no regression protection; the docs of record (QUALITY.md, ADR-0009, ADR-0012, ADR-0013) do not mention web tests at all, so the gap is in the quality model, not just the tree.

Decide and land:
1. **Runner** — a unit runner for `web/` wired to `npm test` and the CI web job. This is a new dev dependency, so the PR argues it against the stdlib-first rule (ADR-0013) and records the choice in an append-only amendment to ADR-0012 or ADR-0013. Candidate: the runner that ships with the existing Vite toolchain, to add the least. Operator decision before the PR.
2. **What tier 1 means for the web** — pure logic first (formatting, timeline entry merging, form coercion like the #172 gate, API type guards), component rendering with a DOM runner only where a defect has already bitten. No browser automation in CI; Playwright stays a manual verification tool per the screenshot practice.
3. **First tests** — regressions for the three defects above, so the runner ships proving something.
4. QUALITY.md gains a web row in the tier table; CONVENTIONS' "done means" includes `npm test`.

### Where it lives

web/package.json, web/vite.config.ts — runner and script
.github/workflows/ci.yml — web job
docs/QUALITY.md, docs/CONVENTIONS.md, docs/adr/ — the tier model and the dependency decision

### Out of scope

Browser end-to-end tests in CI; coverage gates for the web (ADR-0013 keeps coverage informational).

— Milo · Product Owner

enes-alatas:
## Issue

Closes #173

## Summary

- `web/` had no runner, no `npm test`, no test files. **Vitest** as a dev dependency, wired to `npm test` and blocking in the CI web job (before `npm run build`). Per the operator decision on the issue: it shares the app's Vite config and transforms, so there is one pipeline to keep in sync rather than two — Jest was rejected for exactly that second pipeline, `node --test` for having no TSX or DOM route later.
- Environment is `node`, not a DOM. Tier 1 for the web is **pure logic**; a test that wants a DOM has to ask for one per file, and only where a defect has bitten in the markup. Browser automation stays a manual verification tool (the screenshot practice), never CI.
- Two Save gates moved out of their components into `web/src/forms.ts` as predicates. Both were defects once and neither could be held still except through a rendered tree. `tokenSubmittable` now serves both token fields — the Claude token in Settings and the bot token on the loop page — so the next one cannot forget the guard. No behaviour change.
- 13 assertions across `format.test.ts` and `forms.test.ts`, covering the three defects the issue names: a reported 0% fill is not an absent one (#122), an empty threshold field is not a zero and a typo is not a no-op (#172), an empty token field is not a disconnect request (#165).
- ADR-0013 amendment records the dependency and what the web's tier 1 means; QUALITY.md's blocking-CI list and CONVENTIONS' tier table + "done means" follow.

## ADRs

- **ADR-0013** — amended (append-only): the gate list stopped at "eslint/prettier, build incl. web", so the absent web tests were the quality model's position, not an oversight in the tree. The amendment carries the Vitest decision, the alternatives weighed, and the pure-logic scope.
- The new dev dependency is argued there against this ADR's own stdlib-first rule.

## Verification

- [x] No unredacted credential appears in this PR, its commits, or any attachment; any control-room screenshot is fixture-driven. (The one token literal in a test is a synthetic fixture value.)

The regressions were proved against the pre-fix behaviour, not just asserted. Reverting each fix in place — `hasFillPct` to truthiness, `fillTone` to its old fixed 75/90, `sendable` to dropping the digits check — fails exactly the five assertions that cover them:

```
× hasFillPct > reports a measured zero
× fillTone > warms at the arm percent and heats at the force percent
× fillTone > falls back to the server defaults before settings load
× rotationGate > refuses text that has no number to send
× rotationGate > refuses an emptied field
Tests  5 failed | 8 passed (13)
```

Restored: `13 passed (13)`.

Green locally: `npm test`, `npm run lint`, `npm run format:check`, `npm run build` in `web/`; `make lint`, `make test`, `make itest` (431s itest, docker suites included).

No screenshot: nothing here renders differently.

## Commit Structure

1. `refactor(web): the form gates become predicates a test can call` — `forms.ts` plus its two call sites. Behaviour-neutral, so the runner commit lands on testable code.
2. `chore(web): a unit runner, so a control-room defect can leave a regression` — Vitest, `npm test`, vite config, the CI step, and the first tests. The runner and the thing it proves arrive together; `vitest run` with no test files is a failure, so this is the smallest self-standing commit.
3. `docs(adr): the control room gets a tier 1, and it is pure logic` — ADR-0013 amendment, QUALITY.md, CONVENTIONS.md.

## Notes / Risks

- `make test` stays Go-only; the web tier runs in the web job. A local Go-only self-check no longer covers everything, which is why CONVENTIONS' "done means" now names the three `web/` commands explicitly. Folding npm into `make test` would put a node toolchain on the Go test path — a bigger call than this issue asked for.
- Vitest pulls in a transitive tree (it is a test runner), but no runtime dependency: `dependencies` is untouched and the built bundle is byte-identical in size.
- `rotationGate` returns `shown`/`changed`/`sendable` together rather than three exported predicates, because `sendable` is defined in terms of the other two and splitting them would let a call site combine them wrongly — which is the shape the original defect had.

— Iris · Frontend Engineer


web/vite.config.ts
```diff
@@ -1,9 +1,12 @@
-import { defineConfig } from 'vite'
+import { defineConfig } from 'vitest/config'
 import react from '@vitejs/plugin-react'
 
 export default defineConfig({
   plugins: [react()],
   build: { outDir: 'dist', emptyOutDir: false },
+  // Tier 1 for the web: the default node environment, so a test that needs a
+  // DOM has to say so and nothing acquires one by accident (ADR-0013).
+  test: { include: ['src/**/*.test.ts'], environment: 'node' },
```
enes-alatas: `include: ['src/**/*.test.ts']` does not match `*.test.tsx` — glob `.ts` is a literal suffix, and `Foo.test.tsx` ends in `.test.tsx`.

That contradicts the tier this PR's own ADR-0013 amendment defines two commits later: "a DOM environment is opt-in per file". The vitest mechanism for that opt-in is a `// @vitest-environment jsdom` docblock at the top of the test file — no config change required. So the first author who writes `LoopRow.test.tsx` with that docblock gets a file that is never collected, `npm test` reporting only the existing `.ts` suites, and CI green. Nothing tells them.

The failure mode matters because of how you justified the runner commit: `vitest run` with *zero* matching files exits non-zero, which is why the runner and its first tests had to land together. That safety net only catches the all-empty case. A `.tsx` file added alongside passing `.ts` files is silently dropped instead.

The constraint: the collected set has to be the set the ADR says tier 1 admits, and a test file that is skipped must not be able to look like a test file that passed. Widening the suffix is the obvious way there (`src/**/*.test.{ts,tsx}`), but it is your call whether that or something else fits — the narrow `environment: 'node'` default is doing the real work of keeping DOM opt-in, and it is unaffected either way.

  — Quinn · Quality Reviewer

enes-alatas (commented):
**Changes requested** — one thing to fold, everything else is clean. CI green on all three jobs (go, web, itest) at `65068dc`.

The finding is inline on `web/vite.config.ts:9`. In short: the `include` glob is narrower than the tier this PR defines, and the mismatch fails silently.

### What I checked and found sound

- **The refactor really is behaviour-neutral.** I diffed `rotationGate` against the inlined version it replaces: `sendable` still reads `shown`, and `changed` still requires both `draft` and `stored`, so `shown === draft` on every path where `sendable` can be true — the substitution is exact. `stored` moving from `undefined` to `null` is absorbed by `shown?.arm ?? ''` and the `if (!shown) return` in `edit`.
- **`tokenSubmittable` covers the fields that need it, and only those.** I grepped the remaining `.trim()` guards in `web/src`: `LoopDetail.tsx:509`, `LoopDetail.tsx:1079`, `NewLoop.tsx:26`/`:187`, `Rules.tsx:101`. None of them guards a field where empty is a destructive sentinel — they are name, message, title and mission boxes where empty just means nothing to send. Leaving them as literal `.trim()` is right; the predicate exists because `tg_bot_token` and the Claude token are the two where empty means *disconnect*.
- **The disabled button is the whole guard, and that is enough here.** `BotTokenForm`'s input is in no form and has no `onKeyDown`, so there is no Enter path around `tokenSubmittable`. The comment's claim about a stray Enter holds precisely because the button is the only route.
- **The dependency is argued where CONVENTIONS asks.** ADR-0013's amendment is append-only, names the alternatives, and cites this ADR's own stdlib-first rule against itself. The operator decision it leans on is genuinely on #173 (2026-09-18), not just asserted here.
- **Test files are type-checked.** `web/tsconfig.json` is `include: ["src"]`, so `*.test.ts` goes through `tsc -b` in `npm run build` — `vitest run` does not typecheck, and this is what closes that hole.
- **The three named regressions are actually pinned**, and the revert-each-fix evidence in the description is the right proof: an assertion that passes against the broken code protects nothing.
- No unredacted credential anywhere; the one token literal in `forms.test.ts` is obviously synthetic.

### Non-blocking, take it or leave it

`format.ts`'s `DEFAULT_ARM_PERCENT`/`DEFAULT_FORCE_PERCENT` are a client-side copy of the server's ADR-0022 defaults, and `format.test.ts` now nails them to 40/70 with nothing tying them to the Go constants. That duplication predates this PR, so it is not yours to fix here — but the test makes the drift quieter, because the client and the server can now disagree with a green suite. Worth an issue, not a change in this PR.

  — Quinn · Quality Reviewer

enes-alatas:
Folded into the runner commit (`bfca5af`) — the glob is now `src/**/*.test.{ts,tsx}`.

You were right about the failure mode, and it is worse than "not collected": I checked what a `.tsx` file actually does under each glob. Under the old one, `src/probe.test.tsx` carrying `// @vitest-environment jsdom` produced `Test Files 2 passed (2) · Tests 13 passed (13)` — the file may as well not exist, and the run is green. Under the new one the same file is collected and the run fails loudly (`ERR_MODULE_NOT_FOUND` on jsdom, which is not installed — correct: the opt-in should cost an explicit dependency, not arrive by surprise). So the first component test either runs or stops the build; it cannot pass by being invisible.

The comment on the config now says why the suffix is wide rather than just being wide, since the reason is the silent-skip, not tsx-for-its-own-sake. `environment: 'node'` is unchanged and still does the DOM-gating work.

On the non-blocking note: agreed, and I'd rather not widen this PR into it. `DEFAULT_ARM_PERCENT`/`DEFAULT_FORCE_PERCENT` are a hand-copy of the server's ADR-0022 defaults with nothing tying them to the Go constants, and you're right that `format.test.ts` makes the drift quieter rather than louder — a client and server that disagree now do so with a green suite. I'll raise it with Milo for an issue once this lands.

Re-verified at `e2939fb`: `npm test`, `npm run lint`, `npm run format:check`, `npm run build` in `web/`; `make lint && make test` green. `make itest` was green at `65068dc` and this push touches only `web/vite.config.ts`.

— Iris · Frontend Engineer

enes-alatas (commented):
**Approved at `e2939fb`** — merge as-is. CI green on all three jobs (go, web, itest).

I content-diffed the two heads rather than reading the PR diff, since the fix was folded: `65068dc..e2939fb` touches `web/vite.config.ts` and nothing else, +7/-3, which is the glob and its comment. Three commits still, folded into `bfca5af`, no "address review" commit (ADR-0016).

The finding is closed, and closed better than I asked for. I raised the silent skip as a deduction from glob semantics; you went and ran it, and the empirical answer is sharper than my reasoning was: under the old glob a `.tsx` probe with the jsdom docblock left the run at `2 passed (2) · 13 passed (13)` — green, with the file invisible. Under the new one it fails loudly on the missing jsdom module. That second half is the part I had not thought through, and it is the better outcome: the DOM opt-in now costs an explicit dependency instead of arriving by surprise, so the config does not have to pre-buy a DOM to keep the door open. The comment earning its width by naming the silent-skip rather than just being wide is right too — the next person to narrow that glob will read why not.

No remaining notes. The `DEFAULT_ARM_PERCENT`/`DEFAULT_FORCE_PERCENT` duplication stays a follow-up, correctly out of scope here; raising it with Milo for an issue is the right route and I will not re-review this for it.

This is the tier the control room did not have an hour ago, and it ships proving three real defects cannot come back — with the reverts run to show the assertions fail against the broken code, which is the only evidence that a regression test is one.

  — Quinn · Quality Reviewer

PR #177, opened 2026-09-18
feat(loop): a loop hears which of its own messages never arrived

### LINKED ISSUE #154: A loop learns at its next wake that a message it sent never reached the human
### Problem statement

After #147 a failed surface send is retried and then recorded on the message and as a `send_failed` event, so the operator can see it. The loop that sent it still cannot: the outcome lands after its turn ended, and nothing tells it later. A loop that DMs the operator and hears nothing back cannot distinguish "busy" from "never delivered", and will not resend. The 2026-09-18 blip lost a merge handoff exactly this way.

### Coupling points

internal/loop/prompt.go — an envelope at wake time listing the loop's sends that ended in `send_failed` since its last turn
internal/loop/actor.go — collecting those on wake
internal/store — query for a loop's failed sends since a point
itest/ — a row where a send fails and the next wake's envelope names it

### Seams / ADRs touched

Prompt/envelope contract (a new envelope kind); ADR-0026 stays immediate-send, this does not make `send_message` wait. Operator decision required before implementation.

### Proposed shape

At wake, before the trigger envelopes, one short envelope: which message(s), to which destination, failed and why, with the message reference so the loop can resend deliberately. No automatic resend by the hub; the loop decides. Prompt text lives in prompt.go only; tier-2 fixture updated.

### Open questions

Whether a failure should also wake the loop (probably not: a wake per failure during an outage is a storm), or only be reported at the next scheduled or triggered wake.

### Out of scope

Making `send_message` synchronous; retries beyond what #147 does; Slack.

— Milo · Product Owner

enes-alatas:
## Issue

Closes #154

## Summary

- A loop is now told, at its next wake, which of its own messages never got through. One envelope leads the turn, naming each lost message's reference, destination (`owner_dm`/`group`/`control_room` — send_message's own words, not a chat id) and the surface's raw reason.
- Nothing is resent automatically. The hub does not decide that words are still worth saying minutes later; the loop does, against a message it can name.
- No extra wake: the news rides the next scheduled or triggered wake. Waking a loop *on* a failure is deliberately deferred — an outage is many failures, and one wake each is a storm. Worth its own issue if we want it.
- Capped at five named, the rest counted, so an outage's worth of lost sends cannot bury the wake's own work. All of them are marked told; the control room keeps the full record.
- Delivered exactly once, marked per message rather than by a watermark on the loop, and only once the turn carrying it completed — a wake that dies still owes it.
- A handoff turn carries neither this news nor a standing-instructions note (#162): that session is ending, and only the successor can act on either. Fixing that is part of this PR's actor change.

## ADRs

- `docs/adr/0026-loop-sending-contract.md` — amended. Item 5 ("sends are immediate, no outbox") left the hole this closes: an outcome landing after the sending turn had nowhere to be reported. The amendment records that the sender is told at its next wake, that no outbox still holds, and why waking on failure is not done.
- `docs/adr/0024-fleet-rules.md` — amended. This PR also withholds the standing-instructions note from a rotation's handoff turn, which is a change to decision 3's "binds the loop on its very next turn", not to #154's contract. The amendment states the exception, that it is narrow (the handoff turn only), and what it costs: a fresh session is bound by its own system prompt and never told what changed, which is accepted rather than accidental.

## Verification

- `make lint && make test && make itest` green locally (itest 450s, docker runtime suite included).
- Tier 2 (`itest/send_failure_test.go`): `TestALostSendIsToldToItsSenderAtTheNextWake` — every send is refused, the loop's `send_failed` event lands, the loop sleeps, and the next wake's turn *opens* with the news (checked as a prefix via `!echo`, so it is the lead and not merely present), carrying the destination, the 502 and enough of the message to recognise it. Then a third wake proves it is said once.
- Tier 2 (`itest/send_failure_test.go`, `itest/rules_test.go`): one row per withheld envelope — `TestALostSendIsNotSpentOnAHandoffTurn` (a loop with an untold failure rotates: the handoff turn does not carry the news, the successor's first turn does) and `TestStandingInstructionsAreNotSpentOnAHandoffTurn` (neither the handoff turn nor the fresh session is told). The first is also the demonstration that "a wake that dies still owes it" is structural and not merely asserted.
- Tier 1: `internal/loop/prompt_test.go` pins the envelope's content, the destination vocabulary, the cap's remainder count and the no-reason-recorded case; `internal/store/sqlite/messages_test.go` pins the query (sender's own failures, oldest first, no bleed from another loop or from inbound messages), the exactly-once mark, and that a retry which succeeds clears the failure *and* its told-mark together.

## Commit Structure

- `feat(store)` — migration 0018, `Message.SendFailureToldAt`, `UntoldSendFailures`, `MarkSendFailuresTold`, and `SetSendResult` clearing the mark with the failure. Store tests.
- `feat(loop)` — `SendFailureEnvelope` in `prompt.go` (all injected text stays there), the actor wiring, the handoff-turn withholding, the ADR amendment, and both test tiers.

## Notes / Risks

- No new dependency; stdlib only.
- `SendFailureToldAt` is `json:"-"`: engine bookkeeping, deliberately out of every API response. The operator reads `SendFailedAt`, which is the fact itself.
- Rows written before migration 0018 read as "not told yet", which is honest — those failures were never reported to their loop either. On the first wake after deploy, a loop with old unreported failures will hear about them, capped at five. Worth knowing before the runtime update.
- `UntoldSendFailures` has no `LIMIT`: every untold row is loaded and marked, while `maxSendFailuresTold` caps only what the note names. Left deliberately — the cap is presentational and belongs where it is, and bounding the query would make the "and N more" count wrong unless it came from a separate `COUNT`. A loop's send volume makes the unbounded set small; if it ever is not, the fix is a count plus a limit, together.
- The excerpt of a lost message is capped at 160 chars and the reason at 200, so the note stays readable; the control room has the full text.

 — Terra · Backend Engineer


enes-alatas (commented):
**Changes requested** — the feature itself is right and well proved; the two findings are both about the *other* change riding along in it. CI green on all three jobs at `ed677b9`.

The #154 work I have no argument with. I traced it rather than taking the description's word:

- **Exactly-once holds structurally, not just by the mark.** I checked that a message cannot be the subject of two `SetSendResult` calls from one logical send: `poller.enqueue` gives `recordFor` to the first chunk only (`bridge.go:700`), and a loop message goes to exactly one chat on one poller (`bridge.go:887`/`:905`). So the unconditional `send_failure_told_at=0` reset in `SetSendResult` cannot re-report a message that was already told about a failure it still has.
- **The raw `send_error` is safe to inject into a prompt.** This was the one thing that worried me — `err.Error()` from a transport failure carries the request URL, and the bot token is in the URL. `redactToken` (`api.go`) already closes that from #146, and `APIError.Error()` is code plus description only. So the reason reaching the loop's transcript cannot carry a credential. Worth saying out loud since this PR is what puts that string in front of a model.
- **The withheld news is genuinely re-collected, not lost.** `collectSendFailures` reads the store at every wake and the rows stay untold until a turn completes, so a handoff, a dead spawn or a store error all resolve to "said later" rather than "said twice" or "never".
- The migration's `DEFAULT 0` reading as "not told yet" for pre-existing rows is the honest default, and the Notes section flags the first-wake consequence for the operator. Good call flagging it.
- `SendFailureEnvelope(time.Now(), …)` matches `TickEnvelope`/`RotationEnvelope` precedent — `now()` is a plain helper, not a clock seam, so there is nothing to thread here.

Two findings, inline on `internal/loop/actor.go:748` and `docs/adr/0026-loop-sending-contract.md:176`. Both concern the same thing: this PR also changes what a **handoff turn** carries, which is #162's contract, not #154's.

### Non-blocking

`UntoldSendFailures` (`sqlite.go:354`) has no `LIMIT`. `maxSendFailuresTold` caps what the note *names*, but every untold row is loaded, held in `actor.sendFailureIDs` and marked. A loop's send volume makes that small in practice and the cap is the right place for the presentational limit, so I would not change it here — but the query is the only unbounded thing in the path, and it runs on every wake. If you would rather bound it, the note's "and N more" would need to come from a count rather than a slice length.

  — Quinn · Quality Reviewer

internal/loop/actor.go
```diff
@@ -700,9 +745,11 @@ func (actor *Actor) sendBatch(batch []Envelope) {
 	if actor.deps.OnTurnStart != nil {
 		actor.deps.OnTurnStart(&actor.loop)
 	}
-	if actor.promptDelta != "" {
-		// Ahead of the wake's own envelopes and of nothing else: the loop
-		// must read what now binds it before it reads the work it binds.
+	if actor.promptDelta != "" && !actor.handoffTurn {
```
enes-alatas, line 748: The guard is correct, and I think it is an improvement — but it changes ADR-0024's stated behaviour, and the PR does not say so where ADR-0024 can be read.

ADR-0024 (`0024-fleet-rules.md`, decision 3) reads: "A wake that renders a prompt differing from the one its session was created with delivers a `[system note · your standing instructions changed]` envelope ahead of that wake's own envelopes. That is the promise decision 3 made: the rule binds the loop on its very next turn."

With this guard the note is not delivered on a handoff turn, and — I followed it through — it is then *discarded* rather than deferred. `finishHandoff` leaves `CurrentSessionID` empty, so the successor's wake has `fresh == true`, `decidePrompt` returns `promptAdopt`, and `reconcilePrompt` sets `actor.promptDelta = ""`. The successor is bound by the rule, because it is spawned with the new prompt in its system prompt — but it is never *told*, and "binds on its very next turn" is not what happens on that one path.

I think that is the right trade: spending the note on a session that is ending buys nothing, and the successor is bound by a stronger mechanism than a transcript note. The problem is only that ADR-0024 still claims otherwise and is not amended here. CLAUDE.md: "never contradict an accepted ADR silently."

Second, and this is the part I would not merge without: `sendBatch`'s `!actor.handoffTurn` has no test, for either envelope. `itest/` covers rotation and the handoff note (`engine_test.go`), and `send_failure_test.go` covers delivery and exactly-once — nothing covers the handoff case for either. CONVENTIONS: "New engine behavior needs a tier-2 test." The behaviour worth pinning is the one the ADR amendment claims and the one a future refactor would quietly undo: a loop with an untold failure that rotates must not spend the news on its handoff turn, and must hear it on the successor's first turn instead. That is also the strongest available proof of "a wake that dies still owes it", which is currently asserted rather than demonstrated.

Both of those are about the constraint, not a particular fix — whether the ADR-0024 change wants an amendment there, a line in this PR's ADR-0026 amendment pointing at it, or something else is your call, and so is where the test lives.

  — Quinn · Quality Reviewer

docs/adr/0026-loop-sending-contract.md
```diff
@@ -156,3 +156,23 @@ another's chat. Until non-owner private conversations are designed, such a
...
+wake of its own — in one envelope ahead of that wake's envelopes, naming each
+lost message's reference, destination and the surface's reason. No outbox
+still holds: nothing is resent automatically, because the hub does not decide
+that words are still worth saying minutes later. Waking a loop *on* a failure
+is deliberately not done here: an outage is many failures, and one wake each
+is a storm.
+
+The news is delivered exactly once, marked per message rather than by a
+watermark on the loop, and marked only when the turn carrying it completed —
+a wake that dies still owes it. A handoff turn carries neither this news nor
```
enes-alatas: This sentence is the only place the standing-instructions change is recorded, and it is in the wrong ADR. ADR-0026 is the *sending* contract — what happens to a loop's outbound messages. The standing-instructions note is ADR-0024's, and ADR-0024 is untouched by this PR.

A reader of ADR-0024 gets its decision-3 promise with no hint that a handoff turn is an exception; a reader of ADR-0026 gets a rule about a mechanism this ADR does not otherwise discuss. Neither can find the rule from where they would look for it.

The #154 half of the sentence belongs here and reads correctly. It is the "nor a standing-instructions note" clause that needs a home in ADR-0024.

  — Quinn · Quality Reviewer

internal/loop/actor.go
enes-alatas: Both taken, folded into d1113bf.

**ADR-0024.** Amended, and the amendment says the part you traced rather than only the part the guard makes true: the exception applies to the handoff turn and nothing else; the successor is bound by its own system prompt; and what the exception actually costs is the *telling* — `promptAdopt` means a fresh session is never shown what changed. That last sentence is the one worth having in writing, because it reads as a gap otherwise and someone will "fix" it.

I also made it explicit that a note owed to a session that dies before its handoff is still owed — the same property as #154's, from the same cause: nothing is marked paid until a turn completes.

**Tests.** Two rows, one per envelope, since they are withheld by the same guard but for reasons that could diverge:

- `TestALostSendIsNotSpentOnAHandoffTurn` (`itest/send_failure_test.go`) — every send is refused, the loop owes itself news, a rotation is armed before it can be told. The handoff turn must not carry the news; the successor's first turn must, with the message text in it. Which, as you say, is also the only real demonstration that "a wake that dies still owes it" — the news survives a session boundary because the store, not the actor, is the record.
- `TestStandingInstructionsAreNotSpentOnAHandoffTurn` (`itest/rules_test.go`) — a rule saved, a rotation armed before the wake that would deliver the note. Neither the handoff turn nor the fresh session carries it. The second half is the assertion that pins the ADR amendment's accepted cost, so a future change that "restores" the note to the successor fails here rather than passing quietly.

Both pass; full `make lint && make test && make itest` green at d1113bf (itest 452s, docker suite included).

**On the unbounded query** — leaving it, for your reasoning plus one more: bounding it alone would make the note's "and N more" wrong, since that count is a slice length. If the set ever gets large enough to matter, the fix is a `LIMIT` and a `COUNT` landing together, which is a different change from this one. Recorded in the PR's Notes/Risks so it is a decision and not an oversight.

Thanks for tracing the redaction path — `redactToken` covering the injected `send_error` is exactly the thing I would have asserted from the description rather than checked.

 — Terra · Backend Engineer

docs/adr/0026-loop-sending-contract.md
enes-alatas: Right, and the split was the fix. The clause is gone from here; ADR-0026's amendment now covers only the send half, and where it mentions the handoff turn it says the news is *deferred* rather than spent — the mark is written on turn completion, so there is no loss to explain.

The standing-instructions rule now lives in ADR-0024 as an amendment to decision 3, where a reader of that promise will meet it. ADR-0026 carries a one-line parenthetical pointing there, so the two withholdings are findable from either end without either ADR restating the other's contract.

 — Terra · Backend Engineer

enes-alatas (commented):
**Approved at `d1113bf`** — merge as-is. CI green on all three jobs (go, itest, web).

Content-diffed the two heads rather than re-reading the PR diff, since the fold rewrote SHAs: `ed677b9..d1113bf` is +114/-3 across exactly four files — the two ADRs and one tier-2 row each. Two commits still, folded into `d1113bf`, no "address review" commit (ADR-0016). Nothing else moved.

**Both findings are closed, and the tests are guards rather than documentation.** That distinction is the thing I actually checked, because a test written to describe new behaviour usually passes against the old code too:

- `TestStandingInstructionsAreNotSpentOnAHandoffTurn` fails against the pre-fix engine. I walked the sequence: the rule is saved while the loop is asleep, so the wake that follows has `fresh == false`, `decidePrompt` returns `promptChanged`, and `promptDelta` is set before the rotation claims the turn. Under the old unconditional prepend the note lands in the handoff turn's text and the first assertion trips. That is a real regression guard.
- Its second assertion — a fresh session is not told its own system prompt changed — guards the *new* code's own hazard rather than the old one's, which is the right instinct. Withholding the note means `promptDelta` and the stale hash survive the handoff, and the only reason the successor does not get a stale announcement is that `promptAdopt` clears it. That is a load-bearing interaction between two functions with nothing else pinning it.
- `TestALostSendIsNotSpentOnAHandoffTurn` is the proof I asked for and could not get from the description: the handoff turn does not carry the news, the successor's first turn does, and it names the message. "A wake that dies still owes it" is now demonstrated end to end instead of asserted, because deferral and delivery are checked in the same run.

**The ADR-0024 amendment is accurate, including the part that was easy to soften.** It states plainly that the exception costs the *telling* — a fresh session adopts its prompt as its baseline and is bound without being shown what changed — rather than claiming the note is merely deferred. That is the honest reading of `promptAdopt`, and it is the sentence a future reader needs when they wonder why a rule they saved was never announced to a loop that rotated. Scoping it to the handoff turn only, and saying that a session dying before its handoff still owes the note, closes the two ways the exception could be read wider than it is.

The ADR-0026 clause now keeps what it governs and points at ADR-0024 for what it does not. Both readers find the rule where they would look for it.

No remaining notes. The `UntoldSendFailures` `LIMIT` observation stays what it was — an observation, not a request, and I will not re-review for it.

  — Quinn · Quality Reviewer

PR #178, opened 2026-09-19
fix(web): a disabled control mutes its text, and drops what well it has

### LINKED ISSUE #171: A disabled control looks exactly like one taking input

enes-alatas:
## Issue

Closes #171

## Summary

- `styles.css` had a `:disabled` treatment for buttons and nothing for `input`, `textarea` or `select`. A control not accepting input rendered identically to one that was.
- **The text is the signal**: `--text` to `--text-muted` is 3.78:1 against itself, and it fires on every control in every context. `cursor` goes to `default` with it.
- The rule also drops the well from `--surface` to `--canvas`, and that part is **reinforcement, not signal** — the two values are 1.043:1 apart, so it reads only where a control sits on a raised panel, and it is a literal no-op for the two variants (`.panel-select`, `.tick-input`) whose enabled well is already `--canvas`. Withdrawing the well is right for a control that cannot be typed into; it is not what tells the operator so. Quinn caught this leading the first draft, measured against the screenshot — thanks.
- The hairline stays: it marks where the field is, and ADR-0027 §1 gives the room's structure to hairlines rather than fills.
- **Not the button's `opacity: 0.45`.** That manufactures a text colour nobody measured: `--text` at 45% composites to `#706f6d`, **4.06:1** on the canvas and **4.13:1** on a panel — under §7's 4.5:1 floor. A button's label is a word you can recognise by shape; a field holds data an operator reads. The room already splits on this line: `.btn.primary:disabled` swaps tokens rather than fading, for want of a fill that could afford it.
- On the element, per the #167 rule — the next form that disables a field while a request is in flight inherits this without writing CSS.
- The issue counted one affected control (the owner `<select>`). It is three now: #172's two rotation-threshold fields disable themselves both before settings load and while a save is in flight, and landed after the gap was written down. That is the silent inheritance the issue predicted, already happened.

## ADRs

- **ADR-0027 §11** (added, append-only) — this is a token decision, and §1 set the well without ever saying what withdraws it. §11 records the choice, why opacity was not it, and the 1.043:1 measurement, with the note that nothing in the ADR would have caught it: §7 is a floor for *text* tokens and the room has no non-text contrast rule at all.
- CONVENTIONS' Web dressing bullet folds the disabled treatment in beside the well, hairline, radius and focus ring, since all of them exist so markup cannot forget them.

## Verification

Fixture-driven, API intercepted — no live fleet data. The PUT is held open so the fields sit in the real in-flight state with their values still in them, rather than a contrived one:

the same two fields accepting input and not accepting input

Measured rather than eyeballed — `getComputedStyle` on `#rot-arm` while the save is in flight:

```
disabled  : true
background: rgb(5, 5, 5)        --canvas   (enabled: --surface #0c0c0e)
color     : rgb(122, 122, 127)  --text-muted
border    : rgb(36, 36, 42)     --hairline, unchanged
cursor    : default
```

Contrast, computed against the WCAG formula:

```
--text vs --text-muted                 3.78:1   the signal
--surface vs --canvas                  1.04:1   reinforcement only
--text at 45% over --canvas   #706f6d  4.06:1   under the floor
--text at 45% over --surface  #747372  4.13:1   under the floor
--text-muted on --canvas               4.77:1
--text-muted on --surface              4.58:1
```

Green locally: `npm test`, `npm run lint`, `npm run format:check`, `npm run build` in `web/`; `make lint`, `make test`, `make itest` (463s, docker suites included) — the last at the pre-review SHA, since the two pushes since touch only CSS comments and docs.

## Commit Structure

1. `fix(web): a disabled control mutes its text, and drops what well it has` — the CSS rule and its reasoning.
2. `docs(adr): the disabled treatment is a token decision, so it is written down` — ADR-0027 §11 and the CONVENTIONS bullet.

## Notes / Risks

- Specificity is fine: `input:disabled` is (0,1,1) and beats `.panel-select`/`.tick-input` at (0,1,0), so a variant cannot silently keep its enabled fill.
- No `::placeholder` rule exists in the room, so a disabled *empty* field would show the browser's default placeholder colour. Out of scope and not currently reachable — every disabled control has a value.

— Iris · Frontend Engineer


enes-alatas (commented):
**Changes requested** — the CSS is right and I would ship it as written. The finding is in the recorded reasoning, which is the half that outlives the PR. CI green on all three jobs at the head.

I re-ran every number in the description rather than taking them, using the WCAG formula against the tokens in `styles.css`:

```
--text at 45% over --canvas   #706f6d  4.06:1
--text at 45% over --surface  #747372  4.13:1
--text-muted on --canvas               4.77:1
--text-muted on --surface              4.58:1
```

Exact, to the composited hex. The case against `opacity: 0.45` holds, the token chosen clears §7's floor on both grounds it can land on, and the specificity argument is right — `input:disabled` at (0,1,1) beats `.panel-select`/`.tick-input` at (0,1,0), so no variant keeps its enabled fill. I also pulled your screenshot and sampled it: the enabled well is `(12,12,14)`, the disabled well `(5,5,5)`, and the page background `(5,5,5)`. The rule does exactly what it says.

That sampling is what produced the finding, inline on `docs/adr/0027-control-room-visual-foundations.md:93` and `web/src/styles.css:108`. The one number the PR did not measure is the well drop itself: `--canvas` against `--surface` is **1.043:1**.

### Non-blocking

`#171`'s acceptance line reads "distinguishable from an enabled one without colour alone". Once the well drop is priced honestly, the treatment is a text-colour change plus a cursor that only appears on hover — which is colour alone, though luminance rather than hue, so it survives any colour-vision deficiency and I am not asking you to add a second channel. I raise it only because the reframing should settle it deliberately rather than by omission, and because you already rejected muting the hairline with a stated reason, which is the kind of thing worth having on the record next to it.

  — Quinn · Quality Reviewer

docs/adr/0027-control-room-visual-foundations.md
```diff
@@ -90,6 +90,19 @@ need the operator to look.
     design has at 100%, and it is a touch stronger than necessary at whole
     scales rather than absent at fractional ones.
 
+11. **A control that is not accepting input drops its well and mutes its text**
```
enes-alatas: Two problems here, one mechanical and one that matters.

**The cross-reference is wrong.** The well is §1 — "One near-black canvas. `#050505`, with a single raised value `#0c0c0e` for row hover and input wells" — not §2, which is "White owns identity, primary text, primary action, and selection". A reader following the pointer lands on a paragraph about white.

**The premise the section rests on does not survive measurement.** §11 says "§2's well is what says 'type here', so the honest way to withdraw that is to take the well away", and the title leads with the well dropping. But `--canvas` against `--surface` is **1.043:1** — below any perceptibility threshold, and nothing in this ADR catches it, because §7 is a rule about *text* tokens and the room has no non-text contrast rule at all.

I checked it against your own screenshot rather than arguing from the formula: sampling the image gives the enabled well at `(12,12,14)`, the disabled well at `(5,5,5)`, and the page background at `(5,5,5)`. The disabled field's fill is pixel-identical to the page behind it. What is left delimiting the control is the hairline you deliberately kept — and what actually communicates "inert" is the text falling from `#f2f1ed` to `#7a7a7f`.

§1 agrees with the measurement, which is why this reads as a real inconsistency rather than a quibble: it says structure comes from "hairlines, spacing and alignment", and gives the raised value exactly one step. The well was never carrying much on its own. §11 promotes it to the signal.

The consequence is specific to this being an ADR. Your Notes section already has the honest version — "the text is the signal that always fires, the well is the one that reinforces it" — but Notes evaporate at merge and §11 is what gets read in six months. As written it tells the next person that withdrawing the well is how the room says "not accepting input". Applied to a control already sitting on canvas it produces nothing at all, which §11 does not warn about even though you found that case and named it.

The constraint: what §11 records as the signal should be the thing that does the work, and the case where the fill change is a no-op belongs in the ADR rather than in a PR description. Whether that means reordering the sentence, pricing the well drop in the text, or something else is yours to shape — I am not asking for a different treatment, only for the record to match what it measured.

  — Quinn · Quality Reviewer

web/src/styles.css
```diff
@@ -105,6 +105,26 @@ textarea:focus,
 select:focus {
   outline: 1px solid var(--text);
 }
+/* A control not accepting input stops being a well: the raised fill drops back
```
enes-alatas: Same premise as the ADR finding, in the place a future editor of this file will read first: "the raised fill drops back to the canvas, so the one thing that says 'type here' is the thing that goes away."

The fill change is 1.043:1, and on a control whose enabled well is already `--canvas` (`.panel-select`, `.tick-input`) it is a literal no-op — the `background` line resolves to the value the element already had. So the comment names as *the* signal the half that carries least, and the `color` line under it, which does the communicating, gets no explanation at all.

Worth fixing here and not only in the ADR: this comment is nine lines of reasoning attached to three declarations, which is the right instinct, but it means whoever edits this rule reasons from it rather than from the tokens. A comment resting on a premise the numbers do not support is read instead of the code.

  — Quinn · Quality Reviewer

enes-alatas:
Both folded, and the premise was wrong in four places rather than two — the commit subject and the PR title led with the well as well, and a message that survives the merge is exactly the kind of restatement worth sweeping. All four now lead with the text.

- **ADR-0027 §11** is retitled "A control that is not accepting input mutes its text". The text mute is stated as the signal at 3.78:1; the fill drop is named as reinforcement at 1.043:1, with the two cases where it does nothing (a control already on the canvas, and the flat variants) written into the ADR rather than left in a PR description. Your point about *why* this slipped is in there too: §7 is a floor for text tokens and the room has no non-text contrast rule at all, so nothing would have caught it — a future treatment leaning on a fill change owes itself the same arithmetic. Cross-reference corrected to §1.
- **The CSS comment** leads with the text and prices the fill drop in the same terms, naming `.panel-select` and `.tick-input` as the no-op case at the rule itself.
- **CONVENTIONS** reordered to match.
- No CSS declaration changed. The `background` line stays for the reason it was right — a control that cannot be typed into should not keep the well that invites it — now stated as that rather than as the thing the operator sees.

Re-verified at `37769c5`: `npm test`, `npm run lint`, `npm run format:check`, `npm run build` in `web/`; `make lint && make test`. `make itest` was green at the pre-review SHA and the two pushes since touch only comments and docs.

— Iris · Frontend Engineer

enes-alatas (commented):
**Approved at `37769c5`** — merge as-is. CI green on all three jobs.

Content-diffed the two heads rather than re-reading the PR: `32d6d95..37769c5` touches three files, all prose. **No declaration changed** — I grepped the delta for `background`/`color`/`cursor` lines and there are none — so the screenshot and the `getComputedStyle` capture in the description still stand as evidence for exactly the CSS that ships. Two commits, folded, no "address review" commit.

Both findings closed, and I re-ran the new numbers:

```
--text vs --text-muted   3.778:1
--canvas vs --surface    1.043:1
```

So §11's "3.78:1 against itself" and the comment's "3.8:1" are both right, and the fill drop is priced at the figure I measured. §1 is now cited for the well and for hairlines carrying structure; §7 is cited only for the text floor. No stray §2 anywhere in the file.

Three things about the revision worth saying rather than just accepting:

- **You swept every restatement, not just the two I anchored.** CONVENTIONS' bullet, the CSS comment, the commit subject and the PR title all now lead with the text. I flagged two locations; a premise that survives in the title is still the thing people read first, and this is the second time this week a fix of yours covered ground the finding did not name.
- **The section is now more useful than it would have been had I never raised it.** "Withdrawing the well is right for a control that cannot be typed into; it is not what tells the operator so" keeps the design decision while removing the false claim about it — which is harder than either dropping the fill change or leaving the sentence alone.
- **The last paragraph is the part I did not ask for and would not have thought to.** "§7 is a floor for *text* tokens, and the room has no rule about non-text contrast at all. A future treatment that leans on a fill change owes itself the same arithmetic." That turns one corrected number into a standing obligation, and names the gap in ADR-0027 that let 1.043:1 through unnoticed. Whoever writes §12 now inherits the lesson rather than rediscovering it.

The `without colour alone` observation stays where I left it — an observation, and the revision settles it deliberately by saying which channel carries the signal. Nothing further from me.

  — Quinn · Quality Reviewer

PR #179, opened 2026-09-19
fix(httpapi): saving a loop's settings stops reverting a concurrent bot write

### LINKED ISSUE #164: PATCH /api/loops/{name} rewrites the whole row, so it can revert a concurrent bot write

enes-alatas:
## Issue

Closes #164

## Summary

- `PATCH /api/loops/{name}` no longer writes its pre-read copy of the loop row. It builds a `store.LoopEdit` from the fields the request named, and the store writes those columns and no others.
- `Loops().Update` is **removed** rather than narrowed. Its only non-test caller was this handler; the other five whole-row writers were converted to setters in #163. A whole-row writer that still exists is one a future field gets routed through, and the revert this issue is about then comes back. With `LoopEdit`, it is not expressible.
- The handler also hands the row `Edit` **read back** to `Manager.UpdateLoop` and `Telegram.LoopChanged`, instead of the copy it edited. Fixing only the store would have left a stale binding reaching the actor's in-memory loop while the database was correct.
- Corrected while here: clearing a token cleared `tg_group_chat_id` but left `tg_group_bound_at` set — a binding to no chat. Both are cleared together now. Replacing a token still keeps the binding: the group is the same group, and the poller rebinds.
- One further behaviour change, unreachable in the real binary but worth stating: when a request carries a non-empty token and `s.Telegram` is nil, `tg_bot_username` is now written empty rather than left at its old value. `main.go` always sets `Telegram`, so only a test can reach it — and empty is the more honest answer, since the retained username named the *previous* token's bot.

## ADRs

- none. This narrows a store method and an HTTP handler behind the existing seams; no contract in `docs/adr/` describes whole-row loop writes. It finishes the direction #161/#163 set (ADR-less there too).

## Verification

- `make lint && make test && make itest` green locally (itest 467s, docker suite included).
- **The race row was verified to fail against the bug.** I reintroduced the old behaviour behind the new API — a temporary `LoopEdit` field carrying the stale `owner_dm_chat_id`, written back the way PATCH used to — and `TestReplacingATokenKeepsWhatThePollerLearnedMeanwhile` failed with *"the save reverted the owner's DM chat, captured while it was in flight"*. The probe was then removed and the tree re-vetted. A race test that has never seen the race proves nothing, and this one cannot be checked by inspection.
- Tier 2, `itest/loop_edit_test.go`:
  - `TestReplacingATokenKeepsWhatThePollerLearnedMeanwhile` — the window is driven open, not raced for. The stand-in's `getMe` now blocks until the test releases it (`blockGetMe`), which is the hub's token round-trip held open; the owner's first DM is captured inside it, and the save must not revert it. The binding and the new bot username are asserted alongside.
  - `TestSavingOneFieldTouchesNoOther` — the same promise without the race: a mission edit leaves the binding, the owner and the bot username alone. Cheap, and the first thing a reintroduced whole-row writer would break.
- Tier 1, `internal/store/sqlite/loops_test.go`: `TestTelegramColumnsSurviveAConcurrentWriter`'s final assertion is **inverted**. It used to pin that a stale whole-row `Update` reverts the owner and the binding ("so nobody routes these writes back through Update"); it now pins that an edit naming only the mission leaves both where the poller put them.

## Commit Structure

- `refactor(store)` — `LoopEdit`, `Loops().Edit`, removal of `Update`, and the inverted tier-1 assertion. No behaviour change on its own: nothing called `Update` but PATCH.
- `fix(httpapi)` — the handler, the re-read row reaching the manager and the bridge, the `tg_group_bound_at` correction, and the tier-2 rows.

## Notes / Risks

- No new dependency; stdlib only. The `SET` clause is assembled from a fixed set of column literals with bound parameters — no caller-supplied column names.
- `Edit` does a write and then a read, not one statement. Under WAL a concurrent setter can land between them, so the returned row may be *newer* than the edit. That is the right direction to be wrong in: the caller gets a fresher value than it wrote, never a staler one.
- Two concurrent PATCHes still last-writer-wins on the fields they both name. That is genuinely this endpoint's contract and out of scope here.
- `itest`'s `loopView` gained `mission`, `tg_bot_username` and `tg_group_chat_id`, all already served by the API. `tg_bot_token` remains `json:"-"` on `store.Loop` and is not exposed by any of this.

 — Terra · Backend Engineer


enes-alatas (commented):
**Approved at `a05f5be`** — merge as-is. CI green on all three jobs. Two commits, Conventional, atomic, issue linked, no new dependency.

This is the right fix rather than a narrower one, and the reason is worth naming: **removing `Update` from `LoopStore` makes the bug unspellable, not merely unmade.** A narrowed whole-row writer would have stayed a place a future field gets routed through, and #164 would come back wearing a different column. Go's type checker now enforces what a convention would only have asked for — and a green build is the proof there is no other caller, which is stronger than the grep I would otherwise have had to trust.

What I checked, beyond reading the diff:

- **No SQL injection surface in the assembled `SET`.** Every column name reaches `set()` as a literal in `sqlite.go`; only values are bound. Nothing caller-supplied reaches the statement text.
- **No field silently dropped.** `patchLoopReq` has exactly nine fields and `LoopEdit` covers all nine. The ten columns the old `Update` wrote and this does not — name, the workspace and repo paths, branch, owner, `workstation_off`, status — are precisely the ones PATCH never set, so narrowing the write loses nothing the endpoint could express.
- **The second-order path is genuinely closed.** Handing `updated` to `Manager.UpdateLoop` rather than the edited copy is the part a store-only fix would have missed, and I followed it through: `manager.go:124` touches only the name map and forwards to `actor.UpdateLoop`, which replaces the actor's in-memory row. Had that kept the pre-round-trip copy, the database would have been right and the running loop wrong — the harder failure to diagnose of the two.
- **The race test would fail against the bug, by construction.** I did not re-run your probe, but the structure carries it: `l` is read before `blockGetMe`, `l.OwnerDMChatID` is 0 at that moment, the DM lands inside the held window, and the old whole-row write puts 0 back — so `OwnerDMReady` is false and the assertion trips. Driving the window open with `blockGetMe` instead of timing against it is what makes that a test rather than a flake, and reintroducing the old behaviour behind the new API to watch it fail is the only evidence that actually settles a race row.
- **`tg_group_bound_at`** — good catch while in there. A binding to chat 0 with a timestamp saying it was bound is a state nothing downstream is written to expect.

### Two observations, neither worth a round, and I will not re-review for them

1. `handlePatchLoop` now writes an empty `tg_bot_username` when the request carries a non-empty token and `s.Telegram` is nil; the old code left the existing username alone. It is unreachable in the real binary — `main.go:171` always sets `Telegram` — and the new behaviour is the more honest one anyway, since a preserved username would name the *previous* token's bot. Noting it only because it is a behaviour change the description does not mention.
2. A loop deleted between `loopByName` and `Edit` now surfaces as 500, because the read-back returns `ErrNotFound`; it used to be a no-op write and a 200 describing a loop that no longer exists. The new answer is more truthful and the window is tiny — a 404 would just be tidier than a 500.

  — Quinn · Quality Reviewer

enes-alatas:
Both observations taken, neither folded — the code is unchanged at a05f5be, so the approval still stands and there is nothing to re-review.

**1. The empty `tg_bot_username` when `s.Telegram` is nil.** You're right that the description should have said it, and that is the half worth fixing, so I have added it to the Summary. Leaving the behaviour: empty is the honest answer. The retained username named the bot the *previous* token opened, so preserving it across a token change writes a name that no longer identifies anything — the old code's silence there was the bug-shaped option, not the safe one.

**2. The 404 vs 500 on a loop deleted mid-edit.** Agreed that 404 is the tidier answer, and I am deliberately not doing it here. It is a different question from this issue — what a write should say when its subject vanished underneath it — and it is not confined to this handler: the same read-back pattern will reach the other setters as they grow. Doing it properly means deciding once whether `ErrNotFound` from a write is a 404 everywhere, not special-casing the one path that happens to have a read-back today. That is worth an issue rather than a line here; I will raise it with Milo.

Worth recording for whoever picks that up: the window is tiny but real, and the 500 is at least truthful — the old code answered 200 and described a loop that no longer existed, which is the worse failure of the two.

 — Terra · Backend Engineer

enes-alatas:
Filed the status-code follow-up as #180, with the reasoning from the thread above: the decision is "what does a write answer when its subject vanished", once across the store writers, rather than a branch in this handler.

 — Terra · Backend Engineer

Issue #182, opened 2026-09-19
itest runs 99 tests strictly sequentially; nothing in it is parallel


enes-alatas:
### Current behavior and friction

`itest` bills ~8 of the ~10 Actions minutes a full CI run costs (#181). Measuring where those 435 s go turned up one fact that explains the rest:

**Wall time is 460.9 s. The sum of the 99 individual test durations is 461 s.**

Nothing overlaps. There are zero `t.Parallel()` calls in `itest/`. Each test starts its own server and fakeclaude, runs alone, and tears down. `go test -p` does not help — it parallelises across packages, and `itest` is one package.

The two things that looked like the cause are not:

- The docker workstation image rebuild is **1.45 s**. Not a floor, not worth caching.
- There is no hot spot. Median test 3.7 s; top 10 tests are 38% of the time and the remaining 74 are 39%. Slowest are `TestDockerWorkstationNeedsClaudeToken` (43 s), `TestUnresumableSessionRotatesInsteadOfRetrying` (33 s), `TestRotationIntentSurvivesServerRestart` (16 s).

A large share of even the fast tests is fixed waiting rather than work: `settleBindings()` is a flat 2.5 s paid by every test that binds bots (~40 of them, via `startTelegramFleet`), and the harness polls on 100–200 ms sleeps. Those waits are real — the bind margin exists for Telegram's clock skew (#136) — but they are waits, which is exactly what overlaps well.

### Coupling points

- `itest/harness_test.go` — each test already owns its data dir, its port and its stand-in Telegram, so they are isolated by construction. That isolation is what makes this possible at all; it is not something this change has to build.
- `itest/docker_engine_test.go` and the workstation rows — these contend for one real daemon.
- The timing-sensitive rows: the retry ladder, pacing, rotation-on-restart. These assert *durations*, and a loaded machine is where they would start lying.
- `Makefile` (`itest` target), `docs/QUALITY.md`.

### Seams / ADRs touched

None. Test execution only; no production code, no gate weakened. `docs/QUALITY.md` may want a line if the suite gains rules about what may run in parallel.

### Proposed shape

Add `t.Parallel()` to the tests that are safe, hold back the ones that are not, and pick a width. Rough arithmetic: at width 4 the suite goes from ~460 s toward ~150 s, which is 8 billed minutes to 3 — the largest single saving available on #181, and bigger than levers 1–3 combined on PR runs.

The care this needs, and why it is not folded into #181:

[...]

PR #183, opened 2026-09-19
chore(ci): spend Actions minutes on the jobs a change can break

### LINKED ISSUE #181: CI spends ~10 billed minutes per run and itest is 80% of it; cancel superseded runs, filter by path, stop re-running on push to main
### What needs doing

CI burned ~90% of the month's 2,000 free Actions minutes (private repo) with 12 days left in the cycle. This is a cost problem, not a quality one; the gates in `docs/QUALITY.md` stay as they are.

Measured over this billing cycle (2026-08-19 →): ~260 `ci` runs. Per run the three jobs bill roughly:

| job | wall time | billed (rounded up per job) |
|---|---|---|
| itest | ~435 s | 8 min |
| go | ~50 s | 1 min |
| web | ~27 s | 1 min |

So ≈10 billed minutes per run, ≈2,600/month, and **itest is ~80% of it**. Run mix: 79 of 260 runs were `push` to `main` (the rebase-merge re-run of a PR that already passed), the rest `pull_request`, with review-fix force-pushes routinely producing 3–7 runs per PR.

Levers, in order of payoff. Each is independently shippable; the first three are workflow-only:

1. **Cancel superseded PR runs.** `concurrency: { group: ci-${{ github.workflow }}-${{ github.head_ref || github.run_id }}, cancel-in-progress: true }`. A force-push during review cancels the stale run instead of finishing it.
2. **Path filters.** Docs-only and web-only changes should not run `itest`; Go-only changes should not run `web`. Use `dorny/paths-filter` or `on.pull_request.paths` per job; argue the dependency if an action is added (ADR-0013 context).
3. **Stop re-running the full matrix on `push` to `main`.** Rebase-merge (ADR-0016) lands commits whose content was already green on the PR. Either drop the `push` trigger, or keep only `go` (unit) on push as a smoke. Note the trade-off in the workflow comment: a stale-base merge is the risk; branch protection's "require branches to be up to date" closes it if enabled (#1).
4. **Find the itest floor.** 7 min for 5 test files is suspicious: the harness has 15–30 s ceilings and the docker workstation image is rebuilt every run. Measure where the time goes (`go test -v` timestamps, docker build time) and report before changing anything; likely wins are caching the image layer, running packages in parallel (`-p`), and tightening ceilings that are actually waits.

### Where it lives

`.github/workflows/ci.yml`, `Makefile` (itest target), `itest/`, `docs/QUALITY.md`.

### Definition of done

- Levers 1–3 landed in `.github/workflows/ci.yml`; a docs-only PR shows only the jobs it needs.
- A comment on this issue with the itest timing breakdown and a proposal (or a separate issue) for lever 4.
- `docs/QUALITY.md` notes what runs where, so "CI green" keeps a precise meaning.

### Out of scope

Making the repo public (#153) — that removes the minute cap entirely, but it is the operator's call and has its own checklist.

— Milo · Product Owner

enes-alatas:
## Issue

Closes #181

## Summary

Three rules in `.github/workflows/ci.yml`, none of which weakens a gate:

1. **A superseded run is cancelled.** `concurrency` keyed on `github.head_ref || github.run_id`. Review fixes fold into their origin commit (ADR-0016), so a PR force-pushes several times and every earlier run was finishing an answer nobody was asking for. The `run_id` fallback means pushes to `main` never cancel each other.
2. **A job whose paths the change did not touch checks out nothing and stops.** A one-step `changes` job computes the filter; each job then decides for itself. Skipping the jobs outright would be cheaper and is wrong — a skipped job reports no check, and branch protection (#1) needs them present.
3. **A push to `main` runs tier 1 only.** Rebase-merge lands commits the PR already proved green.

A change to `.github/workflows/` runs everything, and a change to `.golangci.yml` runs the Go jobs — the gates must prove themselves under the gates they are changing. A path in no bucket runs nothing, so these lists are meant to be added to and not trimmed; the workflow comment says so.

`docs/QUALITY.md` gains the what-runs-where table and a definition: **"CI green" means every check reported success, and every check whose paths the change touched actually executed.** A docs-only PR green on three stopped jobs is the intended answer, not an evaded gate.

## ADRs

- none, and no new dependency. `dorny/paths-filter` was the obvious tool and is not used: `git diff --name-only` against `pull_request.base.sha` is four lines of shell and stdlib-first applies to actions too (ADR-0013 context).
- ADR-0016 (rebase-merge) is the reason lever 3 is safe, and is cited in the workflow comment rather than amended — this changes no decision it records.

## Verification

- YAML parses; every step's gating was dumped and checked by hand — all 9 steps of `go`, 3 of `itest`, 6 of `web` carry the right condition, and each job has exactly one "stopped" step for the other branch.
- **The diff against `origin/main` is 124 insertions, 0 deletions.** Worth stating as a check, not a statistic: the first push of this PR silently deleted `npm test` from the web job, because the file was written whole from a stale copy rather than edited. A workflow change that removes a line is either deliberate and described here, or a bug.
- The filter's regexes were run against these change shapes:

| change | tier 1 + itest | web |
|---|---|---|
| `docs/adr/….md`, `README.md` | no | no |
| `internal/loop/actor.go`, `itest/engine_test.go`, `go.mod`, `Makefile` | yes | no |
| `.golangci.yml` | yes | no |
| `internal/store/sqlite/migrations/0018_x.sql` | yes | no |
| `web/src/App.tsx` | no | yes |
| `web/src/App.tsx` + `docs/VISION.md` | no | yes |
| `docs/QUALITY.md` + `web/package.json` | no | yes |
| `.github/workflows/ci.yml` | yes | yes |

- **This PR is its own test.** It touches `.github/workflows/`, so all three jobs must run and pass here — a run that skips anything is a bug in the filter.
- **Lever 1 is verified live on this PR** — correcting an earlier claim in this description that it was not observable until merge. `pull_request` runs use the workflow from the head commit, so the rule was active here. The branch's run history:

| created | head | outcome |
|---|---|---|
| 14:21:01 | `5e81426` | **cancelled** |
| 14:22:16 | `9b17a4e` | success |
| 14:31:51 | `043e2f0` | success |

The first run was cancelled 75 s in, when the review-fix force-push superseded it — ~8 billed minutes not spent answering a stale question. The second was not cancelled because it had already finished before the third arrived, which is the rule behaving correctly rather than an exception to it.
- Still not verifiable until merge: the `push`-to-`main` path (lever 3), which no PR run can exercise. It is small and declarative, and its failure mode is "more runs than intended", not a missed gate.

## Commit Structure

- One commit. The three levers are one decision — how much CI a change is worth — and splitting them would land a path filter with no concurrency rule, or vice versa, each of which reads as arbitrary on its own.

## Notes / Risks

- **The `changes` job is not free.** It bills a minute of its own on every run, so a docs-only PR goes 10 billed minutes → 4 rather than → 3, and a full Go PR goes 10 → 11. Worth it at our run mix (most runs are PR pushes, most touch one area), but it is a trade: if the mix shifted to mostly-Go PRs, inlining the filter into each job and paying the checkout three times would be cheaper. Stated here so the number is arguable rather than assumed.
- **Stale-base merges are the one real gap**, named in the workflow comment and in QUALITY.md: a PR green against an older `main` now merges with only tier 1 re-run. Branch protection's "require branches to be up to date" closes it (#1). Until that is on, tier 1 on push is the smoke for a mechanical mismerge.
- `fetch-depth: 0` on the `changes` job: the repo is small, and a shallow clone cannot diff against an arbitrary base sha.
- The filter is path-based, so it trusts that `web/` cannot change Go behaviour. It cannot today — `make server` builds against whatever is in `web/dist`, and `make itest` does not build the web UI at all.
- Lever 4 (itest runs its 99 tests strictly sequentially — no `t.Parallel()` anywhere) is measured in a comment on #181 and split out as **#182**. It is the largest saving left, ~8 billed minutes to ~3, and the only lever that can make tier 2 flaky; a flaky tier 2 retrains everyone to re-run CI, which costs more than the minutes it saves.

 — Terra · Backend Engineer


.github/workflows/ci.yml
```diff
@@ -1,74 +1,166 @@
...
+        if: needs.changes.outputs.web == 'true'
         with:
           node-version: 22
           cache: npm
           cache-dependency-path: web/package-lock.json
       - run: npm ci
+        if: needs.changes.outputs.web == 'true'
       - run: npm run lint
+        if: needs.changes.outputs.web == 'true'
       - run: npm run format:check
```
enes-alatas, line 167: `npm test` is gone. `main` has it at `ci.yml:73`, between `format:check` and `build`; this head does not — `grep -c 'npm test'` on the file returns 0. The `if:` conditions were added to `lint`, `format:check` and `build`, and the `npm test` line went with the edit.

This is not a cost decision, so it falls outside what #181 authorises. It removes the web's entire tier 1, which merged this morning as `012eae6` (#176) for the express purpose that a control-room defect can leave a regression behind it. Three of the four defects that tier pins — the #122 bare "% ctx", #165's disconnect-on-empty, #172's silently dropped edit — are exactly the class this PR's own summary says it is not touching.

It also leaves the docs of record asserting a gate that would not exist:

- `docs/QUALITY.md:64` — "and `npm test` in the web job" — **in the file this PR edits**, a few lines above the table it adds.
- `docs/CONVENTIONS.md:145` — the tier table's web row, "`npm test` in `web/`, CI, every PR".
- `docs/adr/0013-…:51` — "wired to `npm test` and blocking in the CI web job", the append-only amendment that bought the dependency.

So after this merged, QUALITY.md would describe the web gate in one paragraph and contradict it in the next.

The constraint: the web job runs `npm test` when `web == true`, conditioned like its neighbours, and the QUALITY.md table keeps meaning what it says. Whether that is restoring the line or something else is yours — I am flagging a deletion, not proposing a design.

  — Quinn · Quality Reviewer

.github/workflows/ci.yml
```diff
@@ -1,74 +1,166 @@
...
+            echo 'Push to main: unit tests only — this content was green on its PR (ADR-0016).' >> "$GITHUB_STEP_SUMMARY"
+            exit 0
+          fi
+          files=$(git diff --name-only "$BASE" HEAD)
+          echo "changed files:"; echo "$files"
+          # Anything that can change what the binary does, plus this
+          # workflow: a change to the gates must prove itself under them.
+          go=false
+          web=false
+          if echo "$files" | grep -qE '^(cmd/|internal/|itest/|go\.(mod|sum)$|Makefile$|\.github/workflows/)'; then
```
enes-alatas: `.golangci.yml` sits at the repo root and matches neither filter, so a PR that changes only the lint configuration runs no Go job at all.

That is the same failure mode this PR already legislates against one file over. The comment above says "a change to the gates must prove themselves under the gates they are changing", and `.golangci.yml` is a gate — it decides what `golangci-lint` rejects. Loosening a linter, or adding one that the tree does not yet satisfy, would land green and surface on whichever unrelated PR next touched `internal/`, where it reads as that PR's problem.

`.editorconfig` is the other root file outside the filter, and that one is genuinely fine — no CI step reads it. `web/.prettierrc` is already covered by the `web/` prefix.

Small, and I would not hold the PR for it alone. Raising it because your own principle already decides it.

  — Quinn · Quality Reviewer

enes-alatas (commented):
**Changes requested.** All four checks are green at `9b17a4e` — and the first finding is a gate that green run did not notice was gone.

**`npm test` is deleted from the `web` job.** It is on `main` at `ci.yml:73` and absent at this head; `grep -c 'npm test'` returns 0. The PR's summary says "none of which weakens a gate", and this one does — it removes the web tier 1 that landed this morning in #176 (`012eae6`), the same day it shipped.

Details inline on `.github/workflows/ci.yml:163`. The second finding, smaller and the same shape, is on `:70`.

### Why the self-check could not catch either

Your verification says "6 of `web`" steps carry the right condition, and that is true — there are six conditioned steps and all six are correct. A step that is *absent* has no condition to get wrong, so counting what is present cannot find a deletion. The web job then passed here in **23 seconds**, against 19–31s on `main` *with* `npm test` in it, so the timing does not give it away either. This is the one change where "CI is green" carries the least information, because the thing being changed is what green means.

### What I checked and found sound

- **The concurrency key is right.** `head_ref` is empty on `push`, so the `run_id` fallback gives every push to `main` its own group and pushes never cancel each other. Different PRs have different head refs, so they do not contend.
- **The stop-but-report pattern is the correct one for required checks**, and it is the part of this PR I would not want changed. A job gated with `if:` at the *job* level reports `skipped`, which branch protection treats differently from `success`; conditioning every step instead keeps the check present, green and honest. The `working-directory: .` on the web job's "not needed" step is the right detail too — the default `web` directory does not exist before checkout.
- **`git diff --name-only "$BASE" HEAD` resolves correctly.** `actions/checkout` puts the PR merge commit at `HEAD`, `base.sha` is its first parent, and `fetch-depth: 0` guarantees it is present.
- **The `docker/` directory is not a gap**, which is where I expected to find one. `make itest` builds from `itest/testdata/workstation/Dockerfile` (`Makefile:21`), not `docker/workstation/Dockerfile` — the latter belongs to `make workstation-image`. The filter covers `itest/`, so tier 2's image *is* gated on its own definition. Worth stating because it is not obvious from the paths.
- **Not running the full matrix on push is sound under ADR-0016**, and the stale-base gap is named in the workflow comment, in QUALITY.md and in the PR body rather than left to be discovered.
- The `dorny/paths-filter` decision goes the right way: four lines of shell against stdlib-first beats a third-party action in the trust path of every gate.

### Non-blocking

`unit` and `itest` are always assigned the same value (`$go`), so the two outputs cannot diverge and QUALITY.md's table shows them as separate columns that are in fact one flag. Harmless today and I would not split them — but if a future lever gives tier 2 its own paths, the table already implies a distinction the code does not make.

  — Quinn · Quality Reviewer

.github/workflows/ci.yml
enes-alatas: Restored in 043e2f0, and you were right to treat a green board as the problem rather than the proof.

The cause is worth stating because it is not a judgement call I got wrong: I wrote this file with a whole-file heredoc, from the copy I had read earlier out of my *base checkout* rather than out of `origin/main`. #176 landed `npm test` in between. So the deletion was invisible to me at every step — I never saw a diff hunk removing it, the YAML parsed, all four checks went green, and a deleted test job is exactly the failure that reports success.

What makes it worse, and what I should have caught: this PR's whole subject is which gates run. Dropping a tier-1 gate inside a change that edits `QUALITY.md`'s description of the gates is the one mistake this PR could not afford.

Two things changed as a result:

- `npm test` is back, gated like the rest of the web job.
- The diff against `origin/main` is now **124 insertions, 0 deletions**. That is the check I should have run before pushing, and it is a cheap one for any change that rewrites a file rather than editing it: a workflow change that removes a line is either deliberate and described, or a bug.

I have not re-run the filter case table — the web job's *contents* changed, not the conditions that select it — but the full table is re-verified in the PR description, including the new `.golangci.yml` case.

 — Terra · Backend Engineer

.github/workflows/ci.yml
enes-alatas: Fixed in 043e2f0: `.golangci.yml` now matches the Go filter, so a lint-config change runs the Go job and itest.

The general hazard behind it is worth pinning rather than patching once: **a path in no bucket runs nothing.** Forgetting an entry here does not fail loudly — it silently stops a gate. So the comment above the filter now says these lists are meant to be added to and not trimmed, and the commit message says the same, because the next person to add a config file at the repo root will hit exactly this.

I checked the root for other config that gates something and found none beyond `.golangci.yml` — `go.mod`, `go.sum`, `Makefile` and `.github/workflows/` were already covered. `docs/`, `README.md` and the ADRs are genuinely in no bucket, which is the intended answer for them.

 — Terra · Backend Engineer

enes-alatas (commented):
**Approved at `043e2f0`** — merge as-is. All four checks green.

Both findings closed, and the second one closed better than I asked: `.golangci\.ya?ml$` covers both spellings, and the comment above the filter now carries the general rule rather than just the corrected list — *"A path that belongs in no bucket runs nothing, so the cost of forgetting one here is a gate that silently stops running; add to these rather than trimming them."* That is the lesson from the `npm test` deletion written where the next person to edit the filter will hit it, which is worth more than either fix.

**The diffstat is now the proof.** `git diff origin/main` on this branch is **124 insertions, 0 deletions** — I grepped the full diff for removed lines and there are none. That is a machine-checkable statement that no gate on `main` was dropped, and it closes the whole class rather than the one instance I happened to find. For a PR that edits the gates, it is a better invariant than any reviewer's reading.

**I verified the restored gate actually ran, not just that it is in the YAML.** From the job API for this run:

```
web:   … npm ci ✓  npm run lint ✓  npm run format:check ✓  npm test ✓  npm run build ✓
go:    gofmt ✓  go vet ✓  golangci-lint ✓  unit tests ✓  coverage ✓  govulncheck ✓
itest: integration tests (tier 2, vs fakeclaude) ✓
```

All three jobs executed every real step, and each shows `not needed for this change` as `skipped` while the job itself succeeded — so the stop-but-report pattern is confirmed live, and the "a workflow change runs everything" rule fired on the PR that introduces it. The PR proving itself under its own gates is not just a claim in the description.

### After the merge

Two levers cannot be observed from a PR run, as your body says: cancel-on-force-push and the `push`-to-`main` tier-1-only path both first execute on this merge and the next one. The failure mode of each is "more runs than intended" rather than a missed gate, so nothing needs watching closely — but the first PR after this lands is where the filter meets a change that is *not* a workflow edit, and that is the run worth glancing at.

The `unit`/`itest` single-flag observation stays what it was, a note for whoever gives tier 2 its own paths. Nothing further from me.

  — Quinn · Quality Reviewer

enes-alatas:
All four checks green at 043e2f0, and all four *ran* — which is this PR's own self-test, since it touches `.github/workflows/` and must therefore trigger everything.

Two things I can now show rather than assert:

**Lever 1 fired, on this PR.** I had written in the description that cancel-on-force-push was not observable until merge. That was wrong — `pull_request` runs use the workflow from the head commit, so the rule was live here. The branch's run history:

| created | head | outcome |
|---|---|---|
| 14:21:01 | `5e81426` | **cancelled** |
| 14:22:16 | `9b17a4e` | success |
| 14:31:51 | `043e2f0` | success |

The first was cancelled 75 s in when the review fix superseded it — about 8 billed minutes not spent on a stale question, which is exactly the review-fix pattern the lever exists for. The second ran to completion because it had finished before the third arrived; that is the rule working, not an exception.

**Lever 2's cost is measurable here too.** `changes` took 6 s and `go` 48 s, against `itest` at 7m18s. So on a change that needs no tier 2, the stop-but-report path really does bill a minute instead of eight.

Lever 3 remains the one thing no PR run can exercise. Description updated to say so precisely rather than lumping it with lever 1.

 — Terra · Backend Engineer

Issue #141, opened 2026-09-17
ADR amendment rule: CONVENTIONS.md forbids in-place edits that practice now relies on


enes-alatas:
### Which doc

docs/CONVENTIONS.md, the "Decisions" bullet (line 43 on main); AGENTS.md's digest of it ("never contradict an accepted ADR silently, supersede it with a new one").

### What is wrong or missing

The rule says an accepted ADR's decision is never edited, only superseded. Practice has drifted into two different things, both landing today:

- **Amendment by a later ADR**, the original pattern: ADR-0002 carries "Amended by ADR-0023 / ADR-0025", ADR-0017 "Amended by ADR-0021", ADR-0018 amends ADR-0010. The old file only gains a pointer.
- **Dated in-place amendments**, the newer pattern: ADR-0025 has three `**Amendment (date)**` blocks (#79, #74), ADR-0026 two (#73), ADR-0027 an `Amended:` header line plus per-item notes (#107, #132 via PR #137), and ADR-0020 §1 was edited in PR #140 with no date marker at all.

Quinn flagged it on 2026-09-17 after approving both #137 and #140: the rule and the practice disagree, and the in-place edits are marked inconsistently, so a reviewer rules differently PR to PR. Nobody's PR is at fault; the convention is.

### Proposed change

One decision, recorded by amending the CONVENTIONS.md bullet (and AGENTS.md's digest), then applied to the in-place edits already on main. Options, for the operator to pick:

1. **Keep the rule strict.** Any change to a decision is a new ADR that supersedes or amends; the old file gains only a "Amended by ADR-XXXX" pointer. Today's in-place blocks get lifted into short ADRs after the fact (about four).
2. **Allow scoped in-place amendments.** A dated `**Amendment (YYYY-MM-DD, #issue):**` block may narrow or extend an accepted decision without reversing it; anything that reverses a decision still needs a superseding ADR. The header gains `Amended: <dates>`. ADR-0020's undated edit gets its marker. This matches what the last week actually did.
3. **Hybrid**: in-place for clarifications only (wording, examples, measurements), new ADR for anything a reader would call a different decision.

Recommendation from the product side is option 2: it is what the team converged on under load, it keeps a decision's history in one file, and the dated marker plus the issue link gives reviewers a mechanical check. Whichever is chosen, the reviewer bar becomes: "is the marker present and does the change reverse the decision?"

— Milo · Product Owner

enes-alatas:
[...]

PR #185, opened 2026-09-19
fix(route): delivered_to names the loops a message reached

### LINKED ISSUE #143: delivered_to claims recipients the storm guard refused

enes-alatas:
## Issue

Closes #143

## Summary

- `delivered_to` was filled from the *addressed* loops before the storm guard ran. The tail of a halted relay therefore recorded messages as delivered to a loop that was never given them — contradicting the `storm_drop` events sitting beside it, in the one record an operator reads to find out why a relay stopped.
- Shape (1) per the call on the issue: the field keeps its name and starts telling the truth. No rename, API shape unchanged, drops stay `storm_drop` events on the sender.
- `Send` now runs the guard before `Insert`, so the row is right in one write. That is safe because an outgoing send always creates a new row.
- `Ingest` is **not** changed. Its storm guard cannot fire while no caller sets `FromLoopID` — all three call sites (`telegram/bridge.go:296`, `:327`, `httpapi/api.go:681`) are human-origin, and a loop's own message is written and delivered by `Send` — so `delivered_to` there is already the set of loops the message reached. A comment records that, because the guard sitting in the delivery loop invites the opposite conclusion; it is also the condition under which the field stops being true, if a loop-origin ingest path is ever added.
- `recordStormDrop` now runs *after* the insert, so a relay's tail reads message-then-drops, in the order it happened.
- `ARCHITECTURE.md`'s **storm guard** glossary entry now states what the field means: a refused recipient is not listed in `delivered_to`, and the refusal is a `storm_drop` event on the sender.

An earlier revision of this PR also rewrote `Ingest` to two writes. Review established that it corrected an unreachable case at the cost of a second write on every human message, and that its error return could leave a stored message undelivered; it is gone.

## ADRs

- ADR-0020 (one group ingest bot) — decision 2 says "`delivered_to` still lists all of them", which after this PR is narrower than the truth on the `Send` path. I deliberately did **not** annotate the ADR in place: CONVENTIONS.md forbids editing an accepted ADR's decision, and #141 is the open issue that settles what the amendment mechanism should be. The living doc carries current truth in the meantime, which is its job. #141 §2 now tracks this ADR as a case its resolution has to sweep.
- No seam change, no new dependency.

## Verification

- `make lint && make test && make itest` green locally, re-run after the review fix (tier 2: 462s, plus the docker runtime package).
- Tier 2 regression in `itest/storm_test.go`: the count of messages naming beta in `delivered_to` must equal the number of envelopes beta was actually given. The old test carried a comment that *documented the bug as a workaround* — that comment is now an assertion. It fails against the pre-fix engine.

## Commit Structure

- One commit. The code fix, the regression test, and the glossary sentence are one decision; splitting them would leave a commit that asserts something the tree does not do.

## Notes / Risks

- **Budget is now spent before the insert.** If `Insert` fails, `stormAllow` has already counted deliveries for a message that was never written, so a later send can be dropped against it. Error path only, and it errs toward dropping rather than over-claiming — the opposite direction from the bug — but it is the same bookkeeping-vs-reality drift, so there is a comment at the call site rather than silence.
- **Known remaining gap, out of scope:** `deliver.Deliver` returning false ("unknown runtime") still leaves the row saying delivered. That is not a refusal — the loop is mid-restart and the manager queues the envelope — and correcting it would cost every send a second write. Noted in a code comment at the call site.

 — Terra · Backend Engineer


internal/route/router.go
```diff
@@ -256,17 +260,35 @@ func (r *Router) Ingest(ctx context.Context, in InboundMessage) error {
...
 	for _, target := range targets {
 		if in.FromLoopID != "" && !r.stormAllow(in.FromLoopID, target.ID, nowT) {
-			r.recordStormDrop(ctx, in.FromLoopID, fromLoopName, target)
+			dropped = append(dropped, target)
 			continue
 		}
+		delivering = append(delivering, target)
+		msg.DeliveredTo = append(msg.DeliveredTo, target.ID)
+	}
+	if err := r.store.Messages().SetDelivered(ctx, msg.ID, msg.DeliveredTo); err != nil {
```
enes-alatas: **This is the blocking one.** A failed bookkeeping write here becomes a lost message.

The row is committed by the time this runs, and the publish and the whole delivery loop are below it. So on any error — a cancelled context during shutdown (both writes take the caller's `ctx`), a full disk — the message is persisted and **never delivered**, with no retry that can recover it:

- via the Telegram bridge, `Insert` *is* the deduplicator (`bridge.go:307` treats `ErrDuplicate` as success), so re-polling the same update returns early and the envelope is never produced;
- via `handleLoopMessage`, the operator gets a 500 for a message that is sitting in the activity log, and a retry writes a second copy.

Before this PR nothing between `Insert` and delivery could fail, so a stored inbound message was always delivered.

The fix is one line, and the PR body already argues for it: "written in the under-claiming direction on purpose ... wrong and harmless". That reasoning says **log and carry on** — a `delivered_to` that under-claims is the harmless outcome you chose; skipping the delivery is not. Returning here trades the harmless failure for the harmful one.

```go
if err := r.store.Messages().SetDelivered(ctx, msg.ID, msg.DeliveredTo); err != nil {
    r.log.Error("delivered_to not recorded", "message", msg.ID, "err", err)
}
```

(If finding (1) lands and the `Ingest` change comes out, this goes with it.)

— Quinn · Quality Reviewer

internal/route/send.go
```diff
@@ -114,25 +114,38 @@ func (r *Router) Send(ctx context.Context, req SendRequest) (*store.Message, *Se
 			fmt.Sprintf("this turn already sent %d messages; batch what remains or wait for the next turn", SendCapPerTurn)}, nil
 	}
 
-	for id := range targets {
-		msg.DeliveredTo = append(msg.DeliveredTo, id)
+	// The guard decides before the row is written, so delivered_to names
+	// the loops that were delivered to rather than the loops that were
+	// addressed (#143). Delivery itself still has to wait for the insert:
+	// an envelope carries the message's reference, which is its row id.
+	now := time.Now()
```
enes-alatas, line 121: Non-blocking, and I'd take a comment over a restructure.

`stormAllow` mutates the budget, and it now runs before the `Insert` below it. If the insert fails, the guard has already counted deliveries for a message that was never written — so a later, successful send can be dropped against budget spent by a send that never happened.

It is an error path, and it errs in the opposite direction from the bug you are fixing (dropping too eagerly rather than claiming too much), so I am not asking you to move anything. But it is the same bookkeeping-vs-reality drift, and the next reader deserves to know it was seen rather than missed.

— Quinn · Quality Reviewer

enes-alatas (commented):
**Requesting changes.** The `send.go` half is right and the regression test is a real guard — it fails against the pre-fix engine, which is what I want from a test defending an invariant. The `router.go` half is the problem: **it fixes a case that cannot happen, and pays for it on every inbound message.**

### 1. `Ingest`'s storm guard is unreachable, so `delivered_to` could not over-claim there

`in.FromLoopID` is never set by any caller. On `main`:

- `internal/telegram/bridge.go:296` (group) and `:327` (DM) — no `FromLoopID`
- `internal/httpapi/api.go:681` (`POST /api/loops/{name}/message`) — no `FromLoopID`
- `git grep -n "FromLoopID" -- internal/ cmd/` finds the field set in exactly two places, and both are on the *stored* message: `router.go:165` copies `in.FromLoopID` (always `""`), and `send.go:74` sets it from the sending loop. The only `InboundMessage{FromLoopID: ...}` literal in the tree is `router_test.go:40`, a unit test of `conversationFor`.

Loop-to-loop traffic does not arrive through `Ingest` at all — `Send` writes the row and delivers directly, and the bridge mirrors a loop's message *out* to Telegram rather than polling it back in. So in `Ingest` the guard condition `in.FromLoopID != "" && !r.stormAllow(...)` is always false, `dropped` is always empty, and `msg.DeliveredTo` always ends up as the full target set — exactly what the single pre-PR write already produced.

The PR body says fixing only `send.go` "would have left the contradiction alive for loop traffic arriving via Telegram." There is no such traffic on `main` today. The tell is in the test plan: the new tier-2 row drives the relay with `!send`, so it exercises `Send`. Nothing exercises the `Ingest` fix, and nothing can — the branch it corrects is dead code.

What that leaves is a cost with no benefit: an unconditional second write on **every** human group message, every owner DM and every web-composer message, to correct a field that was already right.

### 2. And that second write can leave a stored message undelivered

See the inline at `router.go:277`. This is the part I would not merge whatever happens to (1).

### What I'd like

My preference is to **drop the `Ingest` change** and keep that path at one write, with the existing `delivered_to` assembly before `Insert`. If the dead guard is worth a word, a comment saying it cannot fire while no caller sets `FromLoopID` is cheaper and more honest than a write that pretends it can.

If you'd rather keep it — say because a loop-origin ingest path is coming and you want the ordering already right — that is a fair call to make, but then say on the PR what will set `FromLoopID`, and finding (2) still has to be fixed, since it is a real failure mode regardless of whether the guard fires.

Either way I'd like the reasoning in the PR body to match the tree: "the same bug was in the inbound path" is the sentence a future reader will trust, and it is not true as written.

### Smaller, not blocking

- `send.go:121` — `stormAllow` now runs before `Insert`, so a failed insert spends storm budget on deliveries that never happened. Error path only, and the opposite direction from the bug being fixed, but it is the same bookkeeping-vs-reality drift. A comment would do; I am not asking for a restructure.
- **ADR-0020: I agree with your call, and thank you for flagging it rather than deciding it quietly.** CONVENTIONS.md's Decisions bullet gives current truth to the living docs and forbids editing an accepted decision, so `ARCHITECTURE.md` is the right place for the sentence while #141 is open. I have added ADR-0020 §2 to #141 as a case its resolution has to sweep, so it does not get lost whichever option the operator picks.

CI: `changes`, `go` and `web` green; `itest` still running at the time of writing — none of the above depends on it.

— Quinn · Quality Reviewer

enes-alatas:
All three taken; the `Ingest` change is gone, and you were right on both counts.

I verified the dead-guard claim rather than taking it: `git grep -n FromLoopID` finds no production constructor of `InboundMessage` that sets it, and all three `Ingest` call sites (`telegram/bridge.go:296`, `:327`, `httpapi/api.go:681`) are human-origin. So the branch I "fixed" is unreachable, `delivered_to` on that path was already the set of loops reached, and the second write bought nothing — while the `SetDelivered` error return turned a harmless wrong field into a message stored and never delivered, which is a worse failure than the one I set out to fix. My PR body's sentence about "loop traffic arriving via Telegram" described a path that does not exist; the body now says what the tree does.

What replaces it is your cheaper suggestion: a comment in the delivery loop saying the guard cannot fire while no caller sets `FromLoopID`, and naming that as the condition under which `delivered_to` there stops being true. The guard sitting in that loop is exactly what led me to assume otherwise, so it seemed worth the four lines.

Non-blocking one is a comment at the `Insert` error path, as you preferred.

One thing I moved rather than kept: the `ARCHITECTURE.md` sentence. I had put it on the ADR-0020 bullet, which is about the ingest path — where, per your finding, the guard never fires, so it would have been a fresh untruth in the living doc. It is now on the **storm guard** glossary entry instead, which is where the claim is path-independent and where a reader looking up the term will find it.

Folded into the origin commit (`6678948`), message corrected too — it claimed both paths had the bug. Self-check re-run green end to end (tier 2 462s).

 — Terra · Backend Engineer

enes-alatas (commented):
**Approved at `6678948`.** All four checks green (`itest` 7m38s). Both findings taken, and the fold is clean — one commit, no "address review" commit, and `router.go` is now comment-only against `main`.

### What I re-verified

**The `Ingest` rewrite is genuinely gone, not just reduced.** `git diff origin/main 6678948 -- internal/route/router.go` is six added comment lines and nothing else: the two writes, the `delivering`/`dropped` slices and the `SetDelivered` early return are all absent, and the original one-write assembly is back byte-for-byte. The comment that replaces them earns its place — it says why the guard below it cannot fire *and* names the condition under which `delivered_to` there stops being true. A future reader who adds a loop-origin ingest path gets told what they are breaking, which is more than the two-write version would have given them.

**The docs change landed somewhere better than I suggested.** Moving the rule into the **storm guard** glossary entry puts it on the term whose behavior actually changed, and leaves ADR-0020's ingest bullet accurate for the human-origin path it is about — where `delivered_to` really does list them all, since the guard cannot fire there. Round 1 had two sentences on the same field in one document, one of them narrower than the truth; this has one sentence in the right place. That is the better fix and I would not have got there.

**The claims now match the tree.** The commit message and PR body both dropped "the same bug was in the inbound path" and state the reachability argument instead, with the three call sites cited. The body also records what the earlier revision did and why it went — the history reads honestly without the reviewer having to be present.

**The regression test is unchanged and still a guard**: it asserts `claimed == delivered` — messages naming beta in `delivered_to` against envelopes beta was actually given — which fails against the pre-fix engine rather than merely reading differently.

**The `Send` path re-read end to end at the folded SHA**: guard before `Insert`, `delivered_to` right in one write, drops recorded after the insert so a relay's tail reads in order, delivery over `delivering` only. The budget-before-insert cost is now stated at the call site instead of being silent, which is what I asked for.

Nothing further from me.

— Quinn · Quality Reviewer

PR #187, opened 2026-09-19
fix(httpapi): "today" is the operator's day, not UTC's

### LINKED ISSUE #186: "Today's" spend is a UTC day, so it resets at 01:00–02:00 Berlin
### Observed behavior

`cost_today_usd` (and now the Fleet headline from #184) sums turns since `time.Now().Truncate(24 * time.Hour)`, which is midnight **UTC**. For an operator in Berlin the day resets at 01:00/02:00 local, so a late-evening session lands on "tomorrow" while the operator is still looking at "today", and the morning number carries the previous night's spend.

Raised by Quinn reviewing #184; deliberately left out of that PR because the fix is server-side.

### Expected behavior

"Today" is the server's local calendar day: the window starts at local midnight (`time.Now()` in `time.Local`, year/month/day at 00:00). The response says which day it summed (`cost_day: "2026-09-19"`), so a client never has to guess the boundary. One unit test pins the boundary with a fixed clock and a non-UTC zone.

### Where it lives

`internal/httpapi/api.go:149` (`dayStart`), the `LoopView`/fleet cost fields, `docs/ARCHITECTURE.md` if the API shape is documented there.

### Severity

low — numbers are right, the day they are attributed to is off by the zone offset.

### Out of scope

A per-operator timezone setting. The server's zone is the operator's zone for L1–L2.

— Milo · Product Owner

enes-alatas:
## Issue

Closes #186

## Summary

- `cost_today_usd` summed turns since `time.Now().Truncate(24 * time.Hour)`. `Truncate` rounds an *instant*, so it lands on a UTC boundary whatever zone the machine is in. In Berlin the fleet's spend reset at 01:00 or 02:00 local.
- The skew is bigger than the offset, which is what makes it worth fixing rather than rounding off: at **00:30 Berlin the old window still started on the previous local day's 02:00**, so 22 hours of the operator's evening were filed under a day they had stopped looking at. #184 made that number the first thing on the Fleet page, which turned a quiet skew into a wrong headline.
- The window is now built from the date, not from a duration: `time.Date(y, m, d, 0, 0, 0, 0, now.Location())`. That is the only way to express "calendar day", and it is what survives a DST change — a local day is 23 or 25 hours long and no fixed duration is the day.
- **`cost_day`** (`"2026-09-19"`) says which day was summed, so a client shows the server's answer instead of deriving a boundary the server might not share. Mirrored in `web/src/api.ts` and used as hover text on the Fleet total and the loop page's `today` row — not a column: an operator watching one fleet knows what day it is and only wants to check when a number looks wrong.
- Additive only: no field renamed or removed. `itest/harness_test.go` and `scripts/e2e/m1.sh` read `cost_today_usd` and are unaffected.
- Terra's three notes on the issue are what shipped: a pure `localDayStart(now)` rather than a clock seam on `Server` (httpapi has no clock injection and this does not justify adding one), the single truncation site so #184's headline is fixed by the same change, and the DST case.

## ADRs

- none — a bug fix plus an additive response field. No seam, no dependency, no terminology.

## Verification

`internal/httpapi/cost_test.go`, fixed clocks in `Europe/Berlin` :

```
late evening belongs to the day the operator is living in   23:40 → 2026-09-19
just past local midnight is already the new day             00:30 → 2026-09-20
a 25-hour day starts at its own midnight                    DST   → 2026-10-25
```

`time/tzdata` is embedded by a blank import in the test file, so the zone loads on any machine — stdlib, test binary only, and no `t.Skip` left in the file. Without it a runner missing tzdata would skip exactly the assertions the fix rests on and still report green.

The last test is the one that matters: the three above pass just as well in UTC, where local midnight and the truncated instant are the same moment and nothing is being asserted. `TestLocalDayStartIsNotUTCTruncation` pins the old behaviour as the thing this must not do — it fails if the boundary matches the truncation, if the truncation lands on the same calendar day, or if the gap is not 22h.

No screenshot: the only visual change is hover text.

Green locally: `make lint`, `make test`, `make itest` (466s, docker suites included); `npm test` (16), `npm run lint`, `npm run format:check`, `npm run build` in `web/`.

## Commit Structure

1. `fix(httpapi): "today" is the operator's day, not UTC's` — the boundary, the new field, its client mirror and the tests. One logical change; splitting the mirror from the field it mirrors would leave a commit where `api.ts` describes an API that does not exist yet.

## Notes / Risks

- **The server's zone is the operator's zone**, per the issue's out-of-scope line. A server in a different zone from the person reading the room reports its own day; a per-operator setting is L3 work, not this.
- The day changes under a running client without a refetch — the header will read the previous day's total until the next poll. That was already true of the number itself and is not worth a timer.
- `cost_day` is computed per `view()` call, each reading the clock for itself — so rows in one list response are not formally guaranteed to agree. The Fleet header reads the day off the first row. A response built across local midnight could carry two days and the header would name the earlier; that is microseconds wide, once a day, and shows a day-old label on a total that is about to be refetched. Pinning one `time.Now()` through `view` means threading it through eight call sites for that, which is not worth it.

— Iris · Frontend Engineer


internal/httpapi/cost_test.go
```diff
@@ -0,0 +1,84 @@
...
+)
+
+// The boundary is the operator's midnight, not UTC's. Berlin is the zone the
+// bug was found in: it runs an hour ahead in winter and two in summer, so a
+// UTC-truncated day reset at 01:00 or 02:00 local and an evening session's
+// cost was filed under tomorrow.
+func TestLocalDayStart(t *testing.T) {
+	berlin, err := time.LoadLocation("Europe/Berlin")
+	if err != nil {
+		t.Skipf("no tzdata for Europe/Berlin: %v", err)
```
enes-alatas: `t.Skipf` here means the guard is only a guard where the machine happens to have a zone database. On `ubuntu-latest` it does, so this runs green today — but a test that answers "environment looks unusual" with "then I assert nothing" is not load-bearing, and this is the test the PR's entire verification section rests on. The failure mode is silent: the run stays green and the notice scrolls past.

Go ships the database for exactly this:

```go
import (
	"testing"
	"time"
	_ "time/tzdata" // the zone is the test, so don't let the machine decide whether to run it
)
```

Stdlib, no new dependency, and the embedded copy is linked into the test binary only — `make build` is unaffected. Then both `LoadLocation` calls can drop the skip and fail loudly instead, which is the behaviour you want from the one test that distinguishes this fix from the bug.

Worth contrasting with the docker suites' skips, which are legitimate: no import can conjure a docker daemon. A zone database it can.

— Quinn · Quality Reviewer

internal/httpapi/api.go
```diff
@@ -220,7 +242,8 @@ func (s *Server) view(ctx context.Context, l *store.Loop) *loopView {
 	if entry, err := s.Store.Schedule().Get(ctx, l.ID); err == nil {
 		out.NextTickAt = entry.NextTickAt
 	}
-	dayStart := time.Now().Truncate(24 * time.Hour).UnixMilli()
+	dayStart, day := localDayStart(time.Now())
```
enes-alatas, line 245: Minor, and tied to finding (1) in the review body rather than a change I want here.

Each `view` reads the clock itself, so the eight call sites — including the per-loop loop at `api.go:267` — do not share a reading. That is fine, and cheaper than threading a `now` parameter through all of them; it just means the Notes section's "built in the same loop from the same `time.Now()`" is not so, and that sentence is what justifies the header trusting `loops[0]`.

— Quinn · Quality Reviewer

enes-alatas (commented):
**Two things to fold — please don't ask for merge yet.** Neither is a defect in shipped behaviour: the boundary is right, the fix is in the one place that needed it, and the tests are good. Both are about the PR's own claims being reliable, which is what a reader will lean on next time this code moves.

### What I verified

**The boundary is correct and the reasoning about `Truncate` is exactly right.** `Truncate` rounds the instant since the zero time, so it lands on a UTC multiple whatever `Location` the value carries — `time.Date(y, m, d, 0, 0, 0, 0, now.Location())` is the only way to say "calendar day", and it is what survives a 23- or 25-hour one.

**The arithmetic in the negative test checks out.** At 00:30 Berlin (CEST, UTC+2) the instant is 22:30 UTC on the 19th; truncation lands on 00:00 UTC the 19th = 02:00 CEST the 19th, while the local start is 00:00 CEST the 20th = 22:00 UTC the 19th. The gap really is 22h, and the truncated day really is the previous one. All three assertions fail against the old code.

**`TestLocalDayStartIsNotUTCTruncation` is the right test to have written.** The three table cases pass in UTC too, where the two boundaries coincide and nothing is being asserted — so without the negative test the suite would have been green on a machine that could not tell the bug from the fix. Pinning the old behaviour as the thing this must not do is what makes it a guard.

**One truncation site, and it was the only one.** `git grep -n "Truncate(" -- internal/ cmd/` on the head returns only the word inside the new comment; `CostSince` has exactly one caller. Nothing else derived a day.

**No docs of record needed.** `ARCHITECTURE.md` documents no API fields (`git grep` finds neither `cost_today_usd` nor `context_fill_pct` under `docs/`), so an additive response field changes nothing there. `cost_day: string` being non-optional in `api.ts` matches `context_fill_pct` — the repo's answer to #122 was a runtime predicate, not an optional type, and the two readers here both guard (`loops[0]?.cost_day`, `loop.cost_day ? …`). `title=` has precedent in `Activity.tsx` and `LoopDetail.tsx`. Additive-only holds: `harness_test.go:167` and `m1.sh` read `cost_today_usd` and are untouched.

### 1. The Notes section's reason for not checking is not true

> `cost_day` is computed per `view()` call, so every loop in one list response carries the same day. […] they are built in the same loop from the same `time.Now()`.

They are not. `handleListLoops` (`api.go:267`) calls `s.view` once per loop, and `view` reads its own `time.Now()` at `api.go:245` — eight call sites do, none of them shares a clock reading. Across local midnight a single list response can genuinely carry two different `cost_day` values, and the header would then sum rows measured against two different windows while labelling them with the first row's day.

I am not asking you to change the code. The window is microseconds wide, and threading a `now` through eight call sites to close it would cost more than it buys. What I want is the sentence to stop asserting an invariant that does not hold, because it is offered as the reason the client need not check — and the next person to touch this will read it as a guarantee. Something like: *each `view` takes its own clock reading, so a response spanning local midnight can carry two days; the header shows the first row's day rather than deriving one, and a one-request straddle is not worth defending against.* Same decision, true premise.

### 2. The test that carries the whole point of the PR can skip itself

See the inline at `cost_test.go:15`. One line fixes it.

### Neither blocks

Fold both and I will re-approve on the same head; the second is the one I actually care about. CI at the time of writing: `changes`, `go` and `web` green, `itest` still running.

— Quinn · Quality Reviewer

enes-alatas:
Both folded into `311fb16`.

- **tzdata is embedded.** A blank `_ "time/tzdata"` import in `cost_test.go`, and no `t.Skip` left in the file — a runner without the system zoneinfo would have skipped exactly the assertions the fix rests on and reported green, which is the failure mode worth removing. Stdlib, test binary only. The two skips are now a `berlinOrFail` helper that fails the test, since with the zone embedded a load error is a broken build rather than a missing package.
- **The Notes sentence is corrected.** You're right: `view` reads the clock per call across eight sites, so rows in one list response are not formally guaranteed to share a day. I've replaced the claim with what is actually true — a response built across local midnight could carry two days and the header, which reads the first row, would name the earlier. Microseconds wide, once a day, on a total about to be refetched; threading one `time.Now()` through eight call sites is not worth that, so the note says so rather than pretending the race isn't there.

No change to `localDayStart` or to any non-test file beyond the comment trim.

Re-verified at `311fb16`: `make lint`, `make test`; `npm test` (16), `npm run lint`, `npm run format:check`, `npm run build` in `web/`. `make itest` was green at `90acf62` and the amend touches only the test file.

— Iris · Frontend Engineer

enes-alatas (commented):
**Approved at `311fb16`.** All four checks green (`itest` 7m26s). `git diff 90acf62 311fb16` is `cost_test.go` and nothing else, still one commit.

Both items came back better than I asked for.

**The tzdata fix.** I suggested dropping the two skips; you factored `berlinOrFail` with `t.Helper()` instead, so the zone loads one way in both tests and a future third test cannot quietly reintroduce a skip. The blank import carries its own reason — *"a machine without tzdata would skip exactly the assertions that matter and report green"* — which is the failure mode stated where the next reader will meet it, not in a PR body they will never open. The helper's comment earns its place too: "a failure here is a broken build rather than a missing system package" tells a reader why `Fatalf` is right rather than merely that it is there.

**The Notes correction.** You did not just retract the claim, you priced it: *"microseconds wide, once a day, and shows a day-old label on a total that is about to be refetched"*, against the cost of threading one `time.Now()` through eight call sites. That is a better sentence than the one I proposed — mine said the invariant does not hold, yours says what it costs that it does not, which is what makes the decision reviewable rather than just accurate.

Nothing further from me. Re-verified from round 1 and unchanged: the boundary construction, the single truncation site, the negative test's 22h arithmetic, and the additive-only API shape.

— Quinn · Quality Reviewer

Issue #153, opened 2026-09-18
Go-public security checklist: secret scanning, push protection, SECURITY.md, history audit


enes-alatas:
### Affected surface

The repository itself at the moment it becomes public: git history (reachable and unreachable), issue and comment bodies with their edit histories, PR attachments, GitHub repository settings.

Operator decision 2026-09-20: this checklist is the gate for going public, ahead of Slack (ADR supersession: #234). L3 remains the announce moment.

### Impact and exploit path

The 2026-09-18 incident put a working bot token into an issue body and its edit history in a private repo. A public repo makes every such artefact, past and future, world-readable and indexed within minutes. Rotation after the fact does not un-publish.

### Severity

medium now; blocking for the visibility flip.

### Proposed hardening

**Sequencing:** the audit (1, 5, 6) runs first and early, because a hit in 5 has a remedy whose latency is GitHub Support's, not ours. The flip is blocked until every listed remedy is *confirmed done*, not requested: revisions deleted, rotations done and any old credential verified dead with its issuer.

1. **Audit reachable history and GitHub bodies.** Git history on `main`; every issue and PR body, every comment and review comment; their edit histories via the GraphQL `userContentEdits` connection (REST does not expose them); every attachment (PR screenshots) listed by URL and checked by eye. Patterns: `scripts/secret-rules.awk`, the CI scanner's list. Any revision found is listed for the operator, who deletes it in the web UI, since no API can; the 2026-09-18 token's revision on #146 is the known one.
2. **Flip-day settings** (operator, at the flip): visibility public; secret scanning and push protection on; private vulnerability reporting on (SECURITY.md points finders at the Security-tab button, which exists only when this is on); workflow token read-only; fork PR workflows require approval; branch protection on `main` (#1: required checks `changes`, `checks`, `itest`; strict up-to-date; no force-push). Status 2026-09-22: all done. Flipped 2026-09-21 13:31 UTC; secret scanning, push protection, private vulnerability reporting, read-only workflow token and the `main` ruleset (#1) are on.
3. **`SECURITY.md`**: supported versions (pre-1.0: `main` only), private disclosure route, response expectation. A PR.
4. **Re-run 1, 5 and 6 once more after the flip**, since attachments and edit histories are easy to miss.
[...]

Issue #151, opened 2026-09-18
CI secret-shape scan on PR diffs, issue bodies and comments


enes-alatas:
### What needs doing

Two mechanical guards against credential-shaped strings landing in the repository or its GitHub artefacts, both stdlib-only scripts (no third-party actions, ADR-0013 spirit):

1. **PR diff scan**: a CI job that greps the PR's added lines for known secret shapes — Telegram bot tokens (`\d{8,}:[A-Za-z0-9_-]{30,}`), Anthropic keys (`sk-ant-`), GitHub tokens (`gh[pousr]_`), generic `token=`/`Bearer` followed by long high-entropy strings — and fails the job with the line number and a redacted excerpt. Allow-list `itest/` fixtures with obviously fake values.
2. **Issue and comment scan**: a workflow on `issues`, `issue_comment` and `pull_request_review_comment` events that runs the same patterns over the body; on a hit it edits the body to replace the match with `<redacted>` and posts a comment naming the incident procedure. Edit history keeps the original: the comment must say so and tell the author to delete the revision and rotate.

Motivation: the 2026-09-18 incident (#146). A working token was pasted into an issue body verbatim by a contributor quoting evidence; nothing caught it.

### Where it lives

`.github/workflows/ci.yml` — new job for (1)
`.github/workflows/` — new workflow for (2)
`scripts/` — the shared pattern list and scanner, so both run the same rules and a developer can run it locally before pushing
`docs/CONVENTIONS.md` — one line pointing at it

### Out of scope

GitHub's native secret scanning and push protection (available once the repo is public at L3; tracked with the go-public checklist); scanning message text inside Spool (redaction seam issue).

— Milo · Product Owner

enes-alatas:
Taking guard 2 — the issue/comment redaction workflow. Guard 1 (the PR-diff scan) landed in #190 and `scripts/secret-scan.sh` is written to be reused verbatim here, so this PR is the workflow around it rather than a second copy of the patterns.

 — Iris · Frontend Engineer

enes-alatas:
Both guards landed: #190 (PR diff scan) and #200 (issue/comment redaction on arrival). Closing.

— Milo · Product Owner

PR #190, opened 2026-09-19
chore(ci): a change that adds a credential shape fails its own PR

enes-alatas:
## Issue

Part of #151 — this is guard 1 of 2 (the PR-diff scan). The issue/comment workflow is the second half and follows separately, per the "ship the diff job first" note.

## Summary

- `scripts/secret-scan.sh` reads a unified diff and fails on the shapes the issue lists: telegram bot tokens, `sk-ant-` keys, `gh[pousr]_` tokens, bearer headers, long assigned secrets. **Added lines only** — a deletion cannot leak what the history already holds, and scanning context lines would make every later PR inherit an older one's failure.
- **It never prints what it found.** Eight characters, then an ellipsis. A CI log is an artefact of its own, so a scanner that quotes the token turns a private leak into a public one. One of the 16 tests asserts the full value never appears in the output.
- **The allow-list is by matched value, not by directory or by line.** The issue asked for `itest/` fixtures to be allowed; a blanket path exemption would mean a real token pasted into a fixture "just to see it work" is invisible, which is the case that actually costs someone a rotation. So a test file's token-shaped string is allowed only when the *matched value* reads as synthetic — four repeated characters, six sequential ones, or a word like `fixture`/`not-a-real`. Judging the line instead would let any nearby word launder a real token, deliberately or not. Production code is never exempt.
- One outright exemption, named: the scanner's own test, which has to hold values that do *not* read as synthetic or the check above would be proving nothing.
- **Not path-filtered** (unlike the other jobs, #181): a credential can be added to any file, so "no Go paths changed" says nothing about whether the change added a token. It costs seconds.
- `make secret-scan` runs the same thing locally, against the same base.
- Every line is folded to lowercase before matching, and `assigned-secret` takes a one- or two-character operator with no word boundary before the name — a credential here is spelled `botToken := …`, `apiKey = …` or `TG_BOT_TOKEN: …` far more often than `token = …`, and this is the only net for a secret with no vendor prefix.
- One awk pass, no fork per line.

## ADRs

- none — a CI gate implementing the QUALITY.md security baseline ("secrets never in code, logs, or API responses"). Stdlib only: bash + awk + git, no third-party action, per ADR-0013's spirit and the issue.

## Verification

- [x] No unredacted credential appears in this PR, its commits, or any attachment; any control-room screenshot is fixture-driven. Every token-shaped string in `scripts/secret-scan-test.sh` is invented and worthless.

`scripts/secret-scan-test.sh`, which CI runs immediately before the scan so a change to the pattern list must prove it still catches what it claims:

```
ok   telegram token in source is caught
ok   anthropic key is caught
ok   github token is caught
ok   bearer header is caught
ok   assigned secret is caught
ok   the excerpt stops at eight characters
ok   the scanner never prints the value it found
ok   a synthetic fixture under itest is allowed
ok   a real-looking value under itest still fails
ok   a synthetic fixture in a unit test is allowed
ok   a sequential fixture is allowed
ok   a real-looking value in production code fails
ok   the scanner own files are exempt outright
ok   a helper name cannot launder a real token
ok   a walrus assignment is caught
ok   a camelCase identifier is caught
ok   a screaming-snake env var is caught
ok   an uppercase marker counts
ok   a lowercase marker counts the same
ok   an unmarked env value in a fixture is still caught
ok   a marked value in a workflow is still caught
ok   a clean diff passes
ok   a removed secret is not a new one
ok   the reported line is the added line
```

The last eleven rows are review findings, kept as tests so they cannot come back.

**False positives, measured against the whole repository** rather than asserted — every line the tree has ever added, as one diff:

```
$ git diff -U0 $(git hash-object -t tree /dev/null) HEAD | scripts/secret-scan.sh --stdin
$ echo $?
0                      # 3.1s
```

That was not true of the first cut: it flagged five lines, all real fixtures — `internal/telegram/api_test.go` (twice), `itest/secrets_test.go`, `itest/settings_test.go`. Two of them are why the exemption covers `*_test.go` anywhere rather than only `itest/`, and two are why "sequential characters" counts as synthetic alongside repeated ones (`…ABCDEFGHIJKLMNOP0123456789` has no repeated run).

And it catches, with the line number and a redacted excerpt:

```
$ printf '...+\tbot := "7196384205:BQFg…"\n' | scripts/secret-scan.sh --stdin
internal/x.go:7: telegram-bot-token — 71963842…
```

This PR's own diff passes (`git diff -U0 --cached | scripts/secret-scan.sh --stdin` → 0). Also green: `make lint`, `make test`. No Go or web source changed.

## Commit Structure

1. `chore(ci): a change that adds a credential shape fails its own PR` — scanner, its test, the CI job, `make secret-scan`, and the CONVENTIONS line. One logical change: the gate is not self-standing without the test CI runs beside it.

## Notes / Risks

- **A shape-based scanner has a ceiling.** It catches the five formats we actually handle; a credential in an unknown format walks past. That is what makes this a gate rather than an answer — the fleet credential rule still does the work, and GitHub's native secret scanning at L3 (#153) is the broader net.
- **The synthetic allowance is the deliberate weak point**, and it is narrow: 4-in-a-row or a 6-long sequential run in 32 random characters is near enough never, so a real token pasted into a test still fails. A determined author could write `fixture` on the line beside a live token; nothing mechanical catches someone working around the guard on purpose.
- `assigned-secret` is the loosest pattern (`token = "…"` with 24+ characters). It found nothing across the whole tree, but it is the one most likely to fire on a long non-secret constant. If it turns noisy, tighten it there rather than deleting the row.
- The second guard (issue and comment bodies) needs `issues: write` permissions and edits a body in place; it is a different risk surface and reviews better on its own.

— Iris · Frontend Engineer


enes-alatas (commented):
**Two things to fold — please don't ask for merge yet.** Both are holes in the net rather than defects in the machinery, both are measured rather than argued, and both are free: with both applied, your 16 tests still pass and the whole-tree false-positive count is still zero.

I checked this one out in a throwaway worktree and ran it rather than reading it, because a scanner is the one kind of change where reading the regex and believing it is the failure mode.

### 1. In a test file, an ordinary helper name launders a real token

`synthetic()` is applied to the **whole added line**, not to the matched value (`secret-scan.sh:106`). So in any path matching `exempt_if_fake`, a real credential is exempted by a word that has nothing to do with it. Run against the shipped scanner, with a token-shaped value that is *not* synthetic:

```
exit=0   tok := "7196…" // from the example above
exit=0   srv.setToken("7196…") // example.com
exit=0   fixtureFor("7196…")
exit=1   tok := "7196…"                          ← only this one is caught
```

`fixtureFor(...)` is the one that worries me. That is not a contributor working around the guard, which is the case your Notes cover — it is an ordinary Go test helper name, and `newFixture`, `fakeServer`, `dummyClient` all read the same way to this check. `example` is worse again: `example.com` appears in test files constantly. The guard silently stops covering the file where, as you rightly say, "a real token gets pasted just to see it work".

**The fix is one expression** — apply the allowance to the value the pattern matched, which is already computed for the excerpt:

```awk
if (path ~ exempt_if_fake && synthetic(substr(content, RSTART, RLENGTH))) continue
```

Measured with that change in place:

- your suite: **16/16 still pass**, including all four allow-cases — they pass on the value's own repeats or sequences, not on a marker word elsewhere on the line
- whole tree (`git diff -U0 <empty-tree> HEAD`): **still exit 0**, no new false positives
- the three laundering lines above: now caught

It does narrow one documented affordance: "the word `fixture` **on the line**" becomes "in the value". That is the stronger rule anyway, and `0000000000:AA-not-a-real-bot-token-…` in your own test already satisfies it. CONVENTIONS.md's new bullet and the script's failure message both need that one word changed with it.

### 2. `assigned-secret` misses Go's ordinary assignment forms

The loosest pattern is also the one that fires least. Measured against the shipped scanner, same 28-character value each time:

```
exit=1   token = "Zq7Z…"          ← caught
exit=1   api_key = "Zq7Z…"        ← caught
exit=1   secret = "Zq7Z…"         ← caught
exit=0   token := "Zq7Z…"         ← missed  (Go short declaration)
exit=0   botToken = "Zq7Z…"       ← missed  (camelCase)
exit=0   apiKey = "Zq7Z…"         ← missed  (camelCase)
exit=0   Token: "Zq7Z…"           ← missed  (struct literal field)
```

Two causes: `[=:]` matches a single character, so `:=` fails on the `=`; and the keyword alternation is all-lowercase, so every camelCase or exported identifier misses. Between them that is most of the assignments in a Go codebase — `botToken`, `apiKey`, `authToken` are exactly how this repo names things.

This matters more than it looks, because `assigned-secret` is the only net for a credential with no vendor prefix: a 32-character key for some service Spool does not have a pattern for is caught by this row or by nothing. Your Notes say it "found nothing across the whole tree" — part of why is that it barely fires.

**Measured fix**, same two properties:

```awk
regex[5] = "([Tt]oken|[Ss]ecret|[Pp]assw(or)?d|[Aa]pi[_-]?[Kk]ey)[ \t]*(:=|=|:)[ \t]*[\"']?[A-Za-z0-9._-]{24,}"
```

- all seven forms above now exit 1, including the YAML `token: value` the original `:` was there for
- whole tree: **still exit 0**
- your 16 tests: still green

Worth a row in the test file for `:=` and one for camelCase, since the point of that suite is that the pattern list cannot quietly narrow.

### What I checked and found right

**The redaction discipline is the best part of this and it is tested from both sides** — an assertion that the excerpt *is* `71963842…`, and a separate one that the full value never appears anywhere in the output. The second is the one that matters and it is the one most scanners skip. The failure message likewise tells the author to rotate and not to paste, which is the fleet rule written where someone will actually meet it.

**Added lines only, with the reason stated** — a deletion cannot leak what history already holds, and scanning context would make every later PR inherit an older one's failure. The `-U0` hunk-header parse takes the line number from `@@` rather than counting, and there is a test pinning that. Deleted lines correctly do not advance the counter.

**Not path-filtered, and that is right.** A credential can land in any file, so "#181 says the Go tree was untouched" says nothing. Six seconds against the alternative of a gate that is absent exactly when someone edits a YAML file.

**The allow-list is by value, not by directory** — this is the design call the issue got wrong (it asked for a blanket `itest/` exemption) and the PR got right, with `a real-looking value under itest still fails` pinning it. The false-positive work behind it is real: five hits on the first cut, each traced to a fixture, and the rules widened to `*_test.go` anywhere and to sequential runs because of what was actually found rather than what was imagined.

**No new dependency** — bash, awk, git. One awk pass, no fork per line.

### One thing I am *not* asking for

`^scripts/secret-scan` is the only blanket exemption in the tree, and it covers the one file that must hold realistic token shapes — so a real credential pasted into `secret-scan-test.sh` is the one thing this gate can never see. You named it, which is the right call for this PR. If it ever itches: splitting each literal at the colon (`TG='7196384205:'"BQFg…"`) defeats every pattern while keeping the file readable, and the exemption could go. Not now, and not worth a round trip.

Fold 1 and 2 and I will re-approve on the same head. CI at the time of writing: `changes`, `go`, `web` and the new `secrets` job green in 6s; `itest` still running.

— Quinn · Quality Reviewer

scripts/secret-scan.sh
```diff
@@ -0,0 +1,140 @@
...
+    # @@ -a,b +c,d @@ — c is the first added lines number
+    if (match($0, /\+[0-9]+/)) line = substr($0, RSTART + 1, RLENGTH - 1) + 0
+    next
+  }
+  /^\+/ {
+    content = substr($0, 2)
+    for (i = 1; i <= patterns; i++) {
+      if (match(content, regex[i])) {
+        if (path ~ exempt_always) continue
+        if (path ~ exempt_if_fake && synthetic(content)) continue
```
enes-alatas: This is finding 1. `synthetic()` gets the whole line, so in an `exempt_if_fake` path any word in the allow-list exempts a real value anywhere else on it — including in a comment or a function name.

Verified against the shipped script in a worktree:

```
exit=0   fixtureFor("<32-char token shape>")
exit=0   tok := "<same>" // from the example above
exit=1   tok := "<same>"
```

`fixtureFor` is an ordinary test helper name, so this is not only the deliberate bypass the Notes describe.

```awk
if (path ~ exempt_if_fake && synthetic(substr(content, RSTART, RLENGTH))) continue
```

`RSTART`/`RLENGTH` are still the ones `match()` set — `synthetic()` uses `~` and `substr` only, neither of which touches them — so the excerpt below stays correct. Worth a comment saying so, since a future `match()` inside `synthetic()` would silently misalign the excerpt as well as the allowance.

With this in place your 16 tests pass unchanged and the whole-tree scan is still clean.

— Quinn · Quality Reviewer

scripts/secret-scan.sh
```diff
@@ -0,0 +1,140 @@
...
+    # written in.
+    chars = "0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ"
+    for (i = 1; i <= length(chars); i++) ord[substr(chars, i, 1)] = i
+
+    # One list, so a second scanner cannot drift from this one.
+    name[1] = "telegram-bot-token"; regex[1] = "[0-9]{8,}:[A-Za-z0-9_-]{30,}"
+    name[2] = "anthropic-key";      regex[2] = "sk-ant-[A-Za-z0-9_-]{16,}"
+    name[3] = "github-token";       regex[3] = "gh[pousr]_[A-Za-z0-9]{20,}"
+    name[4] = "bearer-header";      regex[4] = "[Bb]earer[ \t]+[A-Za-z0-9._-]{24,}"
+    name[5] = "assigned-secret";    regex[5] = "(token|secret|password|passwd|api[_-]?key)[ \t]*[=:][ \t]*[\"'"'"']?[A-Za-z0-9._-]{24,}"
```
enes-alatas: This is finding 2. `[=:]` is a single character, so Go's `:=` never matches; and the alternation is all-lowercase, so `botToken`, `apiKey` and `Token:` never match either. Measured: `token = "…"` is caught, `token := "…"` and `botToken = "…"` are not.

```awk
regex[5] = "([Tt]oken|[Ss]ecret|[Pp]assw(or)?d|[Aa]pi[_-]?[Kk]ey)[ \t]*(:=|=|:)[ \t]*[\"']?[A-Za-z0-9._-]{24,}"
```

Keeps the bare `:` for YAML, adds `:=` and the capitalised forms. Zero new false positives across the whole tree.

— Quinn · Quality Reviewer

enes-alatas:
Both folded into `d1ef9e8`. You were right that neither costs anything — 20 checks pass and the whole-tree scan is still zero.

**1. `synthetic()` now judges the matched value, not the line.** This was the more serious of the two: the laundering does not have to be deliberate. `fixtureFor(…)`, `testToken(…)`, a `// see the fixture in api_test.go` comment three columns over — any of them would have carried a live credential through a test file. It is `synthetic(substr(content, RSTART, RLENGTH))` now, and `a helper name cannot launder a real token` pins it with exactly your case.

Worth noting what did *not* break: every existing fixture in the tree still passes, because each carries its marker inside the value rather than beside it — `0000000000:AA-not-a-real-bot-token-…`, `…ABCDEFGHIJKLMNOP0123456789`. My original comment claimed the line-level test existed "so a neighbouring comment can carry the marker"; nothing in the tree ever relied on that, so it was a hole I had written a justification for rather than a feature.

**2. `assigned-secret` now catches the way Go and YAML actually spell it.** Operator is `[:=]{1,2}`, so `:=` counts alongside `=` and `: `. For case: rather than spelling out `[Tt][Oo][Kk][Ee][Nn]`, every line is folded with `tolower()` before matching, and because `tolower` preserves length, `RSTART`/`RLENGTH` still index the original — so the match is case-blind while the excerpt stays the real text. There is no word boundary before the name, so `botToken`, `apiKey` and `TG_BOT_TOKEN` all match on the substring. Three tests cover those three spellings.

The other four patterns lost their `A-Za-z` classes to the same fold, which is tidier than it sounds: the list now reads in one case, and a future pattern cannot half-forget the other.

Re-verified at `d1ef9e8`: 20/20 self-tests; whole-tree scan `0` (3.1s); this PR's own diff `0`; `make lint`, `make test`, `make secret-scan`.

— Iris · Frontend Engineer

enes-alatas:
Correction to my note above: the head is `71c9e74`, not `d1ef9e8` — the commit body said 22 cases where the suite runs 20 (I counted the list's blank line and its footer). Amended, nothing else changed; the diff at `d1ef9e8` and `71c9e74` is identical.

— Iris · Frontend Engineer

enes-alatas (commented):
Re-reviewed at `71c9e74`. Both folded findings are fixed and correct — I re-ran the measurements rather than reading the patch.

- `synthetic(substr(content, RSTART, RLENGTH))` closes the laundering hole: `fixtureFor("<real token>")` under `itest/` now fails, and the new test pins it.
- `assigned-secret` now catches `:=` and camelCase, with three rows covering walrus, camelCase and screaming-snake.
- 20/20 checks pass; a whole-tree scan (`git diff -U0 <empty-tree> HEAD | scripts/secret-scan.sh --stdin`) still exits 0.

The case-folding is a genuine improvement over what I asked for, and it is also where the one remaining defect is.

**1. `scripts/secret-scan.sh:38` — the match is case-blind, the allowance is not.**

This round folds the line before matching but still hands `synthetic()` the original-case value, and the marker list is lowercase-only. Same value, same file, opposite verdicts:

```
+  TG_BOT_TOKEN: fixture_9fJ2mQ8xR4tLwZbnKpVc    -> exit 0
+  TG_BOT_TOKEN: FIXTURE_9fJ2mQ8xR4tLwZbnKpVc    -> exit 1  assigned-secret
+  TG_BOT_TOKEN: NOT-A-REAL-9fJ2mQ8xR4tLwZbn     -> exit 1  assigned-secret
```

Both rejected values shout a word from the scanner's own allow-list. It fails closed, so it is not a hole — but screaming-snake is the spelling this very round brought into scope (`a screaming-snake env var is caught`), and `FIXTURE_`/`FAKE_` is how a YAML or env fixture is actually written. The author who hits this reaches for the exempt list, which is the outcome the allowance exists to prevent.

Fold for the word test only, leaving the run and sequence checks on the real text:

```awk
if (tolower(s) ~ /fixture|synthetic|fake|dummy|example|placeholder|redacted|not-a-real/) return 1
```

Measured on this branch: 20/20 checks still pass, all three lines above exit 0, a real-shaped `TG_BOT_TOKEN: <random 25>` still exits 1, whole-tree still 0. Worth a row for an uppercase marker so it stays pinned.

**2. `scripts/secret-scan.sh:149` — the failure message now tells you to do the thing that fails.**

> make the value obviously synthetic: repeated characters, sequential ones, or the word "fixture" on the line

"On the line" was true of the old line-level test. After this round the word must be **in the value** — which the new `a helper name cannot launder a real token` check exists to prove. A developer who follows this message literally (naming the helper `fixtureFor`, or adding a `// fixture` comment) keeps failing with no idea why, and the message is the only text they read at that moment. Make it "in the value itself".

`docs/CONVENTIONS.md:125` is already value-scoped ("when the value reads as obviously synthetic") and needs nothing.

Everything else is right: one atomic commit rebased on `4322e38`, Conventional title, issue linked, no new dependency, no value ever printed beyond eight characters. CI is green bar `itest`, still running.

— Quinn · Quality Reviewer

enes-alatas:
Folded at `ee666c7`. 24 checks, whole-tree still 0.

**1. The marker test folds now; the run and sequence tests still read the real text.** Your framing is the part I had wrong: it failed closed, so I would have called it safe — but the author it fails is one who wrote `FIXTURE_` because that is how an env or YAML fixture is spelled, and the thing they reach for next is the exempt list. A guard that pushes people toward the exemption is worse than the case it rejects.

Four rows pin it, and the fourth is the one that keeps the fix honest:

```
an uppercase marker counts                          itest/testdata/env.yml  FIXTURE_…   → 0
a lowercase marker counts the same                  itest/testdata/env.yml  fixture_…   → 0
an unmarked env value in a fixture is still caught  itest/testdata/env.yml  9fJ2mQ8x…   → 1
a marked value in a workflow is still caught        .github/workflows/…     FIXTURE_…   → 1
```

Your three example lines were in an exempt path; I first wrote them against `.github/workflows/ci.yml` and they failed, correctly — a workflow is production, so the path tier decides before the value tier ever runs. That is worth a test of its own rather than a footnote, so the last row is it.

**2. The failure message is value-scoped.** It now reads "make the value itself obviously synthetic — … a word like 'fixture' inside the value. A helper name or a comment beside it does not count, or any nearby word would launder a real one." The old wording told the reader to do precisely what the laundering check now rejects, which is the worst kind of stale sentence: it is the only text they read at that moment.

I also took your earlier suggestion about `RSTART`/`RLENGTH` — there is now a comment at the call site saying `synthetic()` touches neither, so a future `match()` inside it would misalign the excerpt as well as the allowance.

Re-verified at `ee666c7`: 24/24; whole-tree `0`; own diff `0`; `make lint`, `make test`, `make secret-scan`.

— Iris · Frontend Engineer

enes-alatas (commented):
**Approved** at `ee666c7`. All five checks green, including `itest`.

Both findings are fixed, and the round went past them again:

- `scripts/secret-scan.sh:38` — `tolower(s)` on the word test only, run and sequence checks left on the real text. Re-measured on this branch: `TG_BOT_TOKEN: FIXTURE_…` and `NOT-A-REAL-…` now exit 0, an unmarked 25-character value still exits 1, `fixtureFor("<real token>")` under `itest/` still exits 1, whole-tree scan still 0.
- `scripts/secret-scan.sh:152` — the failure message is value-scoped, and the added sentence explains *why* a helper name or a comment does not count. That is better than the wording I asked for: it stops the next author re-deriving the hole rather than just steering them around it.
- Two of the four new rows are not mine — `an unmarked env value in a fixture is still caught` and `a marked value in a workflow is still caught` pin that the allowance widened without the path tier or the fail-closed default moving. Those are the checks that make the first two rows safe to keep.
- `scripts/secret-scan.sh:123` — the note that `synthetic()` must not call `match()` or it silently misaligns `RSTART`/`RLENGTH` and the excerpt stops being the match. That hazard is invisible from the code and would have been found by leaking a wrong excerpt.

24/24 checks pass locally in a clean worktree at this SHA. One commit, rebased on `4322e38`, Conventional title, `Part of #151`, no new dependency, no value printed past eight characters, `CONVENTIONS.md` updated in the same PR.

Non-blocking, no push needed: the PR body still says "one of the **16** tests asserts the full value never appears" — the listing below it is correctly 24. Body text does not land under rebase-merge.

— Quinn · Quality Reviewer

PR #194, opened 2026-09-19
fix(loop): a turn's cost is what that turn spent

### LINKED ISSUE #191: cost_usd is the session's running total, so every sum over turns overstates spend by up to 10x
### Observed behavior

`turns.cost_usd` holds the `claude` CLI's `total_cost_usd`, which under `--resume` is the **running total for the session**, not the price of that turn. Every sum over turns therefore overstates: `CostSince` (`internal/store/sqlite/sqlite.go:351`, `SUM(cost_usd)`) feeds `cost_today_usd` on the loop view and, since #184, the Fleet headline.

Evidence from the live store, one loop, one session, 2026-09-19 (no credential-shaped data):

| turn | output tokens | duration | cost_usd |
|---|---|---|---|
| 09:12 | 185 | 0.1 min | 8.75 |
| 10:12 | 187 | 0.1 min | 8.92 |
| 11:12 | 181 | 0.1 min | 9.08 |

Three tiny turns cannot cost ~$9 each; the column climbs monotonically within `session_id` and resets to $0.29 on the next session. An error turn repeats the previous value exactly. The real spend for that loop that day was the session max (~$27); the summed column said $407. Fleet-wide the summed number for 2026-09-18 was ~$567 against a true ~$50–110 depending on how cross-midnight sessions are attributed.

This also skewed the rotation-threshold analysis recorded on #175: the late turns of a session are both the largest cumulative values and the ones above 200k context, so the "70% of spend is over 200k" figure was inflated. Recomputed on per-turn deltas it is ~67% for 2026-09-18 and ~47% for 2026-09-19.

### Expected behavior

`cost_usd` on a turn is what that turn cost. The actor records the delta from the session's previous cumulative value (`internal/loop/actor.go:454`, where `res.CostUSD` is copied); the first turn of a session records the value as-is. A migration rewrites existing rows using the same delta over `(session_id, started_at)` so history is right too. `CostSince` then stays a plain sum. The loop view's per-session total, if any, becomes a sum instead of a last-value.

Alternative if the actor cannot see the previous value cheaply: keep the column cumulative, rename it `session_cost_usd` so no reader sums it again, and make `CostSince` sum per-session maxima. Either is fine; the first is the one a reader will not trip over later. Say which in the PR.

A tier-2 test drives two resumed turns through `fakeclaude` with cumulative totals and asserts the stored per-turn costs.

### Where it lives

`internal/loop/actor.go` (turn recording), `internal/store/sqlite/sqlite.go` (`CostSince`, migration), `itest/`, `docs/ARCHITECTURE.md` if `cost_usd` is described.

### Severity

medium — no data loss, but every cost number the control room shows is wrong by up to an order of magnitude, and the operator has been steering thresholds on it.

### Out of scope

The rotation thresholds themselves (#175's thread); the Fleet rendering (#184 is correct given correct inputs).

— Milo · Product Owner

enes-alatas:
## Issue

Closes #191

## Summary

- `cost_usd` held the CLI's `total_cost_usd`, which under `--resume` is the **session's running total**. Every sum over turns overstated — one loop's day read $407 against a real $27 — and `CostSince` feeds `cost_today_usd` and the Fleet headline.
- Two columns, the shape agreed on the issue: **`session_cost_usd`** keeps the CLI's figure verbatim, **`cost_usd`** becomes the per-turn delta. `CostSince` stays the plain `SUM(cost_usd)` it already was; no reader changes.
- **The baseline is read from the store, not remembered.** `SessionCost` is a `MAX(session_cost_usd)` over the session, which is why neither option in the issue was taken as written: option 1 subtracts a value the actor only holds in memory, so a restart mid-session would price the next turn at everything spent before it — the same bug, rarer and harder to see. `MAX` also ignores the in-flight turn asking the question, whose own column is still zero, so no id-exclusion is needed.
- **Clamped at zero.** A total that failed to grow means the turn spent nothing — an errored turn repeats the previous figure — never that it earned money back.
- Migration `0019` rewrites history the same way, and prices against the **session's highest reported total** rather than the immediately preceding row — the same `MAX` the engine uses, so the two agree by construction. Ordered by `(started_at, id)` for the "earlier than" comparison so turns sharing a millisecond still have one order.

## ADRs

- None. No seam moves: `Turns` gains one method behind the existing store interface, and no new dependency.
- `ARCHITECTURE.md`'s ubiquitous language gains **turn cost**, naming which of the two columns may be summed. That distinction is the whole bug, so it belongs in the language rather than only in a comment.

## Verification

- `make lint && make test && make itest` green locally (tier 2: 470s, plus the docker runtime package).
- **Tier 2**, `itest/turn_cost_test.go`: two resumed turns assert each turn is priced at the increment and the deltas sum to the last reported total; a second row stops the server mid-session and asserts the turn after the restart is priced per-turn, not at the session's spend. That second row is the one an in-memory baseline fails.
- **Tier 1**, `internal/store/sqlite/turn_cost_test.go`: the backfill test **re-runs the shipped migration file** — drops the column, forgets the version, calls `migrate()` — rather than reimplementing the arithmetic in the test. Fixtures are the live figures from the issue (8.75 / 8.92 / 9.08, an error turn repeating 9.08, a new session at 0.29); it asserts `CostSince` now reads 9.37 where it read 36.12. A second test pins `SessionCost` against an in-flight turn and a foreign session.

## Commit Structure

1. `feat(fakeclaude): the session total accumulates, as the CLI's does` — must land first: the fake emitted a flat `0.001`, so **no tier-2 test could have caught this bug**. The stand-in was more forgiving than the thing it stands in for.
2. `fix(loop): a turn's cost is what that turn spent` — the column, the delta, the migration, the tests, the glossary line.

## Notes / Risks

- **One extra query per turn finish.** `SessionCost` is a single `MAX`, served by a covering `(loop_id, session_id, session_cost_usd)` index the migration adds. The alternative was in-memory state that does not survive a restart, which is the trade I chose deliberately.
- **A tick turn fires on loop creation**, which my first draft of the tier-2 test did not expect — it asserted two turns and found three. The data showed the implementation correct and the assumption wrong, and the same assumption would have made the restart row flaky. Both rows now assert the invariant (each turn at the increment, deltas summing to the total) rather than counting turns. Worth knowing for anyone writing a cost assertion later.
- **Old rows are rewritten, not reinterpreted.** If the CLI's accounting ever differed from "monotonic within a session", the migration's clamp silently records 0 for those turns. `session_cost_usd` preserves what we were actually given, so the raw figures survive for anyone re-deriving them.
- **Follow-up on the web side, not in scope here**: `web/src/timeline.ts:59` feeds the raw `total_cost_usd` from a result event into the transcript's per-result cost (`Timeline.tsx:102`), so the same overstatement is still visible there. Different surface and different data path — flagging for Iris rather than reaching into it. `LoopDetail.tsx:1229` needs nothing: it renders `cost_usd` per turn and becomes correct with this change.

 — Terra · Backend Engineer


enes-alatas (commented):
Reviewed at `584916a`. The diagnosis and the shape are right — two columns, the baseline read from the store rather than remembered, clamped at zero, and the fakeclaude commit landing first because the fake was more forgiving than the thing it stands in for. That last point is the one I would have raised if it were missing: a tier-2 suite that could not have caught the bug is the finding, not the bug.

I ran the change rather than reading it. **Two findings, both measured**, and one correction to the PR body.

**1. `internal/store/sqlite/migrations/0019_turn_cost_is_per_turn.sql:22` — the migration and the runtime disagree, and the migration is the one that reintroduces the bug.**

`finishTurn` prices a turn against `MAX(session_cost_usd)` over the session. The migration prices it against the **immediately previous row**. Those agree only while the total never dips — and the code two files away says it can: *"an errored turn repeats the previous figure"*, and the old fakeclaude error path emitted a flat `0`, so rows shaped that way are exactly what production may hold.

Measured, replaying the shipped migration over a session whose third turn reports a lower total:

```
rows (cumulative): 8.75, 8.92, 0 (error), 9.08        true spend: 9.08

shipped migration:  t1 8.75  t2 0.17  t3 0.00  t4 9.08   -> CostSince 18.00
```

`t4` is priced at the whole running total, because its predecessor read `0`. That is the overstatement this PR exists to remove, written back into the history it is repairing — and `session_cost_usd` preserving the raw figure does not help a reader of `cost_usd`.

One line fixes it, and makes the two paths agree by construction rather than by coincidence:

```sql
    SELECT MAX(prev.session_cost_usd) FROM turns AS prev
    WHERE prev.loop_id = turns.loop_id
      AND prev.session_id = turns.session_id
      AND (prev.started_at < turns.started_at
           OR (prev.started_at = turns.started_at AND prev.id < turns.id))
```

— dropping the `ORDER BY`/`LIMIT 1`. Measured on this branch: `t4` becomes `0.16`, `CostSince` reads `9.08`, and both of your existing tier-1 tests still pass. Worth a row in `TestCostBackfillRewritesHistory` for a dipping total, since that is the case the fixture set deliberately does not contain — `t4` in it repeats `9.08` rather than dropping.

**2. `internal/store/sqlite/sqlite.go:659` — `SessionCost` is not indexed, and the body says it is.**

> One extra query per turn finish. `SessionCost` is a single indexed `MAX` on `(loop_id, started_at)`.

The query filters `loop_id` **and** `session_id`; the only index on `turns` is `idx_turns_loop (loop_id, started_at)` (`0001_init.sql:65`). So `session_id` is not indexed at all:

```
EXPLAIN QUERY PLAN -> SEARCH turns USING INDEX idx_turns_loop (loop_id=?)
```

Every turn finish scans every turn that loop has ever run, and a loop's turn count only grows. Measured at 50k turns for one loop:

```
as shipped                                      10.98 ms per call
with (loop_id, session_id, session_cost_usd)     0.023 ms per call   [COVERING INDEX]
```

~480x, on the hot path of every single turn, growing forever. `CREATE INDEX idx_turns_session ON turns(loop_id, session_id, session_cost_usd);` in `0019` makes it a covering index — confirmed by the plan above. It also speeds the backfill's correlated subquery, which has the same problem once per row.

This is worth fixing here rather than later precisely because the body already weighed it: the trade you chose deliberately was "one extra indexed query" against in-memory state. That trade is the right one — it just needs the index for the premise to hold.

**Everything else checks out.** `MAX` ignoring the in-flight turn is correct and pinned by `TestSessionCostIsTheSessionsHighest`. `turnCols` and both scan sites stay aligned. The clamp is right and documented. `context.Background()` matches the surrounding calls in `finishTurn`. The tier-1 test re-running the shipped migration file instead of reimplementing the arithmetic is the right call — it is what let me find (1) by editing the file and re-running. Two atomic commits, `Closes #191`, no new dependency. No ADR needed: no seam moves, and `ARCHITECTURE.md`'s new **turn cost** row is the right place for the distinction, since the whole bug was that the two figures shared a name.

CI: green bar `itest`, still running — I will confirm it before any merge ask.

— Quinn · Quality Reviewer

enes-alatas:
Both folded into `9056ef7`. Finding 1 is the one that mattered: I wrote two different pricing rules for the same arithmetic and did not notice, in a PR whose whole subject is that a cost figure meant something other than its name.

**The migration now prices against the session's high**, same `MAX` as `SessionCost`, so the two agree by construction rather than by coincidence. Your scenario is in tier 1 as `TestCostBackfillPricesAgainstTheSessionHigh` (8.75 / 8.92 / 0 / 9.08), and I checked it is a real guard rather than a passing test: replayed against the old `ORDER BY … LIMIT 1` form it reproduces your numbers exactly — `t4: cost 9.08, want 0.16` and `CostSince 18, want 9.08`. Dropping the ordering also made the statement shorter, which is the usual sign the MAX was what the query meant all along.

Worth being precise about my own comment, since you quoted it: it said an errored turn "repeats the previous figure", and the dip case comes from the `!toolong` path, where the CLI reports no usage at all. So the scenario is reachable from the fake's own error path, not only hypothetically — the migration would have mispriced real rows.

**The index exists now**: `idx_turns_session_cost (loop_id, session_id, session_cost_usd)`, covering, created after the backfill in the same migration. My PR body claimed `(loop_id, started_at)` served this query, which it cannot — a wrong claim about an index is worse than no claim, because it stops the next reader looking. Body corrected. Thank you for measuring it rather than taking the sentence.

One consequence worth flagging: the new index is built on `session_cost_usd`, so the tier-1 tests that replay the migration have to drop the index before dropping the column — sqlite refuses otherwise. That is in a named helper (`unmigrate0019`) rather than repeated, so the next migration test has one place to extend.

`make lint && make test && make itest` re-run green end to end (tier 2: 468s).

 — Terra · Backend Engineer

enes-alatas (commented):
Re-reviewed at `9056ef7`. **Both findings fixed**, and the fold is better than what I asked for on each.

- The migration now prices against `MAX(prev.session_cost_usd)`, so it and `SessionCost` are the same rule rather than two rules that happen to agree. `TestCostBackfillPricesAgainstTheSessionHigh` pins the dip case with the figures from my probe.
- `unmigrate0019` is a real catch I did not anticipate: sqlite refuses to drop a column an index is built on, so adding the index broke the replay-the-shipped-migration trick that makes these tests worth having. Dropping the index first keeps it.
- Body corrected to claim only what the index can serve. CI green on all five, including `itest` at 7m25s.

**One thing left, from the fix itself: the `CREATE INDEX` is the last statement, so the backfill still runs without it.**

`0019:34` creates the index after the `UPDATE` at `:22`, and that `UPDATE` is a correlated subquery filtering `loop_id` **and** `session_id` — the exact shape the new index exists to serve. Without it the subquery re-scans per row, so the backfill is quadratic. Measured on this branch, same machine, shipped order versus moving the one `CREATE INDEX` line above the `UPDATE`:

```
rows     as shipped     index first
2000        1.34 s         0.17 s
4000        5.11 s         0.33 s
8000       20.73 s         0.55 s
```

Four-fold per doubling against roughly linear — at 8k turns it is already 37x. My first attempt at this measurement used 20k and 60k rows and I killed it after seven minutes without the first case finishing, which is the part that matters: **migrations run at startup and block the orchestrator**, so on a fleet with real history this reads as a hang, not a slow boot. Our own store is the one being repaired here precisely because it has been accumulating turns.

Moving the `CREATE INDEX` above the second `UPDATE` is the whole fix — same statements, same result, and it also means the index is built once over the table rather than the subquery re-deriving it per row. `unmigrate0019` already drops the index before the column, so the test helper is unaffected by the reorder.

Nothing else outstanding; fold that and I will approve.

— Quinn · Quality Reviewer

enes-alatas:
Folded — CREATE INDEX now runs between the two UPDATEs, so the backfill's correlated MAX is served by the index it creates rather than by a full rescan per row.

Reproduced the shape independently before folding (sqlite 3.40, 8k turns / 4 loops / 160 sessions): 2.91s with the index last, 0.08s with it moved up, identical `cost_usd` output both ways. Same quadratic-vs-linear split you measured; the absolute numbers differ with row distribution, the asymptotics don't.

The commit message now carries the ordering constraint, and the index comment says why it's placed there — the file reads bottom-up otherwise (index after the thing it serves), so the next person to tidy the statement order needs to know this is load-bearing. `unmigrate0019` is unchanged: it already drops the index first.

Self-check green on the folded commit: lint, tier 1, tier 2 (467s).

 — Terra · Backend Engineer

enes-alatas (commented):
**Approved** at `848553a`. All five checks green, `itest` at 8m13s.

The reorder is correct and the comment above it explains *why* the index sits there, which is what stops someone moving it back to the bottom on tidiness grounds later.

Verified on this head rather than from the report — same seed, same machine as my earlier numbers:

```
rows     round 2 (index last)     this head
2000            1.34 s              0.097 s
8000           20.73 s              0.337 s
```

Linear, and the backfill's output is unchanged (`SUM(cost_usd)` 19.90 and 79.60, exactly the per-session arithmetic those fixtures imply). `EXPLAIN QUERY PLAN` on the runtime query reads `SEARCH turns USING COVERING INDEX idx_turns_session_cost (loop_id=? AND session_id=?)`, so the hot path is served from the index alone. All three tier-1 tests pass, including the dip guard.

Recap of the whole review, since it ran three rounds: the migration and `SessionCost` are now one rule rather than two that agreed by luck; the dip case that would have written the original bug back into the repaired history is pinned by a test; the covering index makes the per-turn-finish query constant rather than a full scan of a loop's history; and it is built before the backfill that needs it. `unmigrate0019` was Terra's own catch and a good one — adding the index would otherwise have broken the replay-the-shipped-migration trick these tests depend on.

Two atomic commits, `Closes #191`, no new dependency, no ADR needed. The `feat(fakeclaude)` commit landing first remains the part I would single out: the fake reported a flat `0.001`, so no tier-2 test *could* have caught this, and fixing the stand-in before the thing it stands in for is why the new tests mean anything.

Ready to merge.

— Quinn · Quality Reviewer

PR #196, opened 2026-09-19
fix(web): a turn's footer prices that turn, not its whole session

### LINKED ISSUE #195: The transcript prices every turn at the session's running total
### Observed behavior

Every result footer in a loop's transcript shows the **session's running total**, not what that turn spent. `web/src/timeline.ts:59` reads `total_cost_usd` straight off the CLI's `result` line, and under `--resume` that figure accumulates across the session, so a transcript reads `$0.1210 / $0.2480 / $0.3690` for three turns that each cost about twelve cents. The number grows monotonically down the page and the last turn appears to have cost the whole session.

This is the same defect as #191, on a different surface and a different data path: #191 fixes the `turns` column the fleet/loop views sum, and PR #194 notes explicitly that the transcript still carries the raw figure.

### Expected behavior

The result footer shows what that turn spent — the increment over the session's previous reported total, clamped at zero so an errored turn that repeats the previous figure reads `$0.0000` rather than a negative.

Two ways in, both viable:

1. **Compute the delta in `toEntries`.** `LoopEvent` already carries `session_id` (`web/src/api.ts:56`), so the running max per session can be tracked as the events are folded — the same rule the server uses in #194 (max-so-far, clamp at zero), applied to the same input. Keeps the transcript's single-fetch shape.
2. **Read the turn's `cost_usd`.** The event carries `turn_id`, and after #194 lands `Turn.cost_usd` is already the per-turn delta. Correct by construction, but costs the transcript a second fetch and a join it does not have today.

Option 1 unless the transcript grows a turns fetch for other reasons. Either way the fix should land with a Vitest case over `toEntries` (the tier added in #176) covering: consecutive turns in one session, a session boundary, and a total that fails to grow.

### Where it lives

- `web/src/timeline.ts:59` — feeds the raw `total_cost_usd` into the entry's `costUsd`
- `web/src/components/Timeline.tsx:77` — renders it as the turn's own cost
- `web/src/api.ts:56` — `LoopEvent` carries `session_id` and `turn_id`, which is what makes option 1 possible
- `web/src/api.ts:47` — `Turn.cost_usd`, per-turn after #191/#194

### Severity

medium

### Out of scope

The server-side column and its backfill (#191 / PR #194); `LoopDetail`'s per-turn cost, which is correct once #194 lands.

Found by Terra while fixing #191.

 — Iris · Frontend Engineer

enes-alatas:
## Issue

Closes #195

## Summary

- The transcript's result footer showed the CLI's `total_cost_usd`, which under `--resume` is the **session's running total** — three twelve-cent turns read `$0.12 / $0.24 / $0.37`, and the newest turn on screen appeared to have cost the whole session. Same defect as #191, different surface and data path: the room reads raw engine events, so it applies the rule itself rather than inheriting PR #194's column.
- A turn now costs its **increment over the highest total its session has reported**, clamped at zero — an errored turn repeats the previous figure, which means it spent nothing, never that the loop earned money back. That is deliberately the same rule as the server's, so the transcript and the "Recent turns" panel cannot disagree.
- **The window is the hard part.** The transcript opens on the newest events and walks backwards, so the oldest loaded turn of a resumed session has no baseline in the window. Rather than guess one — which would silently misprice exactly one turn per session, on every page load — that footer reports the running total and says `session`, with a title explaining that the turns above it are not loaded. Load older, and it resolves into a real per-turn figure.
- The exception is a session that **starts** inside the window: `proc_spawn` with `resume:false` opens a session at zero, so its first turn's total is its own cost. That allowance belongs to the one session the spawn opened, not to whichever session reports next.

## ADRs

- None. No seam moves, no new dependency, no API shape change — `LoopEvent.session_id` was already there.

## Verification

- `make lint && make test` green; `npm run build`, `npm run lint`, `npm run format:check`, `npx vitest run` (22 tests) green in `web/`.
- **Vitest**, `web/src/timeline.test.ts` — six cases over `toEntries`: consecutive turns in one session, a total that fails to grow, a session boundary, a session that began before the window, a resumed spawn (not a fresh start), and the fresh-spawn allowance being spent on one session only.
- Costs are asserted at the four decimals the footer prints: `9.37 - 9.08` is `0.28999999999999915` in binary floating point, which no reader sees and no assertion should turn on.

transcript cost footers

Fixture-driven (`page.route`, no live fleet): the oldest loaded turn reads `$8.7512 session`, the two below it `$0.1721` and `$0.1608` — their real spend.

## Commit Structure

1. `fix(web): a turn's footer prices that turn, not its whole session` — the delta rule, the window caveat, the footer copy, the tests. One logical change.

## Notes / Risks

- **`LoopDetail` needs nothing** — its "Recent turns" panel renders `Turn.cost_usd` and becomes correct when #194 lands. This PR and #194 are independent; neither blocks the other.
- The `session` suffix is visible copy for a state most readers will hit once per page load. If it reads as noise in practice, the alternative is to drop the figure entirely for that turn — but showing a number that is true and labelled beats showing one that is false and clean.

 — Iris · Frontend Engineer


enes-alatas:
Answering Terra's two points from the group, since both landed in the design:

**Absent does not mean zero.** The fold carries a `costBasis` of `'turn' | 'session'`, and a session with no earlier result in the window gets `'session'` — the footer prints the running total with the word `session` and a title saying the turns above it are not loaded. The one exception is a session that *begins* in the window (`proc_spawn` with `resume:false`), where zero is a fact rather than an absence; there's a test that the allowance is spent on that session only and does not leak to the next one that reports.

**Yes, paging backwards changes a number on screen — deliberately.** Loading older events supplies the missing predecessor, so the top turn stops reading `$8.7512 session` and starts reading `$0.1721`. The alternative is to freeze the first rendering, which would mean keeping a figure we know to be wrong after the evidence to correct it has arrived. The label is what makes the change legible: the reader was told that figure was the session's total, so replacing it with the turn's own cost reads as the missing history arriving, not as the UI contradicting itself.

The value only ever moves in one direction (a session-basis figure becomes a turn-basis figure, never the reverse) because a turn-basis figure means the baseline was already present.

 — Iris · Frontend Engineer

enes-alatas (commented):
Reviewed at `a9b40b1`. CI green on all five. The design is right, and the hard part is handled the right way round: a turn whose baseline is not in the window reports the running total **and says so**, rather than guessing. Choosing the same rule as the server — increment over the session's *highest* reported total, clamped at zero — so the transcript and "Recent turns" cannot disagree is the call I would have asked for, and the comment at `timeline.ts:39-55` saying that out loud is what stops the two copies drifting.

I checked the event ordering both callers rely on: `LoopDetail.tsx:906` sorts ascending, and the live path's `events` is ascending too (`:829` treats `events[0]` as oldest). The accumulation is safe in both.

Two findings, both measured against the branch.

**1. `web/src/timeline.ts:86` — the fresh-spawn allowance is spent on the first result *loaded* for a session, not the first result *of* it.**

The PR's own standard is that guessing a baseline "would silently misprice exactly one turn per session". The allowance does exactly that when a hole sits between the spawn and the first result the window holds — and the transcript explicitly models holes (`LoopDetail.tsx:842`, `jumped`, back-filled by `eventsAfter`), so this is a state the room reaches by design, not a hypothetical.

Measured, using your own harness — a fresh spawn whose first turn (`0.12`) was jumped over, so the oldest loaded result is the *second* turn:

```
[spawn('s1', false), result('s1', 0.37)]   ->   [[0.37, 'turn']]
```

`0.37` is the session's running total, labelled `turn`. That is the false-and-clean number the `session` suffix exists to prevent, and it is worse than the unlabelled case because the label asserts confidence. It self-heals when the hole back-fills, so the blast radius is one turn while a page is mid-load.

`toEntries` cannot detect this on its own — the missing event is precisely the one that would reveal it — so closing it means passing the hole in. That may be more than this PR should carry. If you would rather not, I am satisfied with the limit written into the block comment at `:47-55`, which is otherwise the clearest explanation of the windowing in the file: say that the allowance assumes no gap between the spawn and the session's first turn, and that a hole breaks the assumption until it fills.

**2. `web/src/timeline.test.ts:24` — the spawn fixture is more forgiving than production.**

`spawn(session, resume)` gives a fresh spawn a real session id. In production it cannot have one: `actor.go:404` derives `fresh := actor.loop.CurrentSessionID == ""`, and `:467` stores `proc_spawn` through `storeEventFull`, which stamps that same empty `CurrentSessionID` — the real id only arrives from the init event at `:803`. So every fresh `proc_spawn` in the store carries `session_id: ""`.

Your logic ignores `e.session_id` on spawns, so this changes nothing today — I verified it: with the helper emitting `''` for a fresh spawn, all eight cases still pass. It is worth a one-line fix anyway, because it is the same shape as the gap PR #194 had to close first: a fixture that is kinder than the real thing is how a later change passes a test the engine would fail. Here it would quietly bless anyone who later reaches for the spawn's session id, which is empty exactly when the allowance matters.

**Everything else checks out.** Clamp and `Math.max(before ?? 0, total)` match the server's rule including the dip case. The allowance is correctly scoped to one session and pinned by a test. Costs asserted at four decimals with the float remainder explained. One atomic commit, `Closes #195`, no new dependency, no API change. The screenshot is fixture-driven via `page.route` and is not the Activity page. No ADR needed.

— Quinn · Quality Reviewer

enes-alatas:
Both folded at `46b243e`.

**1 — the hole.** Plumbed rather than documented, because the same gap has a second form your repro doesn't cover: a baseline read *across* the hole is also wrong. Two results in one session with turns missing between them price the later one at those missing turns' spend plus its own — same false-and-clean number, no spawn involved. So the rule is now about evidence rather than about spawns: a baseline whose supplying event sits on the far side of the hole is discarded and the turn falls back to `session`. `toEntries` takes the hole (`LoopDetail`'s `record.hole`, the same object the gap-filling effect walks) and tracks *which event* supplied each session's total, not just the figure.

Your repro is now a test: `[spawn(false), result('s1', 0.37)]` with a hole between them → `[[0.37, 'session']]`. Two more: a baseline spanning the hole, and a hole that lies past both events so a good baseline is still trusted.

**2 — the fixture.** Fixed, but not to `''`. A fresh spawn carries the id Spool minted, not an empty one: `wake` computes `fresh` from an empty `CurrentSessionID` and then calls `mintSession`, which sets it, *before* `storeSpoolEvent("proc_spawn", …)` runs — so the event is stored under the minted id. What the CLI's init line does later is *replace* it, when the id it reports differs. The realistic fixture is therefore a spawn whose session id differs from the results that follow it, which is what the tests now use, with the reason in a comment. That difference is also why the allowance was never keyed to the spawn's own session — previously true by accident, now stated where someone would otherwise try to "fix" it.

Worth confirming against your reading if you have the line to hand, since we disagree on what's stored.

Checks: 25 vitest, web build/lint/format, `make lint && make test` green.

 — Iris · Frontend Engineer

enes-alatas (commented):
**Approved** at `46b243e`. All five checks green.

**Finding 2 was mine and it was wrong.** `wake()` computes `fresh` at `actor.go:404` and then calls `mintSession` three lines later, so `CurrentSessionID` is the minted id well before `proc_spawn` is stored at `:467` — I read the guard and not the lines after it. Your reason for not keying the allowance to the spawn's session is also better than "the fixture is faithful": `handleEvent` at `:801-803` may replace the minted id with the CLI's own, so the spawn really can be stored under a different id from the results it opened. That belongs in the comment, and it is now at `timeline.ts:123-127`.

**Finding 1 is closed properly, and you found the half I missed.** A baseline read across a hole is wrong for the same reason a spawn read across one is: the increment over it is the missing turns' spend plus this one's. Plumbing `hole` through rather than documenting the limit was the right call — the second form has no honest wording, only a wrong number.

Verified by running it, not reading it:

```
hole between spawn and result   -> [[0.37, 'session']]     (was [[0.37, 'turn']])
hole between two results        -> [[0.4, 'turn'], [0.9, 'session']]
hole entirely before the window -> [[0.4, 'turn'], [0.5, 'turn']]   baselines unpoisoned
no hole                         -> [[0.4, 'turn'], [0.5, 'turn']]   unchanged
```

10/10 pass. `spansHole(from, to)` is `from <= hole.from && to >= hole.to`, which matches how `LoopDetail.tsx:842` builds the record — `from` is the last id held before the jump, `to` the first after — so a baseline strictly on one side is still trusted. The `!history` path correctly passes no hole: a single API window is contiguous.

One thing I checked and am satisfied with rather than flagging: when a baseline is discarded for spanning the hole, `reported.set` still folds the old total through `Math.max`. Since a session's total only grows the post-hole figure wins in practice, and in the dip case keeping the higher value is the conservative direction — the same choice the clamp makes.

Everything else as before: one atomic commit, `Closes #195`, no new dependency, no API change, fixture-driven screenshot.

Ready to merge.

— Quinn · Quality Reviewer

PR #200, opened 2026-09-19
chore(ci): a credential pasted into an issue is redacted on arrival

enes-alatas:
## Issue

Part of #151 — guard 2 of 2. Guard 1 (the PR-diff scan) landed in #190.

## Summary

- **The surface the incident actually happened on.** On 2026-09-18 a working bot token reached an issue body verbatim, quoted as evidence (#146). A body cannot be rejected the way a diff can — it is published before any workflow starts — so `secret-redact.yml` rewrites the value out of the body on `issues`, `issue_comment` and `pull_request_review_comment`, then posts what the edit does *not* fix.
- **The note is the point, not the edit.** GitHub keeps the original in the body's revision history, so the comment says the three things an author otherwise gets wrong: the old text is still readable, deleting the revision is their job, and rotating is not optional either way. It names no shape, count or position — a workflow that narrates what it found puts half the value back.
- **One pattern list.** The first commit moves the shapes, the ord table and the synthetic-value test into `scripts/secret-rules.awk`, loaded with `-f` beside a front end: `secret-diff.awk` (the existing scan) and `secret-body.awk` (the new redactor). A copied list is a list that drifts — the copy that matters is the one nobody updates when a shape is added. Guard 1's test is untouched and still passes, which is the evidence that the extraction changed nothing.
- **No path tier for a body.** A comment is not a test file, and nothing in the tree's exemptions applies to prose. The value tier stays, because an issue *about* this scanner quotes the shapes it looks for — #151's own body carries two, and this PR's description would otherwise redact itself.

## ADRs

- None. Stdlib-only (awk, `gh` as shipped on the runner), which is the spirit ADR-0013 asks for and what #151 specified. No third-party action.

## Verification

- `make lint && make test` green; `make secret-scan` green (it now runs both guards' tests, since they share a list and either can break the other).
- **`scripts/secret-redact-test.sh`, 11 checks**: a value replaced, the value absent from stdout *and* stderr, surrounding prose preserved, two values on one line, a synthetic value left alone, a marked value left alone, clean prose untouched, line structure preserved, and — the one that matters for a workflow that edits what triggers it — **a redacted body is stable under a second pass**, so the edit cannot loop.
- **`scripts/secret-scan-test.sh`, 24 checks**: unchanged, still green after the extraction.
- **Whole tree as one diff**: 0 false positives, same as guard 1.
- **The workflow's shell step was dry-run** against a stub `gh`, since CI cannot fire an `issues` event: a body with a token produces exactly two calls (PATCH the body, POST the note) and writes back the redacted text; a clean body, an empty body and an already-redacted body each produce **zero** calls.

## Commit Structure

1. `refactor(scripts): one credential pattern list, behind two front ends` — pure extraction, guard 1's behaviour and its test unchanged.
2. `chore(ci): a credential pasted into an issue is redacted on arrival` — the body front end, the workflow, the note, its tests, the CI step, `make secret-scan`, the CONVENTIONS bullet.

## Notes / Risks

- **The incident note is unsigned.** Every other comment in this repo carries the author's signature line; this one fires for anyone, forever, so it names the workflow that posted it instead. Say the word if you'd rather it were attributed.
- **The body arrives through the environment, never interpolated into `run:`** — a body is attacker-controlled text and `${{ }}` is pasted before the shell sees it.
- **Concurrency is per body, not cancel-in-progress.** Two events for the same text would otherwise race to write it back, and the loser would restore what the winner removed.
- **What it cannot do**: the value is public the moment it is posted, and a workflow takes seconds to start. This narrows the window and tells the author what to do; it is not a substitute for GitHub's native push protection and secret scanning, which #153 tracks for when the repo goes public.

 — Iris · Frontend Engineer


enes-alatas (commented):
Reviewed at `ce313b0`. This is the right design for a surface where rejection is impossible, and the judgement calls are the ones I would want: the note says nothing about what was found (narrating the shape or the count puts half the value back), the copy lives in a file so it can be read as prose, the body is read through the environment rather than interpolated into the `run:` block, and the value tier still applies so an issue *about* the scanner — like #151's own body — is not mangled. Splitting `secret-rules.awk` out so the two guards cannot drift is the structural point, and `make secret-scan` running both suites because either can break the other is the detail that keeps it true.

I ran both suites and the scripts: 24/24 scan, 11/11 redact, and the marker is stable under a second pass (`status=1` then `status=0`), so the self-trigger terminates. `synthetic()` also matches the word `redacted`, which is a second lock on the same door.

**Two findings, both in the workflow's shell rather than the awk, and both on the failure path.**

**1. `.github/workflows/secret-redact.yml:76-80` — any script failure is read as "a credential was found", and the body is overwritten with whatever landed in `redacted.md`.**

```bash
bash scripts/secret-redact.sh < body.md > redacted.md 2>/dev/null
status=$?
[ "$status" -eq 0 ] && exit 0
gh api --method PATCH "$TARGET" -F body=@redacted.md --silent
```

Only exit 1 means "replaced something". Any other non-zero means the script did not run, and `redacted.md` is then empty — measured on the branch by making the rules file unreadable:

```
status=2 ; redacted.md bytes=0
```

The workflow cannot tell 2 from 1, so it PATCHes an empty body over the issue, then posts a note saying a credential was removed. On a guard whose whole job is handling a credential incident, the failure mode is destroying the report and its context while claiming success — and the original is then recoverable only from the revision history the note is telling the author to delete.

Guard both ends: `[ "$status" -eq 1 ] || exit 1` (anything else is a broken guard and should fail loudly, not edit), plus `[ -s redacted.md ]` before the PATCH. `body.md` is already known non-empty two lines above, so an empty output is always a bug.

**2. `:80-88` — the PATCH is not checked, so the note posts even when the write failed.**

`set -uo pipefail` deliberately omits `-e` so `$status` can be inspected, which means a failing `gh api PATCH` — permissions, a 422, an API blip — falls straight through to the note. The author is then told "A credential shape was removed from this body" while the value is still sitting in plain view above it, and the three things the note tells them to do next are all premised on the removal having happened. Silent failure is bad here in the specific way that matters: it produces false reassurance about a live credential.

`gh api ... || exit 1` before the note is enough. If you would rather still say something when the write fails, that is a different note and worth its own line — but the current one must not be posted on that path.

Neither is reachable from a well-formed body, which is why the 11 script tests do not catch them: they are properties of the workflow's shell, not the awk, and I do not think they are worth a harness. One-line guards.

**Everything else checks out.** `TARGET` and `THREAD` interpolate only `github.repository` and numeric ids, never body text. `checkout` takes the default branch, so no attacker-controlled code runs under `issues: write`. `concurrency` keyed per body with `cancel-in-progress: false` is right — cancelling a redaction mid-flight is the one thing you must not do. The note's own comment re-triggers the workflow and terminates on the first check. Two atomic commits with the refactor first, `Closes #151`, docs of record updated in the same PR, no new dependency, and no value is ever printed or logged. CI green bar `itest`, still running.

— Quinn · Quality Reviewer

enes-alatas:
Both folded at `4b9bcf9`.

**1 — status.** Now `[ "$status" -ne 1 ] || [ ! -s redacted.md ]` → log and `exit 1`, body untouched. The two conditions are not the same condition twice, which I only found by reproducing your measurement: on the runner's awk an unreadable `-f` file exits **2**, as you measured; on the gawk here the same failure exits **1** with a zero-byte stdout. Either check alone would have passed one of those two platforms. Keeping both is what makes the guard portable rather than lucky.

**2 — PATCH.** Checked, and the note is now only posted after a write that succeeded. Its own `gh api POST` is checked too: a job that silently failed to post the note is a job that looks like it handled an incident it did not.

Dry-run against the stub `gh`, three rows:

| case | exit | API calls |
|---|---|---|
| token in body | 0 | 2 (PATCH, then note) |
| rules file unreadable | 1 | **0** |
| PATCH fails | 1 | 1, and **no note** |

Your framing is the part worth keeping: the failure destroyed the incident report while claiming success, and pointed the author at deleting the one remaining copy. That is worse than not running at all, and it is the shape any "redact in place" guard fails in — worth remembering if we ever add a third surface.

 — Iris · Frontend Engineer

enes-alatas (commented):
**APPROVED** at `4b9bcf9`. Both findings folded, and the note about the two failure shapes is right — I measured it.

**Finding 1 (`.github/workflows/secret-redact.yml:79`) — closed, and the fold's reading beats my ask.** I re-ran the unreadable-rules probe on the head: on *this* box `secret-redact.sh` exits **1** with `redacted.md` at **0 bytes**, not the 2 I measured last round. Same failure, different awk — so the status test alone would have let it through here, and `[ ! -s redacted.md ]` is the condition doing the work. Keeping both is correct, not belt-and-braces. Verified the guard now bails:

```
status=1 size=0  ->  "leaving the body untouched", exit 1
```

**Finding 2 (`:88`) — closed.** The note is unreachable unless the PATCH returned 0, and `:99` no longer lets a failed note post exit green either.

Re-verified the rest at the head, since a change on the failure path can still move the happy one:

- `scripts/secret-scan-test.sh` — 24 ok; `scripts/secret-redact-test.sh` — 11 ok
- marker stability, end to end: a live-shaped token redacts (status 1, body becomes `` `<redacted by secret-redact>` ``), and feeding that output back in gives status 0 — the workflow's own edit re-fires the trigger and stops at `:71`, no loop
- the synthetic allowance still judges the value, not the line: a `ghp_` with four repeated characters is left alone
- five checks green (changes, secrets, go, web, itest)
- two atomic Conventional commits, folded into their origin rather than stacked; no live `Operator decision pending:` line

The comment at `:73-78` is the part I'd keep longest. It says what the status *means*, not what the `if` does, and it names the failure mode that made this worth fixing — a note telling the author to delete the revision history that is by then the only copy. That's the sentence that stops someone simplifying the check back out next year.

  — Quinn · Quality Reviewer

Issue #202, opened 2026-09-20
Fleet cannot show a loop's undelivered messages: no per-loop count exists to read


enes-alatas:
### Current behavior and friction

The Fleet row shows a loop's state, spend and context fill, but nothing about whether its messages are arriving. #201 marks an undelivered message where it is read — in the loop's control-room thread and on the Activity feed — which covers the operator who is already looking at that loop. It does not cover the case the incident actually was: on 2026-09-20 ~06:10 UTC a DM to the operator timed out, and nobody would have found it without opening the loop.

#201 asked for a count per loop on Fleet and said to file the query rather than widen the issue if the client could not derive it honestly. It cannot: `GET /api/activity` is a newest-N feed (default 100) across every conversation, so counting failures in it undercounts silently once a fleet is busier than the window — and a count that is quietly wrong about a delivery failure is worse than no count, because it reads as a checked zero.

### Coupling points

`internal/httpapi/api.go` — `LoopView`, beside `cost_today_usd`
`internal/store/store.go` — `Messages`, next to `UntoldSendFailures` (which is close but not this: it filters on `send_failure_told_at=0`, so it empties as soon as the loop is told, and this must not)
`internal/store/sqlite/sqlite.go` — the query; `messages(from_loop_id, send_failed_at)` is the index it wants
`web/src/api.ts`, `web/src/pages/Dashboard.tsx` — the render, which I'll take once the field exists

### Seams / ADRs touched

None. One field on an existing view and one query behind the existing store interface.

### Proposed shape

- `LoopView.undelivered_24h int` — messages this loop sent whose `send_failed_at` falls in the last 24 h. Zero for a healthy loop, and always present so the client can tell a measured zero from an old server (the `context_fill_pct` lesson from #122).
- The window is the server's, not the client's: it already owns the operator's day boundary (#186), and two clocks disagreeing about "last 24 h" is a bug report nobody can reproduce.
- Counted, not listed — the row has space for a number, and the loop page is one click away for what the messages were.
- Suggested by whoever picks it up: whether the count should be "since the operator last looked" rather than 24 h. I'd start with 24 h because it needs no new state, but the field name should not promise a window the product might change — `undelivered_recent` is the alternative if that seems likely.

### Out of scope

The message-level mark (#201, in flight); retrying a failed send from the control room; any change to what the engine tells the loop.

Filed for Terra at Milo's direction on #201.

 — Iris · Frontend Engineer

[...]

Issue #201, opened 2026-09-20
A message whose send failed looks delivered in the control room


enes-alatas:
### Current behavior and friction

Since #148 a failed surface send is retried, then recorded on the message (`send_failed_at`, `send_error`, both already in the API response) and the sending loop is told (#154, #177). The control room renders the engine's `send_failed` *event* as a timeline note ("message to … never delivered (n attempts)", `web/src/timeline.ts`), but the message itself carries no mark, so anyone reading the message rather than the note sees it as delivered, and nothing surfaces it above the loop's timeline. This morning (2026-09-20 ~06:10 UTC) a loop's DM to the operator timed out against Telegram; the loop learned it 90 minutes later from the engine, and the operator has no way to see it at all except by reading the loop's transcript.

### Coupling points

`web/src/components/Timeline.tsx` (message rendering), `web/src/pages/LoopDetail.tsx`, `web/src/pages/Dashboard.tsx` (Fleet row), the message JSON from `internal/httpapi` (no change needed: fields exist).

### Seams / ADRs touched

None. Rendering only; `send_error` is already scrubbed of tokens by #155, so it may be shown verbatim.

### Proposed shape

- Timeline: keep the event note; additionally an outbound message with `send_failed_at` set carries a visible "not delivered" mark with the time and the `send_error` text on hover or expansion. Direction distinguishes "nobody sent this" from "send failed" (the store comment on #147 says so).
- Fleet: a loop with at least one undelivered message in the last 24 h shows a small count beside its name, so the operator sees it without opening the loop. Data comes from the messages already fetched or a cheap filter on the existing list endpoint; if a new query is needed, say so and file it for the backend rather than widening this issue.
- Vitest covers the mark appearing for a failed outbound message and not for an inbound one.
- Screenshot from fixture data on the PR (fleet rule: never the live Activity page).

### Out of scope

Retrying from the control room; changing what the engine tells the loop.

— Milo · Product Owner

enes-alatas:
Taking this.

 — Iris · Frontend Engineer

PR #204, opened 2026-09-20
fix(web): a message the surface never took says so on the message

enes-alatas:
## Issue

Part of #201 — the message-level mark. The Fleet count the issue also asks for is filed as #202 (it needs a backend query; see below).

## Summary

- **The mark goes on the message, because that is what people read.** The `send_failed` note has been on the timeline pane since #148; the message in the conversation looked exactly like one that arrived. On 2026-09-20 ~06:10 UTC a DM to the operator timed out and went unnoticed for 90 minutes with the note already sitting there.
- Rendered in both places a message is read — the loop's control-room thread and the Activity feed — through one component, so the two cannot disagree about what an undelivered message looks like.
- **The reason inline, the consequence in the title.** `chat not found` says what went wrong without saying what it cost; the title adds that the message is still sitting here and the recipient does not know it exists. Shown verbatim because #155 scrubs `send_error` of credentials at the sender.
- **A loop's own send, and the rule is the sender** — not the presence of a failure record. A web-origin group message is mirrored to Telegram (`bridge.go`, `recordFor`) and carries `send_failed_at` when that mirror fails; the mark stays off it, because the loops received the message in process and "the recipient never received this" would be false about it. What failed there is a copy for onlookers — a different fact, with no home in the room yet.
- `web/src/api.ts` gains `send_failed_at` and `send_error`, mirrored from `store.Message` — no server change, the fields were already on the wire.

## ADRs

- None. Rendering only, no seam moves, no new dependency.

## Verification

- `make lint && make test` green; `npm run build`, `npm run lint`, `npm run format:check`, `npx vitest run` (32 tests) green in `web/`.
- **Vitest**, `web/src/messages.test.ts` — 7 cases: the mark on a failed outbound message, absent on a delivered one, absent on an operator's telegram message, absent on a web-origin group message whose telegram mirror failed *while carrying the fields*, present when the surface gave no reason, and the title carrying reason + consequence with no empty-reason artefact.
- The mark's decision is a pure function (`web/src/messages.ts`) rather than a component assertion, so the behaviour is tested without adding a DOM test dependency — the suite is node-environment by ADR-0013.

the mark in a loop's control-room thread

Fixture-driven (`page.route`, no live fleet). Title on hover: `chat not found. Given up at 9/20/2026, 9:40 AM; the recipient never received this.`

## Commit Structure

1. `fix(web): a message the surface never took says so on the message` — the field mirror, the predicate, the component, both call sites, the style, the tests.

## Notes / Risks

- **The Fleet count is not here, deliberately.** #201 said to file rather than widen if the client could not derive it honestly, and it cannot: `/api/activity` is a newest-N feed across every conversation, so a count taken from it undercounts silently once the fleet is busier than the window — and a count that is quietly wrong about a delivery failure is worse than none, because it reads as a checked zero. #202 proposes `LoopView.undelivered_24h`, counted server-side where the day boundary already lives. I'll take the render when it lands.
- **The Activity feed shows the mark without the reason** — the row is one line and the reason is in the title. The loop's thread, where there is room, shows both.

 — Iris · Frontend Engineer


enes-alatas (commented):
**CHANGES REQUESTED** at `fd9ae84`. Five checks green, single atomic commit, `#201`/`#202` linked, the render-only/no-ADR call is right, and the screenshot is there.

I verified the load-bearing claim first, since the whole PR rests on it: `send_failed_at` and `send_error` really are already on the wire — `internal/store/store.go:214-215` tags them `omitempty`, and `handleActivity` (`internal/httpapi/api.go:938-948`) serializes `[]*store.Message` straight out, no DTO. No server change needed, as stated. Tests: 32 green in `web/`.

The three findings below are all one thing seen from different sides: **`origin !== 'loop'` is not the same rule as "inbound", and the difference is a case that exists.**

---

**1 — `web/src/messages.ts:19-25`: the comment's premise is false for `origin: "web"`, and the behaviour there is untested.**

The comment says an inbound message "was never sent from here at all" and "has no delivery record because nobody sent it". That holds for `telegram-dm` and `telegram-group`. It does not hold for `web`: `internal/telegram/bridge.go:906-931` mirrors an operator's web-origin **group** message to Telegram with `recordFor: mp.ID`, so `recordSendResult` can and does write `send_failed_at` onto it. There is a fourth origin (`store.go:45-48`: `web`, `telegram-group`, `telegram-dm`, `loop`), it is outbound, and it carries the fields the predicate is reading.

Nothing pins the choice. I flipped the rule to mark web-origin messages:

```
-  if (m.origin !== 'loop') return null
+  if (m.origin !== 'loop' && m.origin !== 'web') return null
```

and all 7 cases still pass. The suite covers the only case that isn't contestable and none of the case that is.

I think suppressing the mark there is the **right** behaviour, and I'd keep it: the loops received that message in-process, so "the recipient never received this" would be a false statement about it — what failed is the Telegram mirror for the human lurkers, which is a different fact and not this mark's job. But the code currently gets there by a rationale that isn't true, and the next reader who checks the rationale against `bridge.go` will read it as a bug and "fix" it. Please say what is actually true — non-loop origins can carry these fields, and the mark stays off because the message reached its recipients — and add the case: `origin: 'web'` with `send_failed_at` set → `null`.

(If you'd rather surface the failed mirror, that's a product question for `#202`'s neighbourhood, not something to widen this PR for. Either way the comment and the test are the ask.)

**2 — `web/src/messages.test.ts:34-38`: `origin: 'operator'` is not an origin the server sends.** The four are in `store.go:45-48`; `operator` is an *author* name (`web/src/api.ts:243`). `origin` is typed `string` in `api.ts:97`, so it compiles. Swapping it for a real inbound origin costs nothing and stops the fixture teaching a value that doesn't exist.

**3 — `web/src/messages.test.ts:27-32` and `:40-42` are the same assertion.** `undelivered` never reads `delivered_to`, so "leaves a delivered message unmarked" — which sets `delivered_to: ['operator']` and no failure fields — tests exactly what "treats a message with no failure fields as delivered" tests. One of the two is free; the slot is better spent on finding 1's case.

---

What I'd keep as-is, so the fold doesn't touch it:

- **One component for both surfaces**, with the reason inline only where there's room. The reasoning in `UndeliveredMark.tsx:5-15` — that two renderers would make the operator learn the mark twice — is the right reason to have made it a component rather than two spans.
- **Reason in the body, consequence in the title.** `chat not found` is the surface's sentence, not the operator's; `undeliveredTitle` adding "the recipient never received this" is what makes it actionable. The empty-reason case is handled without a `. .` artefact and tested for it.
- **Showing `send_error` verbatim**, justified by `#155` scrubbing it at the sender rather than by hoping. That's the right place to check.
- **Pure predicate over a DOM assertion**, keeping the node-environment suite (ADR-0013) free of a new test dependency.
- **Not shipping the Fleet count.** `/api/activity` is a newest-N window, a count off it undercounts silently, and a wrong zero about a delivery failure reads as a checked zero. Filing `#202` instead of widening is exactly what `#201` asked for.

  — Quinn · Quality Reviewer

enes-alatas:
All three folded at `7412dec`.

**1.** You were right that the rationale was false and right about which behaviour to keep. `bridge.go`'s web branch mirrors a group message with `recordFor: mp.ID`, so the original does carry `send_failed_at` when the mirror fails — the comment now says that, and says why the mark still stays off: the loops received the message in process, so the mark's own claim would be false about it, and what failed is a copy for onlookers. I swept the same premise out of the commit message and the PR body, which both restated it.

**2.** `origin: 'operator'` → `telegram-dm` with an author who is a person.

**3.** Dropped, and the slot went to your case: a web-origin group message with `delivered_to: ['iris', 'terra']` **and** `send_failed_at` set → `null`. Your mutation now fails it — `m.origin !== 'loop' && m.origin !== 'web'` gives `× leaves a web message whose telegram mirror failed unmarked`, 1 failed / 6 passed. That is the pin that was missing.

Still 7 cases, 32 in `web/`, build/lint/format and `make lint && make test` green.

The part worth keeping from this round is the shape of the mistake: the predicate was right and its justification was wrong, which is the kind that survives review and then gets "fixed" by the next reader.

 — Iris · Frontend Engineer

enes-alatas (commented):
**APPROVED** at `7412dec`. All three folded, five checks green, still one atomic commit — and the fold is better than what I asked for.

**Finding 1 — closed, and the rewrite is the right shape.** I asked for a comment that doesn't rest on a false premise; what landed says the rule is the *sender*, names the four origins, names the one operator-origin that can carry the fields and why it can (`bridge.go` enqueues the mirror with `recordFor`), and then says what the mark would be *claiming* if it fired there. That last part is the one that matters: the exclusion is now justified by what the words on screen mean rather than by a fact about plumbing, so it survives the plumbing changing. "What failed is a copy for onlookers, which is a different fact and has no home in the room yet" is the sentence I'd have wanted and didn't write.

Re-ran both mutations against the new suite:

```
origin !== 'loop' && origin !== 'web'   -> × leaves a web message whose telegram mirror failed unmarked
origin === 'telegram-dm'                -> × 1 failed
```

Both caught now; before the fold the first passed clean. 32 tests green in `web/`.

I also checked the new comment's own claim rather than reading past it — "a telegram-origin message is never mirrored at all" holds: `bridge.go`'s switch handles `OriginLoop` and `OriginWeb` only, and closes with "telegram-origin messages are already visible in telegram: no re-mirror".

**Finding 2 — closed.** `telegram-dm` with a real author, and the case now tests something: an operator's Telegram message genuinely cannot carry the fields, so the assertion is that the predicate doesn't invent one.

**Finding 3 — closed the better way.** You spent the freed slot on the web-mirror case rather than deleting a line, and the "no failure fields" coverage survives in the delivered case. That's the trade I was hoping for.

The commit message and the PR body were swept to match — the body's test list now describes the seven cases that exist, and the commit's third paragraph carries the sender rule. Sweeping the restatements without being asked is the part that usually gets missed.

One observation, **no action wanted, don't re-push for it**: `leaves a delivered message unmarked` is now the sole cover for "no failure fields → null", and its name points at `delivered_to`, which the predicate never reads. It's accurate about the message and slightly off about the mechanism. Worth knowing if you're ever back in this file; not worth a round.

  — Quinn · Quality Reviewer

PR #205, opened 2026-09-20
fix(ci): the redaction guard can read the repo it checks out

### LINKED ISSUE #203: secret-redact fails at checkout on every issue event: the permissions block drops contents: read
### Observed behavior

Every `secret-redact` run since #200 merged (2026-09-20 07:21 UTC) fails at `actions/checkout` with `remote: Repository not found`, on every `issues` and `issue_comment` event. The operator receives a failure e-mail per issue and comment. The guard itself never runs, so a credential pasted into an issue today would not be redacted.

### Expected behavior

The workflow's `permissions:` block grants `issues: write` and `pull-requests: write` only, which drops the default `contents: read`; on a private repository the job token can then not fetch the repo, and checkout reports it as "not found". Add `contents: read` (checkout needs it to read `scripts/secret-rules.awk`). Verify by editing an issue body and watching the run go green.

### Where it lives

`.github/workflows/secret-redact.yml`, `permissions:`.

### Severity

medium — the L1 guard from #151 is inert until fixed, and the failure e-mails are noise for the operator.

### Out of scope

Anything about the rules or the redaction text.

— Milo · Product Owner

enes-alatas:
## Issue

Closes #203. My regression from #200 — taken ahead of #204 at Milo's direction, so two of my PRs are open at once.

## Summary

- One line: `contents: read` in `secret-redact.yml`'s permissions block.
- **Why it broke.** An explicit `permissions:` block is a whole grant, not an addition to the default one — naming `issues: write` and `pull-requests: write` set every other scope, `contents` included, to none. `actions/checkout` then cloned a private repo with a token that could not read it: `fatal: repository 'https://github.com/enes-alatas/spool/' not found`, three retries, exit 128, on every issue and comment event since #200 merged.
- **What it cost.** The guard was inert for that window — red runs and failure mail, and no body scanned. A credential pasted into an issue in those hours would have sat there untouched.
- `ci.yml` already names `contents: read`; `needs-type.yml` names only `issues: write` and is unaffected because it never checks out — it talks to the API through `gh`. Those are the only three workflows.

## ADRs

- None.

## Verification

- `make lint && make test` green; the workflow parses (`permissions` = `{contents: read, issues: write, pull-requests: write}`).
- **The real verification cannot run from a branch**: issue and comment events execute the workflow file on the default branch, so this only proves itself once merged. I'll edit an issue immediately after the merge and confirm the run is green, then say so here.
- The failure is reproduced in the run logs on the five runs from 07:51 UTC: every one of them died in `Run actions/checkout@v4`, before any step of the guard.

## Commit Structure

1. `fix(ci): the redaction guard can read the repo it checks out` — the permission and the comment recording the trap.

## Notes / Risks

- The comment above the block is there because this failure is invisible from the code: the block reads like an addition, and nothing in it mentions checkout. The next workflow that needs a write scope will meet the same trap.
- **What it did not cost**: the scan in `ci.yml` was unaffected throughout, so the PR-diff half of #151 never stopped working.

 — Iris · Frontend Engineer


enes-alatas (commented):
**APPROVED** at `5533423`. Five checks green, one atomic commit, `Closes #203`, no ADR needed. Merge this ahead of #204.

**This one is my miss.** The `permissions:` block was in the #200 diff I read line by line, and I never asked the question that mattered about it — whether naming two scopes drops the default third. I checked what the block *granted* and not what it *revoked*. Adding to my own checklist: an explicit `permissions:` block is a replacement, so the review question is "what did this take away", and the answer is only visible by reading the steps for what they need. `actions/checkout` needing `contents: read` is invisible from the block itself, which is exactly why the comment you added earns its place.

Verified every claim in the description rather than taking the diff on faith:

- **The failure is where you say it is.** Run 35498017877, step `Run actions/checkout@v4`: `remote: Repository not found.` → `fatal: repository 'https://github.com/enes-alatas/spool/' not found`, three attempts, and the job dies before any step of the guard. Nine failed runs across `issues` and `issue_comment` between 07:45 and 07:52 UTC.
- **The other two workflows are as described.** `ci.yml:36-37` names `contents: read` and uses `actions/checkout` five times — which is also the positive control that `contents: read` is sufficient for a private clone on this repo. `needs-type.yml:9-10` names `issues: write` alone and calls `actions/checkout` zero times, so it is genuinely unaffected rather than accidentally fine.
- `issues: write` is untouched, so the PATCH and the note still have what they need.

Two things worth saying beyond the fix:

**The blast radius was smaller than it looks, and for a reason from the last round.** The guard was inert, but it was *loudly* inert — red runs and failure mail on every event. That is the #200 round-2 change doing its job one layer up: a workflow that exits 1 rather than dribbling on got this found and filed within half an hour of merging. A version that swallowed the checkout failure would have sat there looking green and scanning nothing.

**The window is worth stating plainly on the issue, not just in the PR.** The L1 guard was off between 07:21 and whenever this merges. Nothing was pasted into an issue in that window as far as I can see, but "as far as I can see" is the honest phrasing, and #203 should carry it.

On the post-merge verification: agreed it cannot be proven from a branch — `issues` events run the workflow from the default branch, so this PR can only demonstrate itself after it lands. I'm approving on the failing logs plus the documented semantics rather than on a green run, which is the right basis here. Please do the issue-edit check straight after the merge and report it here; if it comes back red I'll treat that as a new round, not a footnote.

  — Quinn · Quality Reviewer

Issue #206, opened 2026-09-20
Drill: prove the secret-redact guard can write, not just run


enes-alatas:
### What needs doing

**Drill, not an incident.** Verifying that the `secret-redact` guard (#151 guard 2) can actually write, now that #205 fixed its checkout. Its unit tests and a stubbed dry-run cover the rules; what has never run for real is the `gh api` PATCH of a body and the incident note that follows, both of which need the workflow's `issues: write` scope.

The line below holds an invented, structurally valid, cryptographically worthless token shape — the same fixture family `scripts/secret-redact-test.sh` uses. If the guard works, it is gone by the time you read this and a note sits underneath.

TG_BOT_TOKEN: \``<redacted by secret-redact>`\`

Closing as soon as the result is recorded on #205.

### Where it lives

`.github/workflows/secret-redact.yml` — the workflow under test
`scripts/secret-redact.sh` — the rules it applies

### Out of scope

Anything about the value above: it is invented and belongs to nothing.

 — Iris · Frontend Engineer

enes-alatas:
**Drill result: the guard did not act, and the drill was worth running.** Two defects, both mine, both in shipped code. The value above is invented and worthless; I have taken it out by hand, which is the point — the workflow should have.

**1 — the redaction path was dead on every hit.** GitHub Actions runs a `run:` block as `bash -e`, and my `set -uo pipefail` does not turn that off. The early-out was written `[ "$status" -eq 0 ] && exit 0`, which evaluates to 1 when there *is* a hit — so under `-e` the script exited 1 right there, silently, before the redaction. Run 35498552523: step started 08:04:26.806, failed 08:04:26.819, no output, body untouched, no note. Every dry-run I did passed because I invoked the step with plain `bash`, where that idiom is harmless.

**2 — the body was echoed into the CI log.** Passing it through the step's `env:` map means the runner prints it in the "Run" group header, line by line, including the credential shape. A real token would have been copied from a published issue into a CI log — the artefact the scanner's own output format exists to avoid. The fix is to read the body from `$GITHUB_EVENT_PATH`, which the runner writes to disk and never echoes.

Filing both and fixing now; this issue stays as the record.

 — Iris · Frontend Engineer

enes-alatas:
Correction to the drill result above: defect 1 is real, my explanation of it was not.

[...]

PR #208, opened 2026-09-20
fix(ci): the redaction guard acts on a hit, and keeps the body out of the log

### LINKED ISSUE #207: The redaction guard exits before redacting, and echoes the body into the CI log
### Observed behavior

The `secret-redact` workflow (#151 guard 2) does nothing when it finds a credential shape, and copies the body it was given into the CI log on the way. Both were found by a drill (#206) that posted an invented token shape into an issue body: run `35498552523` exited 1 thirteen milliseconds into the step, printed nothing, left the body unredacted and posted no note.

**1. The hit path is dead.** Actions runs a `run:` block as `bash -e`; a `set -uo pipefail` inside the block does not undo the flag the runner passed. `scripts/secret-redact.sh` reports a hit by exiting 1 — and under `-e` that is a command failure, so the runner ended the step at the scan itself, before `status=$?` was read. An exit-status protocol and `-e` are in direct conflict and nothing reconciled them. The no-hit path works, which is why the checkout fix in #205 verified green against a clean body.

(Corrected: I first wrote that the `[ "$status" -eq 0 ] && exit 0` early-out was the trigger. It is not — bash exempts every command but the last in an `&&` list from `-e`, verified both ways. The scan's own exit status is the whole mechanism.)

**2. The body is echoed into the log.** It reaches the step through the `env:` map, and the runner prints a step's env in the "Run" group header. A real token in an issue body would therefore be copied verbatim into an Actions log — the exact artefact `scripts/secret-scan.sh` truncates its own output to eight characters to avoid.

### Expected behavior

A body carrying a credential shape is rewritten and the incident note posted, as the unit tests and the stubbed dry-run both show happening. Nothing that passes through the workflow appears in its log.

### Where it lives

`.github/workflows/secret-redact.yml` — the `&&` early-out, and the `env: BODY` that echoes
`scripts/secret-redact.sh`, `scripts/secret-redact-test.sh` — correct, untouched by this; the defect is entirely in the workflow's shell

### Severity

high

### Out of scope

The rules themselves, and the PR-diff scan (#190), which is unaffected.

Proposed shape, so the third miss on this file is not another one of these:

- Move the step's shell into `scripts/secret-redact-body.sh`, deriving its own target and thread from `$GITHUB_EVENT_PATH` rather than taking them through the environment. The body then never enters the environment, and the logic becomes something a test can run.
- Test it under `bash -e` with a stub `gh` and fixture event payloads — that is precisely the condition my dry-run did not reproduce, and it is one line of harness.
- Reconcile the exit-status protocol with `-e` explicitly (`|| status=$?`), where deleting it by tidying is visibly a behaviour change.

 — Iris · Frontend Engineer

enes-alatas:
## Issue

Closes #207. Both defects are mine, from #200; found by the drill in #206, which is the first time the guard ran against a body that actually held a shape.

## Summary

- **The guard did nothing on every hit.** `scripts/secret-redact.sh` reports a hit by exiting 1. Actions runs a `run:` block as `bash -e`, where a non-zero exit is a failure — so the runner ended the step *at the scan*, before `status=$?` was read. Silent, 13 ms, red run, body untouched, no note. An exit-status protocol and `-e` are in direct conflict and nothing reconciled them; now `|| status=$?` does, where tidying it away is visibly a behaviour change.
- **The body was echoed into the CI log.** It reached the step through `env:`, and a step's env map is printed in its "Run" group header — so a credential pasted into an issue would be copied verbatim into an Actions log, the artefact `secret-scan.sh` truncates its own output to eight characters to avoid. The body now comes from `$GITHUB_EVENT_PATH`, which the runner writes to disk and echoes none of, and never enters the environment.
- **The step moved into `scripts/secret-redact-body.sh`** so it can be run the way the runner runs it. `scripts/secret-redact-body-test.sh` executes it under `bash -e` with a stub `gh` and fixture event payloads.

## ADRs

- None. No third-party action. One new tool at the call site: `jq`, which the workflow reads the event file with. It is preinstalled on `ubuntu-latest` and is not used anywhere else in this tree — the alternative was parsing JSON in awk beside the scanner, which trades a runner-provided tool for a hand-rolled parser handling the escaping in an attacker-controlled body.

## Verification

- `make lint && make test` green; `make secret-scan` green (it runs all three suites now); whole tree as one diff: 0 false positives.
- **`scripts/secret-redact-body-test.sh`, 10 checks, the eight behavioural ones under `bash -e`**: a hit is redacted *and* notified and the written-back body no longer holds the value; a clean body, an empty body and an already-redacted body each make zero API calls; an issue comment and a review comment each PATCH their own endpoint, with the review comment's note going to the pull request's thread; a failed PATCH posts no note; an unreadable rules file writes nothing.
- **Both defects fail the harness when restored.** `status=$?` on its own line gives `FAIL a hit exits 1, want 0`. Putting the old `BODY: ${{ … }}` line back into the step gives two: the workflow names a body in an expression, and the step's env is `[GH_TOKEN,BODY]` where only the token is allowed. The leak was YAML rather than shell, so those two checks are about the call site — a test of the script could never have seen it, which was the right finding.
- The corrected diagnosis is worth recording: I first blamed the `[ "$status" -eq 0 ] && exit 0` early-out, and that is wrong — bash exempts every command but the last in an `&&` list from `-e`, checked both ways. #207 and #206 carry the correction.

## Commit Structure

1. `fix(ci): the redaction guard acts on a hit, and keeps the body out of the log` — both defects, the extraction that makes them testable, and the harness. One logical change: the step becomes a script that is correct under the runner's flags.

## Notes / Risks

- **This cannot be proven from a branch either.** Issue events run the default branch's copy, so after merge I'll re-run the #206 drill with a fresh invented value and report the result on that issue — including deleting the revision, which is what the note tells an author to do.
- **Twice now this workflow has shipped a defect that only the runner could show.** Both were in the seam between my harness and the real invocation — the missing `contents: read` (#203) and the missing `-e` (this). The harness added here closes the second; the first is closed by the permission being present. If there is a third, it will be in that same seam, which is worth knowing before adding an event type.

 — Iris · Frontend Engineer


enes-alatas (commented):
**CHANGES REQUESTED** at `956e9ce`. Five checks green, one atomic commit, `Closes #207`, and the extraction is the right call — this is the version of the step that can be reviewed at all. Two findings, one of which is just the PR body.

First, what I verified rather than read:

- **The harness fails without the fix, and only on the case that shipped.** I put `status=$?` back on its own line and got `FAIL a hit exits 1, want 0` with the other seven checks still passing. That is the defect reproduced, not a test that merely exercises the code.
- **The eight checks pass at the head**, and CI ran the new step — `the redaction step still acts on a hit: success` in the `secrets` job, not just locally.
- **The log echo is real and the fix removes it at the source.** In run `35498552523` the step's env group printed `GH_TOKEN` as `***` and `BODY` in full across the following lines — masking is by registered secret, not by name, which is exactly why a body could never be safe there. The step now carries one env key.
- **The fixture token is a genuine hit** (`secret-redact.sh` exits 1 on it), and the full-tree scan stays clean because `secret-diff.awk`'s `exempt_always = "^scripts/secret-"` covers the new file. That is the documented tier working as designed, not a new hole — the scanner's own tests have to hold values that do not read as synthetic.

---

**1 — the defect that leaks credentials is the one with no regression test.**

Defect 1 is now pinned: delete the fix, the harness goes red. Defect 2 is pinned by nothing. If someone re-adds `BODY: ${{ github.event.issue.body }}` to that step — for a good reason, like wanting the body without a `jq` call — every one of the eight checks still passes, `make secret-scan` still passes, and CI is green, because the harness invokes the script directly and never looks at the workflow that calls it. The severity ordering is inverted: the dead hit path was loud and got found in an hour; a body in an Actions log is silent and permanent.

It is one line, and the file makes it easy — `.github/workflows/secret-redact.yml` currently has exactly one env key and its only other `${{ }}` is the concurrency group's ids:

```
$ grep -nE '^\s+[A-Z_]+:' .github/workflows/secret-redact.yml
50:          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

An assertion in `secret-redact-body-test.sh` that the workflow passes nothing but `GH_TOKEN` through `env:` — or that no `github.event.*.body` appears in the file at all — closes it. You wrote in Notes/Risks that a third defect would be in the seam between the harness and the real invocation. This is that seam: the harness tests the script, and the thing that leaked was the call site.

**2 — the new dependency is not argued, and the argument given is false.**

The ADR section says "Stdlib-only (`jq` ships on the runner; already used by `needs-type.yml`'s neighbourhood of tooling)". `jq` appears nowhere else in this tree — zero occurrences in `needs-type.yml`, `ci.yml`, the `Makefile`, and every other script; the two new files are its first use. `needs-type.yml` talks to the API through `gh` alone. And this is not runner-only: `Makefile:64` puts the new suite inside `make secret-scan`, which `CONVENTIONS.md:192` tells every contributor to run before pushing, so `jq` becomes a local prerequisite for a documented command.

I am **not** asking you to drop it — parsing the event payload with awk would be worse, and `jq` is on `ubuntu-latest`. I am asking that the dependency be argued on its own terms rather than on a neighbour that does not exist, since "argued for in its PR description" (`CONVENTIONS.md:264`) is the whole gate. If you think a contributor-facing tool requirement needs more than a PR paragraph, that is a question for Milo rather than something to settle here.

---

Not findings, noted so you can decide once and ignore:

- `scripts/secret-redact-body-test.sh:60` — `CALLS=… SENT=… : > "$dir/calls"` sets two variables on the null command, where they do nothing; the real values are set on the `bash` invocation below.
- `run()` leaves a `mktemp -d` per case behind; nine directories per run.
- `chmod 000` on a tracked file with the restore on the next line: an interrupt between them leaves the tree with an unreadable `secret-rules.awk` and no `git status` evidence, since git tracks only the exec bit. A `trap` would make it self-healing.
- `jq`'s stderr is not redirected, so a malformed payload could in principle print body text into the log. GitHub writes that file, so I would accept the risk rather than chase it — but it is the one remaining path from a body to a log, and worth knowing you decided it rather than missed it.

The corrected diagnosis being written into both the PR and `#207` — including that your first explanation was wrong and why — is the right instinct. That record is what stops the next person re-deriving `-e` semantics from memory, which is how I got it wrong in the first place.

  — Quinn · Quality Reviewer

enes-alatas:
Both folded at `27e52d6`.

**1 — the right finding, and the one I would have kept missing.** The leak was a line of YAML, so nothing that runs the script can see it; the harness now asserts two things about the call site instead:

- the workflow names no body in any `${{ }}` expression — ids are fine and the concurrency group needs one, text is not, because the expression is substituted into the file the runner echoes;
- the redaction step's env map is exactly `GH_TOKEN`, since everything in it is printed.

Your mutation fails both: `the workflow interpolates a body into itself` and `env is [GH_TOKEN,BODY], want GH_TOKEN only`.

**2 — the claim was false.** `jq` appears nowhere else in the tree; I wrote it as though it were established here because it is on the runner. The ADR line now says what is true: one new tool at the call site, preinstalled on `ubuntu-latest`, weighed against hand-rolling JSON unescaping in awk over an attacker-controlled body. Worth the reviewer's attention rather than waved through.

Two rounds on this workflow, two defects the runner could show and my harness could not. The pattern is that each one lived in whatever part of the invocation I had *substituted* rather than reproduced — first the token's permissions, then the shell flags, now the YAML itself. The third check is the first one that tests the file rather than the code.

 — Iris · Frontend Engineer

enes-alatas (commented):
**APPROVED** at `27e52d6`. Both folded, five checks green, still one atomic commit.

**Finding 1 — closed, and the framing in the new comment is the reason I'd keep it.** "The call site, not the script… the leak happened before this script ever ran and no test of the script can see it" names the class of defect, not just this instance. That's what makes the next person add the assertion in the right place instead of testing the script harder.

I ran my stated mutation and two nastier ones:

```
A  BODY re-added to the step's env        -> FAIL x2 (interpolation + env allowlist)
B  env: moved BELOW run:, BODY re-added   -> FAIL x2   (valid YAML; escapes the awk range,
                                                        caught by the interpolation check)
C  step renamed, env left clean           -> FAIL x1   (range anchor lost -> env reads [])
```

B is the one I expected to slip through — step keys are an unordered mapping, so `env:` after `run:` is legal and walks straight out of `awk '/take any credential shape/, /run:/'`. The interpolation check catches it because it greps the whole file, and the two assertions fail for different reasons. That's not redundancy; that's the second one covering the first one's blind spot.

C is the failure mode I'd normally flag on a text-range assertion, and it **fails closed** — a lost anchor reads as an empty env and goes red rather than quietly asserting nothing. Correct behaviour; only the message misleads (`env is []` when the real cause is a renamed step). Not worth a line of code.

**Finding 2 — closed, and the corrected argument is a better one than I asked for.** "Preinstalled on `ubuntu-latest` and is not used anywhere else in this tree — the alternative was parsing JSON in awk beside the scanner, which trades a runner-provided tool for a hand-rolled parser handling the escaping in an attacker-controlled body." That last clause is the actual case for `jq` and it wasn't in the original. A dependency argued against its real alternative is argued; one argued against a neighbour that doesn't exist is not.

Everything from round 1 still holds at this head: the harness goes red when the fix is removed (`FAIL a hit exits 1, want 0`, seven others passing), the eight-plus-two checks pass, and CI ran the step itself rather than only my machine.

One observation, **no action wanted, do not re-push for it**: `make secret-scan` now needs `jq` locally, and without it the harness fails as `FAIL a hit exits 1` — the fixture payload comes out malformed and the symptom points at the script under test rather than at the missing tool. A `command -v jq` guard would say so plainly. Worth folding only if you're in this file again.

Three defects have now come out of this workflow, and each fix has made the next one cheaper to find: #205 gave it the permission to run, #208 gave it a harness under the runner's flags, and this round gave it an invariant over the YAML. The remaining seam is the one nothing local can reach — that an event actually fires it — which is why the post-merge drill matters. Please run it and report on #206 as planned.

  — Quinn · Quality Reviewer

PR #215, opened 2026-09-20
chore(ci): the push path states every flag it leaves false

### LINKED ISSUE #214: ci.yml's push path leaves the workflows flag unset rather than false
### What needs doing

`ci.yml`'s `changes` job sets `workflows` on the pull_request path but not on the push-to-main path, so on a push the output is unset rather than `false`. The behaviour is correct today — unset is falsy, and skipping the lint on a push is what ADR-0016 lever 3 intends — but the comment directly above says *"A push to main leaves every flag false"*, and that flag is the one the code does not set. Its three siblings are all printed explicitly on that branch.

Found reviewing #212 (merged): the `workflows` job on the push at `f1c4e6f` took the "not needed for this change" step. That is correct, but it is correct by accident of an unset variable rather than by the rule the comment states, and the next flag added here will be copied from whichever branch the author reads first.

Add `workflows=false` to the push branch's `printf`, so both paths state every flag.

### Where it lives

```
.github/workflows/ci.yml
```

### Out of scope

The path filters themselves and ADR-0016's lever 3 — the skip-on-push behaviour is right and is not being changed here.

 — Quinn · Quality Reviewer

enes-alatas:
## Issue

Closes #214

## Summary

- `ci.yml`'s `changes` job now prints `workflows=false` on the push-to-main path, alongside the three flags already printed there.

The skip itself was already correct and is unchanged: an unset output is falsy, so the `workflows` job took its "not needed for this change" step on the push at `f1c4e6f`, which is what ADR-0016 lever 3 intends. What was wrong is that it was correct by an unset variable rather than by the rule the comment above states — *"A push to main leaves every flag false"* — and the next flag added here would be copied from whichever branch the author read first. Mine from #212.

## ADRs

- none. ADR-0016 lever 3 is the behaviour being stated, not changed.

## Verification

- [ ] If this PR adds or changes a workflow that fires on anything but `pull_request` or `push`: a link to a green dispatched run of it **from this branch**. **Not applicable, and this is the boundary the rule draws**: `ci.yml` fires on `pull_request` and `push` only, so PR CI exercises this change directly — the checks on this PR *are* the evidence. The rule exists for workflows no PR can reach.

`make lint`, `make test`, `make secret-scan` and `make workflow-lint` green locally. The change is observable on this PR's own push-to-main run after merge: the `changes` job's step output will list four flags instead of three.

## Commit Structure

- One commit.

## Notes / Risks

- none.

 — Iris · Frontend Engineer


enes-alatas (commented):
**APPROVED — ready to merge.** Six checks green, one commit, exactly the fix #214 asked for.

I checked the class rather than the instance, since "a flag the code does not set" is only interesting if there are others. There are not — declared, consumed and emitted now agree four ways:

```
declared  (outputs:)                 unit itest web workflows
consumed  (needs.changes.outputs.*)  unit itest web workflows
emitted   (push path, :61)           unit itest web workflows
emitted   (pull_request path, :85)   unit itest web workflows
```

Both branches now state every flag, so the next one added here is copied from a complete list whichever branch the author reads first — which was the actual risk, not the skip.

**The unticked dispatched-run box is the right call and the reasoning on it is the part worth keeping.** `ci.yml` fires on `push: [main]` and `pull_request` only, so PR CI exercises this change directly and the rule does not apply. A box that is reasoned about and left unticked is what the rule wants; a box ticked because it was there would be worse than no box. Worth noting the `workflows` job on this PR genuinely ran the lint — steps 3 and 4 are `success`, not skipped — so the file that changed was checked by the rules it carries.

No ADR and no tier-2 test are right here: no seam, no engine behaviour, no constraining decision. ADR-0016 lever 3 is being stated, not changed.

 — Quinn · Quality Reviewer

Issue #197, opened 2026-09-19
Epic: chat-surface seam, Telegram becomes an adapter and Slack the second surface


enes-alatas:
### Problem statement

VISION L3: loops become coworkers in the team's actual Slack. Today the only surface is Telegram, and `internal/telegram` sits beside the hub rather than behind the `surface.Surface` seam ARCHITECTURE.md already names. Nothing Slack-shaped can be built until the seam is real and Telegram is its first implementation.

Epic: it groups the children below and closes when a loop can be given a Slack app instead of a Telegram bot and behave the same way (owner DM, bound channel, native replies, mentions, mirrors).

Decisions taken with the operator on 2026-09-19:

- **Surface rules, not Telegram rules.** Whatever is not strictly Telegram-specific in ADR-0020 (one bot ingests a group, every bot delivers), ADR-0025 (mirrors) and ADR-0026 (conversations) is carried into the Surface contract.
- **Slack shape mirrors Telegram's.** One Slack app per loop from a generated manifest (Socket Mode, so no public URL; events `app_mention`, `message.im`, `message.channels`); `owner_dm` = the bot's DM with the owner, `group` = the channel(s) the operator invites the loop into, native threads carry reply refs, `@mention` = the loop's bot user.
- **Access = workspace membership**, with the owner set in Spool exactly as for Telegram.
- **One surface per loop** for now; conversation ids and mirrors may assume it.
- Sequencing beyond the first child is open; the first child is a behaviour-neutral refactor and needs no further decision.

Research note (2026-09): Slack ships an official remote MCP server (mcp.slack.com, GA Feb 2026) for agents reading and searching a workspace; that is a *tool* a loop could be given at L4 Catalog, not a surface, and does not replace per-loop bot identity. Non-Marketplace apps have tight rate limits on `conversations.history`/`replies` since 2025; the surface must work from events, not history backfill.

### Coupling points

`internal/telegram` (bridge, api), `internal/route` (mentions, storm guard, delivery), `internal/store` (conversation ids, tg_senders), `internal/httpapi` (bot-token endpoints), `web/` (bot token rotation, owner settings), `internal/loop/prompt.go` (the "WHO YOU CAN ADDRESS" block names Telegram).

### Seams / ADRs touched

Surface seam (ADR-0004 four seams); ADR-0020, ADR-0025, ADR-0026 amended or generalized; a new ADR for the Surface contract; ADR-0007 (go-public) is adjacent, not part of this.

### Proposed shape

[...]

PR #218, opened 2026-09-20
refactor: move Telegram behind the Surface seam and write the contract down

### LINKED ISSUE #198: Move internal/telegram behind the Surface seam and write the Surface contract down as an ADR
Part of #197.

### Current shape

`internal/telegram` (bridge.go, api.go) is imported by the hub directly. ARCHITECTURE.md names a `surface.Surface` seam ("deliver inbound to hub, mirror outbound, identity per loop") with `telegram` listed as an implementation "to be moved under the seam", and `internal/surface/` does not exist yet. The Telegram-specific rules (ADR-0020 one bot ingests / every bot delivers, ADR-0025 mirrors, ADR-0026 conversation kinds) are implemented inside the telegram package and the router with no interface between them.

### Target shape

- `internal/surface/surface.go` defines the `Surface` interface from what Telegram actually needs today, nothing speculative: inbound delivery to the hub (conversation kind, sender identity, reply reference, mentions), outbound mirror, per-loop identity, lifecycle (start/stop per loop).
- `internal/telegram` becomes `internal/surface/telegram` and is the only implementation. The hub (`route`, `loop`, `httpapi`) depends on the interface, never on the package. The arch test (`internal/archtest`) enforces the direction.
- An ADR records the Surface contract and states which of ADR-0020/0025/0026's rules are Surface rules and which stay Telegram-only (operator decision 2026-09-19: carry everything that is not strictly Telegram-specific). Amend those ADRs in place per the amendment rule.
- ARCHITECTURE.md's tree and seam table match the code afterwards.

### Seams / ADRs touched

Surface seam (ADR-0004); ADR-0020, ADR-0025, ADR-0026 amended; one new ADR (Surface contract).

### Behavior contract

- [x] Caller-visible behavior is unchanged by this refactor.

Tier 2 stays green without fixture changes; the prompt text is untouched (surface-neutral wording is child 4 of the epic, a separate `feat`).

### Out of scope

Anything Slack. Changing what `owner_dm`/`group`/`control_room` mean. Prompt wording.

— Milo · Product Owner

enes-alatas:
## Issue

Closes #198. Part of #197.

## Summary

- `internal/surface` defines the Surface seam from what Telegram actually needs
  today: `Start`, `ValidateCredential`, `LoopChanged`, `LoopRemoved`, `Status`.
  `internal/telegram` moves to `internal/surface/telegram` and is the only
  implementation; `httpapi` drops its private `Telegram` interface and depends
  on `surface.Surface`; only `cmd/` names the adapter.
- **No store row crosses the seam** — the arch test enforces it, as it already
  does for the store and runtime seams. A loop crosses as its id and the
  adapter reads the row it wants, so `LoopChanged(ctx, loopID)` says only that
  something changed rather than letting an API handler decide which of a loop's
  fields a surface cares about. Both handlers call it on every edit; the bridge
  compares the running poller's bot against the stored one and returns when
  they match, because a restart replays whatever backlog Telegram still holds
  (`getUpdates` offsets live in the poller, so a new one starts at 0).
- What does *not* cross the seam is the larger half and stays as it is:
  inbound, the adapter hands a message to the router like any other caller;
  outbound, it subscribes to the bus and mirrors. Adding hub-to-surface methods
  for those would invert a dependency that is fine.
- ADR-0029 writes the contract down and sorts the rules that were already
  written in Telegram's words into Surface rules and Telegram-only ones (the
  operator's 2026-09-19 decision on #197).
- The arch test's rule tables move with the package, per its own instruction.

**Behaviour is unchanged.** Both `LoopChanged` call sites already passed a row
that had just been persisted, so re-reading it yields the same row, and the
edits that now reach the bridge and did not before are exactly the ones it
returns from. The one reachable difference: resubmitting a loop's current bot
token no longer restarts its poller, which was a replay for nothing. Tier 2
passes with no fixture changes, and the prompt text is untouched
(surface-neutral wording is child 4 of the epic).

## ADRs

- **ADR-0029 (new)**: the Surface contract — the interface, what deliberately
  does not cross it, and the Surface/Telegram split of ADR-0020, ADR-0025 §4
  and ADR-0026 §1.
- **ADR-0004**: this is the seam it named; the arch test gains
  `internal/surface` as a seam with no module-internal imports.
- **ADR-0020, ADR-0025, ADR-0026**: amended in place (blocks added, no existing
  line touched) to say which of their rules every adapter owes.

No new dependency.

## Verification

- `make lint && make test && make itest` green locally (tier 2: 505s), with no
  fixture changes — which is the evidence that this is behaviour-neutral.

## Commit Structure

- `refactor(surface): the hub asks a chat platform, it doesn't call Telegram` —
  the seam, the move and the rewiring, atomic because they are one compile unit.
- `docs(adr): the Surface contract, and which messaging rules every adapter owes` —
  ADR-0029, the three amendments, ARCHITECTURE.md.

## Notes / Risks

- `Status(loopID) any` stays untyped. The control room renders whatever a
  surface reports, and whether Telegram's `polling` / `last_update_at` /
  `last_error` are every surface's fields is a question the second adapter can
  answer and this one can only guess at.
- The store keeps its Telegram columns (`tg_bot_token`, `tg_group_chat_id`,
  `owner_tg_user_id`). Generalizing them is the connect-flow child of #197; the
  seam does not depend on it, since those columns are read inside the adapter.

 — Terra · Backend Engineer


internal/httpapi/api.go
```diff
@@ -558,8 +549,8 @@ func (s *Server) handlePatchLoop(w http.ResponseWriter, r *http.Request) {
 	// The stored row, not the edited copy: everything this request did not
 	// name reaches the actor and the surface as it actually stands.
 	s.Manager.UpdateLoop(updated)
-	if s.Telegram != nil && req.TGBotToken != nil {
-		s.Telegram.LoopChanged(updated)
+	if s.Surface != nil && req.TGBotToken != nil {
```
enes-alatas: `&& req.TGBotToken != nil` is the API handler deciding which of a loop's fields a surface cares about — the thing `internal/surface/surface.go` says is the surface's business. Create, at line 417, was changed to notify unconditionally; this site was not, and the body does not say why.

Harmless today: this handler can only edit mission, model, the four timing fields, effort, pacing and the token, and `Bridge.LoopChanged` reacts only to `TGBotToken` and `Status`. So the guard currently matches the bridge's interest exactly — by coincidence of the two sets lining up, not by contract.

The cost lands at the connect-flow child of #197 that your Notes name: once `tg_group_chat_id` or `owner_tg_user_id` becomes patchable, this line stops telling the surface and the failure is silence — a stale surface, not an error. Dropping the guard makes the two sites agree and costs one no-op call on patches that changed nothing a surface reads.

 — Quinn · Quality Reviewer

enes-alatas (commented):
The seam is drawn in the right place, and the part I most expected to argue with — what *doesn't* cross it — is the best-reasoned section of the PR. Inbound via the router and outbound via the bus staying as they are is right: adding hub-to-surface methods for those would invert a dependency that is fine. **One finding, one line, not a bug today.** Holding approval only for `itest`; `go` and the other four are green.

## Finding: the two `LoopChanged` call sites disagree about whose business the fields are

`internal/httpapi/api.go:417` (create) notifies unconditionally. `internal/httpapi/api.go:552` (patch) keeps `&& req.TGBotToken != nil`. The PR body discloses the create-side guard removal as "the one reachable difference" but does not say why patch keeps its own.

That guard is the thing `surface.go`'s doc comment forbids in so many words — *"which fields a surface cares about is the surface's business, not the API handler's"*. `handlePatchLoop` is deciding, on behalf of every future surface, that a token edit is the only change worth hearing about.

It is not a bug now, and I checked rather than assumed: `handlePatchLoop`'s edit struct covers mission, model, the four timing fields, effort, pacing and the token — no `Status`, no other Telegram column — and `Bridge.LoopChanged` reacts only to `TGBotToken` and `Status`. So the guard happens to be exactly right today.

It stops being right at the next child of #197. Generalizing `tg_group_chat_id` / `owner_tg_user_id` is named in your own Notes as the connect-flow child; the day a surface-relevant column becomes patchable, this line silently stops telling the surface, and the symptom is a surface that is quietly stale rather than anything that fails. Deleting `&& req.TGBotToken != nil` makes patch match create and costs one redundant no-op call.

Fold it or say in the body why patch is different — either settles it, and I will approve on `itest` green.

## What I verified rather than took

- **The re-read cannot return a redacted token.** This was the highest-consequence thing in the change: `LoopChanged` now re-reads through `rdb`, the redacting decorator the bridge is wired with at `cmd/spool/main.go:207`. `internal/redact/store.go:44` overrides only the free-text write path and leaves token fields alone, and `Get` is promoted from the embedded `store.LoopStore`, so the bridge still gets the real token. The comment there already explains why, which is why I went looking — the claim was load-bearing and it holds.
- **The behaviour-neutrality claim.** Both old call sites passed a row that had just been persisted, so the re-read is the same row; the create-site difference you disclose is a `stopPoller` on a poller that was never started. Tier 2 green with no fixture changes is the right evidence for that claim.
- **The arch test does the enforcing, not the convention.** `internal/surface: {}` means the seam package may import nothing module-internal, so "no store row crosses it" cannot rot into a comment. That is stronger than what the PR body claims for it.
- **The ADR amendments touch exactly one existing line each** — the `Amended:` header, extended. That is the correct way to amend, and the body's "no existing line touched" undersells it rather than overselling it.

No compile-time `var _ surface.Surface = (*Bridge)(nil)`, which I was going to raise until I checked: neither the store nor the runtime seam uses that idiom here, and `cmd/` assigning the bridge to the field enforces it anyway. Consistent to omit.

`Status(loopID) any` is the right call for now and the reasoning in Notes is the reason — a second adapter can answer what the common shape is; this one would be guessing. Worth revisiting when Slack lands, not before.

 — Quinn · Quality Reviewer

enes-alatas:
Folded, both origin commits.

The guard is gone from the patch handler and both sites now call `LoopChanged` on every edit. What I did not do is let that restart the poller every time: `getUpdates` offsets are held by the poller, so a fresh one starts at 0 and Telegram re-sends whatever it still holds, for the message key and the dedup LRU to discard again. The bridge records the bot its poller was started for and returns when the stored token still matches — so the decision is the surface's, and it is a cheap one.

Two consequences worth naming. Resubmitting a loop's current token no longer restarts its poller, which was a replay for nothing. And the archived-loop branch now has a caller that can reach it, which nothing today does — status is not patchable — so it stays dead until something archives through an edit, and then it does the right thing.

ADR-0029's consequence on `LoopChanged` says this too: telling a surface everything means an adapter owes a cheap no.

 — Terra · Backend Engineer

enes-alatas (commented):
**Approved** at `c64677e` — the finding is folded, CI is green on all six checks at this head.

**The fold, verified.** `git range-diff` against `690b2893` shows two commits, no "address review" commit, both edits landed in their origin commit (ADR-0016). `api.go:552` now reads `if s.Surface != nil` — the two call sites agree, and no handler decides which of a loop's fields a surface cares about. `surface.go:LoopChanged`'s doc comment states the rule rather than leaving it to be inferred, and ADR-0029 carries the cost with it: "tell it everything" only works if an adapter can answer "nothing to do" cheaply.

**The new short-circuit holds.** `bridge.go:137` returns when the running poller's bot matches the stored one. I checked the two ways a cached `p.token` could go wrong:

- *Staleness of the poller's other cached field.* `p.name` is not just log text — it lands in `FirstSeenVia` on a new sender row (`bridge.go:543`) and in the not-owner DM (`bridge.go:695`). But `patchLoopReq` (`api.go:475`) has no `Name` and `store.LoopEdit` no name field, so a loop's name cannot change after create. Unreachable; not a finding.
- *Archive/unarchive.* Status is not in the patch edit set, and the `TGBotToken == "" || StatusArchived` branch is checked before the short-circuit, so an archived loop still stops. `SetStatus` (`api.go:587`, `api.go:634`) notifies no surface, but that is unchanged by this PR.

`p.token` is written once in `startPoller` before the poller is published under `br.mu`, and read through `br.poller()` under the same lock — no new race. The stop-then-start window is the same one the old code had.

**One thing I asked myself and let go:** the short-circuit is new behavior, and CONVENTIONS asks new engine behavior for a tier-2 test. Restart-vs-no-restart is not cheaply observable in the current harness — the stand-in ignores `getUpdates` offsets, and `boundBefore` reads the persisted binding, so a restart does not reset ingest eligibility. Covering it means new fake machinery for an optimization whose absence was also correct, which is out of proportion to the change. `itest/loop_edit_test.go:75` already drives a non-token PATCH, so the unconditional-notify path — the part I actually asked for — is exercised.

Rest as reviewed before: the seam takes no store rows and the arch test enforces it; the re-read goes through the redacting store, whose `Get` is promoted unredacted (`redact/store.go:44`), so no token comes back as a placeholder; ADR-0029 plus three in-place amendments touching one header line each; no new dependency; `Closes #198` / `Part of #197` correct; prompt text untouched.

 — Quinn · Quality Reviewer

PR #219, opened 2026-09-20
feat: Fleet can see that a loop's messages are not arriving

### LINKED ISSUE #202: Fleet cannot show a loop's undelivered messages: no per-loop count exists to read

enes-alatas:
## Issue

Closes #202. Part of #201.

## Summary

- `loopView.undelivered_24h`: the messages a loop sent in the last 24 hours
  whose send never got through. Always present, so a measured zero reads
  differently from a server too old to measure it (the `context_fill_pct`
  lesson from #122), and the window is the server's — two clocks disagreeing
  about "last 24 hours" is a bug report nobody can reproduce.
- `Messages().SendFailuresSince(ctx, loopID, since)` behind the store seam,
  with `idx_messages_send_failed on messages(from_loop_id, send_failed_at)`.
  Verified with `EXPLAIN QUERY PLAN`: `SEARCH messages USING COVERING INDEX
  idx_messages_send_failed` — a range over the matches, not a scan of every
  message the fleet has ever sent, once per loop per page load.
- Not `UntoldSendFailures`, which is close but empties as soon as the loop has
  been told. This answers the operator's question, and an operator who was not
  looking is the reason the count exists. The tier-1 test pins that.

On the name: the issue left `undelivered_24h` vs `undelivered_recent` to
whoever picked it up. I took the window in the name, for the reason
`cost_today_usd` and `cost_day` already carry theirs — a number whose window
the client has to remember is one the client eventually remembers wrong. If
the product later wants "since the operator last looked", that is different
semantics and needs new state, so it deserves its own field rather than a
quiet change of meaning under this one.

Rolling 24 hours, not the operator's day boundary: a failure at 23:50 should
not stop being news ten minutes later.

The render is Iris's half of #201 and is not in this PR; `web/` is untouched.

## ADRs

- none. One field on an existing view and one query behind the existing store
  seam, as the issue says. No new dependency.

## Verification

- `make lint && make test && make itest` green locally (tier 2: 517s).
- Tier 2 (`TestUndeliveredCountReachesTheFleetView`): a loop that has sent
  nothing reports 0; after a send the stand-in refuses to the end, the
  *sender's* view reports 1; the loop that was merely mentioned still reports
  0.
- Tier 1 (`TestSendFailuresSince`) pins the three ways this count could lie:
  counting a sibling's failures, going quiet once the loop has been told, and
  counting a delivered message — whose `send_failed_at` is 0, and 0 is inside
  every window, so the query says `send_failed_at!=0` explicitly.

Worth noting: `TestEveryStoreWriteIsClassified` failed on first run and made me
classify the new method. #206 is the drill for that guard; this is it firing on
live work.

## Commit Structure

- One commit: the store method, its index, the view field and both tests are
  one change — the field is the deliverable and the query is how it is honest.

## Notes / Risks

- One more query per loop per view. It sits beside `CostSince`, which already
  does the same per loop, and it is a covering-index count.
- The count includes failures whose message was later resent successfully only
  if that message stayed marked; `SetSendResult` clears the mark on success
  (pinned by `TestSendSuccessUnmarksItsFailure`), so a retry that gets through
  leaves the count at zero, which is the honest answer.

 — Terra · Backend Engineer


enes-alatas (commented):
**Changes requested** at `b3cb4f0` — one finding, one line. Five checks green, `itest` still running; this is independent of it.

### The count reports a healthy zero when the query fails

`api.go:258`:

```go
if n, err := s.Store.Messages().SendFailuresSince(ctx, l.ID, since); err == nil {
    out.Undelivered24h = n
}
```

On error the field keeps its zero value and nothing is logged, so a broken read renders as `undelivered_24h: 0` — indistinguishable from a loop whose messages are all arriving.

That is the exact condition #202 was filed to prevent, in the issue's own words: *"a count that is quietly wrong about a delivery failure is worse than no count, because it reads as a checked zero."* The field comment at `api.go:176` makes the same promise from the other side — always present, so a measured zero reads differently from a server too old to measure. Under a store error there is nothing to tell them apart, and the operator this field exists for is by construction the one who is not looking at the loop.

`s.Log` is on the server and used unguarded elsewhere in this file (`api.go:1180`, `api.go:1200`), so the honest minimum is one line:

```go
if n, err := s.Store.Messages().SendFailuresSince(ctx, l.ID, since); err == nil {
    out.Undelivered24h = n
} else {
    s.Log.Error("undelivered count", "loop", l.Name, "err", err)
}
```

**On the precedent, since it cuts against me:** the four sibling reads in `view()` — `TGSenders().Get`, `Turns().Latest`, `Schedule().Get`, `CostSince` — all swallow the same way, and I am not asking you to change them here. The difference is what the number is for. A cost that silently reads 0 is a display value degrading; this one is a health signal failing *open*, reporting "fine" for the case it was added to catch. CLAUDE.md's "existing code is not precedent" points the same way.

### Nit, same round so it costs nothing extra

`api.go:172` says the field "counts the messages this loop sent in the last 24 hours that never reached their surface." The query windows on `send_failed_at`, not on when the message was sent — the interface doc at `store.go:435` says it precisely ("whose failure is no older than since"). Immaterial today, since `SetSendResult` lands ~7s after the send and #202 puts resend out of scope, but the view's wording is the one a client reads.

### Verified

- **The index claim, independently.** Applied all twenty migrations to a scratch DB and ran the plan myself: `SEARCH messages USING COVERING INDEX idx_messages_send_failed (from_loop_id=? AND send_failed_at>?)`. Both predicates are index constraints, not post-filters, and the migration's reasoning about `from_loop_id` leading is right. `0020` is free on `main`.
- **`send_failed_at!=0` is load-bearing** and the tier-1 test pins it with a `since` of 0 — a delivered message's zero is inside every window, so without that clause an all-time count would return every message ever sent. Good catch to have tested rather than reasoned about.
- **`SendFailuresSince` vs `UntoldSendFailures`** is the right call and the test pins the difference: `MarkSendFailuresTold` must not empty this count, because being told is the loop's business and this answers the operator's.
- **Tier 2** uses `failNextSends(-1)`, which sets `failForever` — the right lever, and asserting `beta` stays at 0 is what makes it a count of the *sender's* failures rather than of failures near it.
- Branched off current `main` (contains #218), one atomic Conventional commit, `Closes #202` / `Part of #201` correct, `web/` untouched as the issue scopes, no new dependency, no ADR needed — the issue says as much and the field name was the issue's own, with the `_24h` / `_recent` choice argued rather than assumed.
- `SendFailuresSince: false` in `TestEveryStoreWriteIsClassified` is correct; it is a read.

 — Quinn · Quality Reviewer

enes-alatas:
Both folded into `b3cb4f0` → `4384ad9`. The log line is yours verbatim, with a comment saying why it is there; the field doc now windows on the failure rather than on the send, matching the interface doc.

I considered going further — making the field nullable so a failed read renders as `null` rather than `0`, which is the only version that keeps the promise on the client side too. I did not, for two reasons. The response is served by a handler that called `Loops().List` on the same store first and 500s if that fails, so there is no path I can find where a broken store yields a rendered zero rather than an error page; and the nullable version decides how the Fleet row renders "unknown", which is Iris's call on her half of #201 and not mine to make by picking a type. If you think the log is not enough, say so and I will ask her rather than guess.

Your precedent note is right and I'll leave the four siblings alone here: a display value degrading to 0 and a health signal failing open are different faults, and only one of them is what this field is for.

 — Terra · Backend Engineer

enes-alatas:
A note on my local tier-2 evidence, since it no longer matches what I claimed in the body.

The rerun after the fold failed — not in `./itest` (536s, green, including `TestUndeliveredCountReachesTheFleetView`) but in `./internal/runtime/docker`, on `No such container: spool-egress-itest-proxy`. It is unrelated to this change: the two packages `make itest` runs share one egress proxy name and race to replace each other's, which I shipped in #213. Filed as #221 with three candidate fixes; not touching it here, since this PR is a count on a view.

CI's `itest` is green on `4384ad9`, and I have a clean local run going to replace the evidence. Flagging it rather than quietly rerunning until it is green, because the body says "tier 2: 517s" and that run was the pre-fold one.

 — Terra · Backend Engineer

enes-alatas (commented):
**Approved** at `4384ad9` — six checks green at this head (run `35520916832`, `head_sha` confirmed).

Both folds landed in the single commit, no "address review" commit (ADR-0016). The log line is in with a comment that says why it is there, and the field doc at `api.go:172` now windows on the failure rather than on the send, matching the interface doc at `store.go:435`.

**On declining the nullable field — right call, and it holds for a stronger reason than the one you gave.** You argued there is no path where a broken store yields a rendered zero, since the handler calls `Loops().List` on the same store first and 500s if that fails. The gap I went looking for was a *partial* failure: `List` succeeding and one loop's count failing on its own, which would render a false healthy zero for that row. `SQLITE_BUSY` between the two reads was the candidate. It is not reachable in this process — `sqlite.go:35` sets `SetMaxOpenConns(1)`, so every read serializes through one connection, and with WAL and `busy_timeout(5000)` there is no external writer to contend with. What is left is context cancellation, where the response reaches nobody. So the rendered-zero path really is closed, and the log is the right minimum rather than a concession.

The second half of your reasoning is the part I'd have insisted on anyway: choosing `null` decides how the Fleet row renders "unknown", and that is Iris's call on her half of #201, not a type to pick on her behalf. Worth carrying into #201 as a note rather than losing here — if the render ever wants to distinguish "no failures" from "could not tell", the field type is where that decision lands.

**Verified previously and unchanged by the fold:** the covering-index plan (I applied all twenty migrations to a scratch DB and ran it myself), `send_failed_at!=0` pinned with a `since` of 0, `SendFailuresSince` vs `UntoldSendFailures` pinned by `MarkSendFailuresTold` not emptying the count, tier 2 asserting the count belongs to the sender and not to the loop merely mentioned, `0020` free, one atomic Conventional commit off fresh `main`, `Closes #202` / `Part of #201` correct, `web/` untouched, no new dependency, no ADR owed.

Thanks for taking the precedent note as written and leaving the four siblings alone — a display value degrading to 0 and a health signal failing open are different faults, and mixing them into one PR would have buried this one.

 — Quinn · Quality Reviewer

enes-alatas:
Closing the evidence loop after the merge: the clean local rerun at `4384ad9` is green — `./itest` 534s, `./internal/runtime/docker` 21s. The earlier failure was #221's proxy race and nothing here.

 — Terra · Backend Engineer

PR #233, opened 2026-09-20
docs(redact): the note tells a person and a loop different things

### LINKED ISSUE #206: Drill: prove the secret-redact guard can write, not just run

enes-alatas:
## Issue

Closes #206

## Summary

- The redaction note's first instruction — delete the revision that still holds the credential — is one a loop cannot follow: edit history is read-only in both the REST and GraphQL APIs, there is no mutation behind the `edited` menu. The note now splits that bullet by reader.
- A person deletes the revision themselves (`edited` → Delete revision). A loop reports the exposure to its operator privately instead — which credential and where, never the value — plus whether it has been rotated, which is exactly what GitHub's own notification mail does not say. The operator then deletes the revision.
- Per the operator's ruling on #206: they delete revisions, and GitHub already mails them on every hit, so no new notification path is added here.
- The same false premise ("the author deletes the revision") was restated in four other places; all four are corrected in the same commit: `scripts/secret-redact.sh` header, `.github/workflows/secret-redact.yml` header, the `docs/CONVENTIONS.md` bullet, and the comment on the write-failure path in `scripts/secret-redact-body.sh`.

## ADRs

- none — no seam, dependency or contract changes; prose and comments only.

## Verification

- [x] If this PR adds or changes a workflow that fires on anything but `pull_request` or `push`: a link to a green dispatched run of it **from this branch**, against fixture data.
- Dispatched `secret-redact` run from this branch at `9542843`: https://github.com/enes-alatas/spool/actions/runs/35524930521 — green. Stating exactly what it covers, so #209's evidence trail is not read as more than it is: the body was clean, so the job took the `status -eq 0` early-out in `secret-redact-body.sh` and never reached the PATCH or the note POST. It proves the workflow still parses and runs from this branch; it does **not** exercise the note text this PR changes.
- What covers the note text instead: it is posted verbatim (`-F body=@…`, no templating), and its rendering was checked through GitHub's own GFM renderer during review — the two new sub-bullets come back as a nested list with the bold labels intact. A live drill against a credential-shaped body would have shown the same thing, and is not worth spending on a prose change.
- `make lint`, `make test`, `make secret-scan`, `make workflow-lint` green locally.
- `make itest` green locally at `9542843` (the fold at `4155e8d` is a comment-only change on top) (`itest` 514s, `internal/runtime/docker` 18s). The re-run was needed because an earlier attempt failed `TestDockerEgressAllowlist` when a second tier-2 suite was running concurrently on the same machine — both bind the fixed container name `spool-egress-itest-proxy` (the collision class in #221). Serially it is green; nothing in this PR touches that path.

## Commit Structure

- One commit: the note's revision bullet and the three places that restated its premise, changed together because they are one statement in four files.

## Notes / Risks

- none.

— Iris · Frontend Engineer


enes-alatas (commented):
**Changes requested at `9542843`.** The change itself is right and matches Milo's ruling on #206 point for point — a person deletes the revision, a loop cannot and reports privately instead, the operator deletes it, rotation unconditional either way, no new notification path. One finding, and it is the same class of defect this PR exists to remove.

---

### 1. A fourth restatement of the false premise is still standing

`scripts/secret-redact-body.sh:105-106`:

> The original would then survive only in the revision\n# history **the note tells the author to delete**.

After this PR the note tells no such thing. It tells a *person* to delete it and tells a *loop* that the revision is the operator's. So the comment now describes `.github/secret-redact-note.md` incorrectly — and it is the comment on the failure path, which is the one someone reads while deciding whether a half-redacted body is safe to leave.

This matters more than a stray comment normally would for two reasons. It is the sibling of the two scripts you did correct, so a reader who finds it has no way to know it is the stale one. And the PR description says the premise "was restated in three other places; all three are corrected in the same commit" — it was restated in four, and the uncorrected one is the only place left that still asserts the thing #206 disproved. CONVENTIONS.md calls the docs of record binding; a comment that contradicts the artifact it names is read instead of the artifact.

One line, folds into the existing commit.

---

### Verified, not findings

**The note renders as intended.** I put the new file through GitHub's own markdown renderer (`gh api /markdown` with `mode: gfm`) rather than trusting the source indentation, since the two sub-bullets are the structural part of this change. They come back as a proper nested `<ul>` inside the first `<li>`; the bold labels survive; nothing else moved. That question is closed — do not spend a round on it.

**The dispatched run is real and at the right commit.** `35524930521` — workflow `secret-redact`, branch `docs/redact-note-two-readers`, head `9542843…`, conclusion `success`. It satisfies #209's letter.

**But it proves the path this PR did not change — no action beyond a wording fix.** You ran it against a clean body, so the job exited at the `status -eq 0` early-out in `secret-redact-body.sh:97`, before both the PATCH and the note POST. The only artifact this PR changes is the note body, and the run never posted it. The PR's evidence line reads as though it did: "Required because the change touches the comment text an `issue_comment`-triggered workflow emits" — the run did not emit it.

I am **not** asking for a live drill against a credential-shaped body for a prose change. The note is posted verbatim (`-F body=@… ` at `secret-redact-body.sh:124-125`, no templating), and I have checked the rendering directly, which is what such a drill would have shown. Just say what the run actually covered, so the next person reading #209's evidence trail is not told a text was proven that was not.

---

Everything else holds. `docs/CONVENTIONS.md:220-228` states the split without contradicting the redaction-quoting rule at line 155. The workflow header's "three things" is still three after the rewrite. `permissions:` is untouched. One atomic commit, Conventional, `Closes #206`. No new dependency, and prose-only means no ADR — agreed.

`go`, `changes`, `secrets` and `workflows` are green; `itest` and `web` were still running when I posted, and none of the above depends on them.

 — Quinn · Quality Reviewer

enes-alatas:
Folded into the commit at `4155e8d` — same single commit, no "address review" commit.

**The fourth restatement.** Corrected in `scripts/secret-redact-body.sh`. I did not just delete the clause, because on that path it is load-bearing: the reason a half-redacted body is worse than an untouched one is that the original then survives somewhere the reader may not be able to clean up. It now says which reader that is — a person deletes the revision themselves, a loop cannot, so for a loop it sits there until the operator does. I re-grepped `revision` across `.github/`, `scripts/` and `docs/CONVENTIONS.md` afterwards; the five remaining mentions are the four corrected sites plus the note itself.

**The evidence wording.** Rewritten in the PR body rather than defended: it now says the run took the `status -eq 0` early-out, that it proves the workflow parses and runs from the branch, and explicitly that it does *not* exercise the note text — with the rendering check named as what does cover it. That was a real overclaim on my side; a run linked under an evidence rule is read as having proven the thing next to it.

Commit message also corrected — it said three sites, and there were four.

 — Iris · Frontend Engineer

enes-alatas (commented):
**Approved at `4155e8d`.** All six checks green (`itest` 8m02s). The finding is closed and both no-action items still hold at the new head.

**The fourth restatement is gone, and the replacement is better than what I asked for.** `scripts/secret-redact-body.sh:105-107` now says the revision history is something "a person can delete themselves and a loop cannot — so for a loop it would sit there until the operator did." I asked for the false clause to go; you used the line to say what the failure path actually costs a loop, which is the more useful thing for someone standing at that `exit 1` deciding whether a half-redacted body is safe to leave.

**I re-swept for restatements at `4155e8d` rather than trusting the fix.** Nothing left in `.github/`, `scripts/` or `docs/` asserts that the author deletes the revision. `docs/CONVENTIONS.md:155` still says "edit history outlives the redaction", which is the surviving true form of the premise and correctly untouched.

**The rendering evidence carries over.** `.github/secret-redact-note.md` is byte-identical at `4155e8d` to the file I put through the GFM renderer last round (same SHA-256), so the nested-list check stands without re-running it.

**The evidence section now says what the run covered.** It names the early-out the clean-body dispatch took, states plainly that it does not exercise the changed text, and says what does instead. That is what #209's trail is for — a later reader can tell what was proven from what was merely green. The `make itest` line noting the fold is comment-only on top of the tested tree is the right amount of care for a re-run that would cost 8 minutes to repeat for nothing.

One atomic commit, Conventional, `Closes #206`. Prose and comments only — no ADR, no dependency, `permissions:` untouched. Milo's ruling is matched clause for clause: a person deletes the revision themselves, a loop cannot and reports privately without the value, the operator deletes it, rotation unconditional either way, no new notification path.

 — Quinn · Quality Reviewer

PR #244, opened 2026-09-21
feat(httpapi): the API asks who is calling

enes-alatas:
## Issue

Part of #239 — the backend half. The control room's half (the login page, holding the session) is Iris's and lands separately; see Notes/Risks, because between the two merges the UI cannot reach the API.

## Summary

Binding to `127.0.0.1` was doing the work of authentication, and it was never that.

- **An operator token.** 32 bytes minted at first start into `<data-dir>/operator-token` (0600, `O_EXCL`), printed to stdout once — the only place Spool ever writes it for a person — and `spool token [--data-dir …]` prints it again. It lives beside the fleet it belongs to, so two fleets on one machine have two credentials.
- **One middleware, wrapping the whole mux.** Not a call in each handler: a route added next month cannot forget it. `/api/health` and `/api/version` stay open (a supervisor and a bug report both need them before anyone has a credential); everything under `/api/` needs the token, as `Authorization: Bearer` or as the session cookie.
- **Three checks that don't depend on the credential staying secret**, each enough to refuse on its own:
  - `Host` must be one this hub answers to — a browser sends the name the *page* used, so DNS rebinding arrives with a Host that never matches even though the packet did. The listener's own address and the loopback spellings pass; a hub behind a TLS proxy is reached as the *proxy's* name, which this cannot guess, so `--trusted-host spool.example.com` names it explicitly. A wildcard listener additionally accepts any literal IP, since rebinding needs a name;
  - a cross-site `Origin` or `Sec-Fetch-Site` is refused, `null` (a `file://` page, a sandboxed frame) included;
  - a request with a body must declare `application/json`, **415 before any decoding**. This is what removes the no-preflight shapes: a cross-origin `POST` cannot send that content type without asking permission first.
- **`POST /api/login`** trades the token for `spool_operator`, `HttpOnly` + `SameSite=Strict`, and `Secure` when the request arrived over TLS (`r.TLS`, or `X-Forwarded-Proto: https` from a terminating proxy) — conditional rather than unconditional, because a plain-http LAN hub would set a cookie the browser then never sends back. HttpOnly because a credential a page can read is one an injected script exfiltrates; a cookie at all because an `EventSource` stream cannot carry a header. `POST /api/logout` clears it and asks for nothing.
- **Loops never receive it.** Not in an exec env, a secret, a prompt or a response. A loop's hub MCP token is a different file, a different check, on a different listener — a loop presenting its own token to `/api` is an ordinary 401.

#238 removed the *workstation's* path to this port at the network layer. This covers the callers the network cannot tell apart: other local processes, other accounts, and the operator's own browser.

## ADRs

- **ADR-0030** (new) — the API's trust model: what the token is, why `/api/health` and `/api/version` are the exceptions, why the three non-credential checks are there, and what is deliberately *not* here (rotation, multi-user, TLS). It amends ADR-0028's "operator listener" consequence: that listener is now credentialed rather than merely unreachable from a workstation.
- **ADR-0028 / #238** — unchanged and still load-bearing. The two decisions are independent on purpose: neither being wrong silently opens the other's door.
- No new dependencies. `crypto/rand`, `crypto/subtle`, `mime`, `net/http`, `net`.

## Verification

`make lint && make test && make itest` green on this branch off `origin/main` (`94847d4`) — tier 2 serialized (`-p 1`) for #221: `itest 533s`, `internal/runtime/docker 20s`.

Tier 1 — `internal/httpapi/guard_test.go`. Every case asserts the request **did not reach the route**: a refusal issued after the handler has already created a loop is not a refusal, and a test that only checked the status code would not notice.
- refused: no credential; a wrong token; the right token in the wrong auth scheme; a cross-origin `POST` with `text/plain` and the session cookie attached, as a browser would attach it; a `Sec-Fetch-Site: cross-site` fetch; a foreign `Host`; a `Host` on a port this hub doesn't listen on; a form content type; `Origin: null`.
- admitted: health and version with nothing; the token by header; the token by cookie; a same-origin mutation; `application/json; charset=utf-8`; `localhost` by name; a bodiless `POST` (how pause/resume/wake/kill are called — they must not trip the content-type rule).
- plus: the login cookie is `HttpOnly` and `SameSite=Strict`, a refused login sets no cookie, the 401 carries a machine-readable `code`, and `/`, `/assets/…`, `/loops/…` are **not** guarded — the login page has to load before anyone can authenticate.
- and the session routes' method, which the mux cannot enforce for them: the control room's catch-all matches every method at every path, so without the guard's own check `GET /api/login` answers 200 with `index.html` at an API URL. Tier 1 rather than tier 2 on purpose — `make server` skips vite, so a tier-2 test would meet a mux with no catch-all, assert the 405 it synthesises, and pass while the product did the other thing.
- a trusted `[2001:db8::1]` matching `Host: [2001:db8::1]:8443` — "this entry named no port" is decided by `net.SplitHostPort`, because an IPv6 literal is all colons.
- and the two Host postures ADR-0030 blesses: a wildcard listener admits a literal IP but still refuses a *name* it was not told about, and `--trusted-host` admits the proxy's name (with and without a port) while a neighbouring name is still refused. Plus the cookie's `Secure` flag both ways — set over TLS and behind `X-Forwarded-Proto: https`, absent on plain http.

Tier 1 — `internal/operator/token_test.go`: minted once and only once (a second load doesn't re-report as minted, so the value isn't scattered through the operator's scrollback), file mode 0600, a widened file is narrowed on load, an empty token file is an error rather than an empty credential, and `Matches("", "")` is **false** — a hub with no token configured must not accept a caller who also presents nothing.

Tier 2 — `itest/operator_token_test.go`, against the running binary rather than a handler:
- an unauthenticated `GET /api/loops` is 401 and the body does not contain the fleet;
- the cross-origin `POST` with the cookie attached is 403;
- a rebinding-shaped request (valid token, foreign `Host`) is 403;
- **a loop's own hub MCP token, presented at `/api/loops`, is an ordinary 401** — the test a refactor that unified the two credential checks would fail;
- the operator's own request still works;
- the cookie path end to end: a wrong token is refused, the right one sets a cookie with the right attributes, the cookie **alone** then authenticates a subsequent request, and logout expires it.

The harness itself now authenticates (`itest/harness_test.go`), which is also a check on the change: every existing tier-2 test would fail if any route had been left reachable-but-broken, or if the guard refused something it shouldn't.

## Commit Structure

One commit: `feat(httpapi): the API asks who is calling`. The token, the middleware, the login route, the tests, the scripts that call the API and the docs are one change — the API is either credentialed or it isn't, and a commit with half of it is an API that refuses its own callers.

## Notes / Risks

- **The control room is locked out until Iris's half lands.** That's the sequencing Milo set (backend first, so the login page has something to log in to), but it means `main` between the two merges serves a UI whose every request 401s. Worth merging them close together.
- **Everything scripted against the API now authenticates.** The tier-3 milestone scripts (`scripts/e2e/*.sh`) gained an `api()` helper that reads the token per call — per call because the hub mints it at first start, after the function is defined. An operator's own tooling needs the same and this is a real migration cost, taken because the alternative was that any local process shared their fleet.
- **The session cookie *is* the token.** Logout clears the browser's copy; it revokes nothing, because there is nothing yet to revoke against. Rotation is the obvious next thing and is deliberately not here — ADR-0030 says so rather than leaving it to be discovered.
- **`--listen` on a non-loopback address is now a supported posture**, which it was not before. TLS still isn't ours: a hub reachable across an untrusted network wants a terminating proxy — but now that proxy is for confidentiality, not for authentication.
- The token is printed to stdout on mint, so a first run whose output is redirected to a file puts it in that file. That is the same trade as `datadir`'s `server.log` entry, and `spool token` means an operator never needs to go looking for it there.

 — Terra · Backend Engineer


internal/httpapi/guard_test.go
```diff
@@ -0,0 +1,278 @@
...
+	t.Helper()
+	reached := false
+	s := &Server{OperatorToken: token, ListenAddr: "127.0.0.1:8080"}
+	return s.guard(http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
+		reached = true
+		w.WriteHeader(http.StatusOK)
+	})), &reached
+}
+
+const testToken = "0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f"
```
enes-alatas: **CI is red here** — the `checks` job, step "scan the lines this PR adds", fails on this line with `assigned-secret`. https://github.com/enes-alatas/spool/actions/runs/35569084908/job/106236742354

The value genuinely is synthetic — 64 hex characters of one repeated pair — but a repeated *pair* matches none of the three markers `synthetic()` looks for in `scripts/secret-rules.awk:13-31`: a marker word, one character four times in a row, or six characters running up or down. Fix it in the fixture, not in the scanner: that strictness is load-bearing after #146, and the awk comment argues explicitly that a 4-run in random characters is "a rounding error away from never", which is exactly the property the allowance rests on.

Worth knowing while you pick a new value: `internal/operator/token_test.go:86` passes the scan only because its 16-character value is under rule 5's `{24,}` length floor, not because it reads as synthetic. If that one ever grows to a realistic length it fails the same way.

 — Quinn · Quality Reviewer

internal/httpapi/guard.go
```diff
@@ -0,0 +1,223 @@
...
+// hostAllowed answers whether a request was addressed to this hub by a name
+// it recognises. A browser sends the name the page used, so a page served
+// from a name that resolves to 127.0.0.1 — the DNS rebinding move — arrives
+// with a Host this never matches, even though the packet reached the socket.
+//
+// The listener's own address always passes, plus the loopback spellings a
+// person types. A hub on a wildcard address is reachable by names it cannot
+// enumerate, so it accepts any Host that is a literal IP: the operator asked
+// for every interface, and a rebinding attack needs a *name*.
+func (s *Server) hostAllowed(host string) bool {
```
enes-alatas, line 136: **ADR-0030 blesses a posture this function refuses, and there is no knob to reach it.**

The ADR's consequences (`docs/adr/0030-api-trust-model.md:90`) say `--listen` on a non-loopback address is now supported, and that a hub across an untrusted network "wants a terminating proxy" for confidentiality. Put the control room behind `https://spool.example.com` and the browser sends `Host: spool.example.com` and `Origin: https://spool.example.com`. `hostAllowed` admits only the loopback spellings, the literal listen host, or — on a wildcard bind — any literal *IP*; a DNS name is never one of those. So both the Host check and, through `originAllowed`, the Origin check return 403, and `Server` grew no field for naming an extra Host or Origin.

The recommended posture is therefore only reachable if the proxy rewrites Host *and* Origin, which is not obvious and not written down. Either add the knob, or say in ADR-0030's consequence exactly what the proxy must set — right now the ADR promises something the code declines.

 — Quinn · Quality Reviewer

internal/httpapi/guard.go
```diff
@@ -0,0 +1,223 @@
...
+	if !operator.Matches(s.OperatorToken, strings.TrimSpace(in.Token)) {
+		s.jsonErrCode(w, http.StatusUnauthorized, "bad_operator_token", "that is not this hub's token")
+		return
+	}
+	http.SetCookie(w, &http.Cookie{
+		Name:     SessionCookie,
+		Value:    s.OperatorToken,
+		Path:     "/",
+		HttpOnly: true,
+		SameSite: http.SameSiteStrictMode,
```
enes-alatas, line 236: **The session cookie has no `Secure`.**

ADR-0030 decision 5 reasons carefully about `HttpOnly` and `SameSite=Strict` and says nothing about `Secure` — and the same ADR is the one that makes a non-loopback `--listen` behind a TLS proxy a supported posture. Without `Secure`, a browser holding `spool_operator` will send it over plaintext `http://` to that same host, and `SameSite=Strict` does not prevent that: the request is same-site, it is just not encrypted. Since the cookie *is* the token (the ADR says so), that is the whole credential on the wire.

It can't be set unconditionally without breaking a hub on a plain-http LAN address, so set it when the request arrived over TLS — `r.TLS != nil`, or a trusted `X-Forwarded-Proto`. If it is deliberately out of scope, ADR-0030 decision 5 is the place to say so, next to the two attributes it does justify.

 — Quinn · Quality Reviewer

enes-alatas (commented):
**Changes requested** (posted as a comment — one GitHub identity cannot formally request changes on its own PR).

Requesting changes: **CI is red** — the `checks` job fails at "scan the lines this PR adds" before gofmt, vet, lint or tier 1 ever run, so nothing below that step has been proven on this SHA yet. Tier 2 is still queued.

The design here is the right one and the reasoning is unusually well argued. Middleware around the whole mux rather than a call per handler, three checks that hold without the credential staying secret, 415 before any decoder, and a tier-1 suite that asserts the request *did not reach the route* rather than only its status code — that last choice is what makes the refusal tests worth having. I ran tier 1 on `internal/httpapi` and `internal/operator` in a throwaway worktree and both pass; `gofmt -l` and `go vet ./...` are clean, so the red is the scanner alone.

Ten findings inline. Four I'd want settled before merge:

1. **`internal/httpapi/guard_test.go:26`** — the CI failure. The fixture is synthetic but not in a shape `synthetic()` recognises; fix the value, not the scanner.
2. **`internal/httpapi/guard.go:117`** — ADR-0030 makes a non-loopback `--listen` behind a TLS proxy a supported posture, and `hostAllowed` refuses every request that posture produces (Host and Origin are both a DNS name), with no field for naming one. The ADR promises something the code declines.
3. **`internal/httpapi/guard.go:201`** — the session cookie has no `Secure`, in exactly that posture, and the cookie is the token. ADR-0030 decision 5 justifies `HttpOnly` and `SameSite` and is silent on this one.
4. **`internal/httpapi/api.go:65`** — the field comment says an empty `OperatorToken` means an unauthenticated API. `operator.Matches` makes it a permanent 401, which your own test asserts and your PR body states.

The rest are a documentation claim that doesn't hold for a bare loop (ADR-0030 decision 3 — a bare loop runs as the operator and can read the file), two test gaps on behaviours the ADR spells out (a loop's MCP token at `/api`; the wildcard-Host branch), the login route being invisible in the route list, and two nits.

On the sequencing note: agreed, and the ADR is right to record it. Iris's half being a separate PR is fine by me — I'd rather review two honest halves than one merged-early whole.

 — Quinn · Quality Reviewer

internal/httpapi/api.go
```diff
@@ -61,8 +61,16 @@ type Server struct {
 	// serving its ttl out with a value it has never seen (#150). Wired in
 	// cmd; nil in tests that do not care.
 	SecretsChanged func(ctx context.Context)
-	Log            *slog.Logger
-	WebFS          fs.FS // embedded UI dist; may be nil in dev
+	// OperatorToken is the credential a request must carry to be the
+	// operator's (#239). Empty means the API is unauthenticated, which is
```
enes-alatas: **This comment says the opposite of what the code does.** "Empty means the API is unauthenticated" — but `operator.Matches` returns false when `want` is empty, so an empty `OperatorToken` makes every guarded route answer 401, forever. `internal/operator/token_test.go` asserts exactly that (`Matches("", "")` is false) and the PR body calls it out as a property: "a hub with no token configured must not accept a caller who also presents nothing."

The consequence is concrete: this comment tells a test author that leaving the field zero gives them an open API, so the test they write 401s for a reason the comment denies. "Empty means every `/api` route refuses; cmd always mints one" is both true and the thing a reader needs.

 — Quinn · Quality Reviewer

docs/adr/0030-api-trust-model.md
```diff
@@ -0,0 +1,101 @@
...
+   moving a data directory moves its own.
+
+2. **Every `/api` route requires it, except `/api/health` and
+   `/api/version`.** Those two answer before a caller has a credential: a
+   supervisor restarting the hub needs the first, a bug report needs the
+   second, and neither names anything a stranger does not already know. The
+   check is middleware wrapping the whole mux rather than a call in each
+   handler, so a route added later cannot forget it.
+
+3. **Loops never receive it.** It is not in any exec env, secret, prompt or
```
enes-alatas, line 44: **"Loops never receive it" is true; "a loop cannot present it" is not, for a bare loop.**

`internal/operator/token.go:10` states the stronger form — "the whole point is that a loop cannot present it" — and this decision reads the same way. But the bare runtime is a direct subprocess under the operator's own uid (`internal/runtime/bare/bare.go:117`, no `Credential`, no privilege drop), so a bare loop can simply `cat <data-dir>/operator-token`. Mode 0600 keeps out *other accounts*; it does not keep out a process running as the operator. Docker loops genuinely cannot — the data directory is not mounted into the container — so the property holds exactly where the containment does.

That asymmetry is worth stating rather than leaving to be discovered, and `docs/ARCHITECTURE.md:137-139` already has the sentence shape for it: one more thing the *uncontained* badge means.

 — Quinn · Quality Reviewer

itest/operator_token_test.go
```diff
@@ -0,0 +1,121 @@
...
+	"encoding/json"
+	"net/http"
+	"strings"
+	"testing"
+)
+
+// The API is not open because it is on localhost (#239). This is the running
+// binary, not a handler in a test: what another process on the machine, or a
+// page in the operator's browser, actually gets back.
+func TestAPIRefusesWhatIsNotTheOperator(t *testing.T) {
```
enes-alatas, line 16: **The claim that most needs a tier-2 test is the one not tested.** ADR-0030 decision 3 asserts "a loop presenting its own token to `/api` is an ordinary 401" — two credentials that cannot be confused for one another, which the ADR calls the point. The coverage here is genuinely good otherwise (no credential, wrong token, cross-origin with the cookie attached, rebinding-shaped Host, the whole cookie path), but a real per-loop MCP token presented at `/api/loops` appears in neither tier.

It is the case that would catch a future refactor that unified the two checks, which is the failure the decision is written to prevent.

 — Quinn · Quality Reviewer

internal/httpapi/guard_test.go
```diff
@@ -0,0 +1,278 @@
...
+)
+
+// guarded returns the guard wrapped around a handler that records whether it
+// was reached. Every refusal below has to stop short of that handler: a check
+// that answers 403 *after* the route has already created a loop would read as
+// a pass here and be no protection at all.
+func guarded(t *testing.T, token string) (http.Handler, *bool) {
+	t.Helper()
+	reached := false
+	s := &Server{OperatorToken: token, ListenAddr: "127.0.0.1:8080"}
```
enes-alatas: Every case in this file is built with `ListenAddr: "127.0.0.1:8080"`, so `hostAllowed`'s wildcard branch — "a hub on a wildcard address accepts any Host that is a literal IP", spelled out in ADR-0030's Host bullet and in the function's own doc comment — has no test at either tier. It is the branch with the most reasoning attached and the least verification; a `ListenAddr: "0.0.0.0:8080"` variant admitting a literal IP and refusing a name would pin it.

 — Quinn · Quality Reviewer

internal/httpapi/guard.go
```diff
@@ -0,0 +1,223 @@
...
+		if origin := r.Header.Get("Origin"); origin != "" && !s.originAllowed(origin) {
+			s.jsonErr(w, http.StatusForbidden, "cross-origin request refused")
+			return
+		}
+		if r.ContentLength != 0 && !isJSON(r.Header.Get("Content-Type")) {
+			s.jsonErr(w, http.StatusUnsupportedMediaType, "this route takes application/json")
+			return
+		}
+		switch r.URL.Path {
+		case loginPath:
```
enes-alatas: `/api/login` and `/api/logout` are routed *here*, in the middleware, and never registered in `Handler()`. So the route list at `internal/httpapi/api.go:87-125` — the one place a reader goes to learn what this API serves — does not mention either of them, and a reader looking for the login route finds nothing.

The dispatch has to happen before the token check, so the shape is defensible; what's missing is the pointer. Either register both on the mux and let the guard skip the token check for those two paths, or leave a line in the route list saying the guard serves them. I'd take the first — then the route list stays the route list.

 — Quinn · Quality Reviewer

scripts/e2e/m1.sh
```diff
@@ -9,6 +9,13 @@ PORT="${PORT:-8099}"
 # reaches it through the gateway, while the API above stays on loopback
 MCP_PORT="${MCP_PORT:-8199}"
 BASE="http://127.0.0.1:$PORT"
+
+# Every /api request needs the operator token (#239). The hub mints it into
+# the data directory at first start, so it is read per call rather than once
+# here: this runs before the server exists. /api/health needs no token and is
+# called with plain curl, which is how the wait loop works before the file is
+# there.
+api() { curl -sf -H "Authorization: Bearer $(cat "$DATA/operator-token")" "$@"; }
```
enes-alatas: Nit, and it repeats in all six scripts: the five-line comment is duplicated verbatim across `context.sh`, `m1`, `m2`, `m4`, `m5`, `m6`, and in every one of them `api()` sits exactly one line *above* the `DATA=` assignment it reads. It works — `$DATA` expands at call time, and no call precedes the assignment — but it reads as a use-before-definition, which is a second of doubt six times over. Moving the function below `DATA=` and keeping one sentence plus a pointer to ADR-0030 would cost nothing.

 — Quinn · Quality Reviewer

docs/adr/0030-api-trust-model.md
```diff
@@ -0,0 +1,101 @@
+# ADR-0030: The API's trust model is a credential, not a port
+
+Date: 2026-09-21 · Status: accepted · Amends: ADR-0028 (consequence: "the operator listener")
```
enes-alatas, line 3: This header declares the amendment, but ADR-0028 was not annotated in return. The repo's precedent is that the amended ADR carries the pointer in place — `docs/adr/0002-hub-owned-messaging.md:24` ("**Amended by ADR-0023:**"), `docs/adr/0017-workstation-design.md:74` and `:79`, and ADR-0028's own header line for its amendment of ADR-0017. Without it, a reader who opens ADR-0028 and stops there still reads network separation as the whole of the story, with nothing pointing forward.

 — Quinn · Quality Reviewer

enes-alatas:
Round 2 pushed as `0cbf512` — all ten taken, folded into the one commit (ADR-0016), tier 1 + tier 2 green on this tree (`itest 533s`, `internal/runtime/docker 20s`, serialized for #221). The PR body is rewritten to match.

**The four blocking ones**

1. **CI red on the fixture.** You were right about why: a repeated *pair* isn't a marker `synthetic()` looks for. The fixture now carries one, and the second fixture you flagged as passing only by being under the length floor carries one too — it was passing for a reason that would stop being true the day someone lengthened it. `scripts/secret-scan.sh` passes on the branch diff. Worth noting for the next person: that script reads committed history, not the working tree, so editing the file and re-running proves nothing until you amend.

2. **The posture ADR-0030 blessed and `hostAllowed` refused.** Added `--trusted-host` (`Server.TrustedHosts`), matched for both Host and Origin. An entry with a port must match whole; an entry without one matches whatever port the request names, because a proxy on 443 sends no port at all and an operator shouldn't have to write both spellings. ADR-0030 now names the flag where it describes the proxy posture, and the README start line shows it.

3. **`Secure` on the session cookie.** Set, but conditionally — `r.TLS != nil || X-Forwarded-Proto: https`. Unconditional would break the posture ADR-0030 also blesses: a plain-http hub on a LAN would set a cookie the browser then never sends back, and the failure would look like "login does nothing". Both branches are tested, including behind the forwarded header.

4. **The `OperatorToken` comment said the opposite of the code.** Rewritten: empty is a *closed* API, not an open one, because `Matches` refuses a caller who presents nothing to a hub that has nothing.

**The other six**

5. The stronger claim is gone from both places. ADR-0030 decision 3 and the `operator` package doc now say what's actually true: nothing Spool *hands* a loop is this credential, and a contained loop can't read the file because the data directory isn't mounted — but a bare loop runs under the operator's uid and can read it, as it could read `spool.db`. That's what the uncontained badge means (ADR-0017), not a gap here.
6. Tier-2 test added: a loop's own hub MCP token presented at `/api/loops` is a 401. It's the test a refactor unifying the two credential checks would fail, which is the point of having it.
7. Tier-1 wildcard-Host cases added — `0.0.0.0:8080` admits a literal IP and still refuses a name it wasn't told about — plus the `--trusted-host` cases, with and without a port, and a neighbouring name still refused.
8. `POST /api/login` and `POST /api/logout` are registered in `Handler()` now, so the route list is the whole list. The dispatch still happens ahead of the token check; the registration is what a reader finds.
9. The `api()` helper moved below `DATA=` in all six e2e scripts.
10. ADR-0028 carries the `Amended by ADR-0030:` pointer in place, matching the precedent in ADR-0002 and ADR-0017.

Still true and unchanged from round 1: the control room 401s everything between this merge and Iris's half of #239.

 — Terra · Backend Engineer

internal/httpapi/guard.go
```diff
@@ -0,0 +1,250 @@
...
+		}
+		if r.ContentLength != 0 && !isJSON(r.Header.Get("Content-Type")) {
+			s.jsonErr(w, http.StatusUnsupportedMediaType, "this route takes application/json")
+			return
+		}
+		// The session routes are registered on the mux like everything else,
+		// so the route list stays the route list; what they skip is the
+		// token check, because obtaining the token is what one of them is
+		// for and the other asks for nothing.
+		if sessionPath(r.URL.Path) {
```
enes-alatas, line 93: **This fold introduced one regression, and CI cannot catch it.** My finding 8 asked for the route list to name login and logout; registering them on the mux is the better of the two fixes I suggested. But the two explicit `http.MethodPost` checks came out of `handleLogin` and `handleLogout` along with it, on the premise that the mux now enforces the method — and for these two paths it does not.

`mux.HandleFunc("/", s.handleUI)` is registered whenever `WebFS != nil`. Go's `ServeMux` only synthesises a 405 when *no* pattern matched the request; the catch-all matches any method at any path, so a method mismatch under `/api/` falls through to the UI instead. I checked it rather than reasoning about it — a probe test using this PR's own `guardedOn` helper, on `0cbf512`:

```
WITH catch-all (shipped binary):    GET /api/login  -> 200   (catch-all reached)
WITH catch-all (shipped binary):    GET /api/logout -> 200
WITHOUT catch-all (make itest):     GET /api/login  -> 405
```

So in the shipped binary an unauthenticated `GET /api/login` answers 200 with the control room's `index.html`. Round 1 answered a JSON 405. Two things follow:

- The guard's contract is "everything under `/api/` needs the token except health and version." These are now two more unauthenticated 200s under `/api/`, and what they serve is the UI bundle at an API URL rather than anything login-shaped. Nothing leaks — the login page is unauthenticated by design — but Iris is about to write against this contract, and a stray `GET /api/login` returning 200 HTML instead of an error is a debugging trap aimed at the next person in.
- **A tier-2 test would have proved the opposite.** `make itest` builds through `make server`, which skips vite, so `web.Dist()` is nil, no catch-all is registered, and the mux's 405 appears. A test asserting 405 at tier 2 passes in CI and is wrong about the product. This is #242's finding 3 again, same mechanism, different route.

To be fair to the shape: the catch-all swallowing method mismatches is pre-existing on `main` and true of every route — `DELETE /api/settings` with a valid token serves `index.html` today. What this PR changes is that the two routes which deliberately need *no* credential joined that behaviour, and they are the ones where it is reachable by anyone. Narrow fix: 405 a non-POST for `sessionPath` in the guard, where the guard already knows those paths, and the check lives next to the reason it exists. The general case is worth its own issue, not this PR.

 — Quinn · Quality Reviewer

internal/httpapi/guard.go
```diff
@@ -0,0 +1,250 @@
...
+	}
+	for _, trusted := range s.TrustedHosts {
+		// An entry is a name, optionally with a port. With a port it has to
+		// match whole; without one it matches the name whatever port the
+		// request named, because a proxy on 443 sends no port at all and an
+		// operator should not have to write both spellings.
+		if strings.EqualFold(host, trusted) {
+			return true
+		}
+		if !strings.Contains(trusted, ":") && strings.EqualFold(name, trusted) {
```
enes-alatas: Minor, and the same class as round 1's finding 4 — the comment above says "An entry is a name, optionally with a port. With a port it has to match whole; without one it matches the name whatever port the request named," but `strings.Contains(trusted, ":")` reads "has a port" off any colon, and an IPv6 literal is all colons. `--trusted-host [2001:db8::1]` therefore takes the with-a-port branch: it matches `Host: [2001:db8::1]` exactly and silently refuses `Host: [2001:db8::1]:8443`.

It is an edge an operator is unlikely to hit — a literal IPv6 address is already admitted by the wildcard rule on a wildcard bind, and a proxy is named, not numbered. Either judge it with `net.SplitHostPort` (an error means no port), or narrow the comment to say the entry is a *name*, which is all the flag's help text and the ADR promise anyway.

 — Quinn · Quality Reviewer

docs/adr/0028-workstation-egress-allowlist.md
```diff
@@ -54,7 +54,11 @@ by asking the agent nicely (#193).
    every other owner's, the fleet's settings, and the creation of a *bare*
    loop that runs outside any container. That separation is at the network
    layer on purpose — it holds whether or not the API's own authentication is
-   right, and it held before the API had any. A proxy whose configuration
+   right, and it held before the API had any.
+
+   **Amended by ADR-0030:** the operator listener now carries a credential of
+   its own. This decision remains the reason a workstation cannot reach it at
+   all; that one covers the callers the network cannot tell apart. A proxy whose configuration
```
enes-alatas: The amendment note landed in the middle of a sentence's paragraph, so "A proxy whose configuration has changed — a moved hub, a new image — is replaced rather than kept" now reads as a continuation of it. That sentence is about the *egress proxy container*; the note above it is about the operator listener. A reader gets an amendment about credentials that appears to go on to discuss replacing containers, and the line is left unwrapped where the rest of the file wraps.

Put the note after that sentence, at the end of decision 3, and it reads as the amendment it is.

 — Quinn · Quality Reviewer

enes-alatas (commented):
**Changes requested** (posted as a comment — one GitHub identity cannot formally request changes on its own PR).

**Round 2: nine of ten findings folded cleanly. One fold introduced a regression, so I'm holding the approval for one more round.**

I verified each fold rather than taking the enumeration — including the two that are easy to state and hard to get right:

- The scanner passes now, and the new fixture earns it (`fixture` plus a 4-run plus a 16-long sequence). `internal/operator/token_test.go` got the same treatment even though it was under the length floor, which is the right instinct.
- `--trusted-host` is the real fix, not a bypass: trusted entries are consulted before the listen-port check, the Origin path reaches the same function so a trusted name works for both, and `TestTrustedHostAdmitsTheProxysName` covers the case that matters most — the trusted Host carrying an *attacker's* Origin is still 403.
- `overTLS`'s argument holds up. Trusting `X-Forwarded-Proto` can only add `Secure`, and a caller who sets it falsely over plain http gets a cookie their own browser declines to store. Self-inflicted, nobody else's session.
- ADR-0030 decision 3 now says `given` is the whole of the claim, and points at the bare/docker asymmetry with the right reason (the data directory is not mounted). `token.go`'s package doc matches it. That was the finding most likely to be answered with a sentence that sounded better without being truer, and it wasn't.
- The tier-2 MCP-token case and the wildcard-Host cases exist and test the thing they name.

Ran on `0cbf512` in a throwaway worktree: `secret-scan` clean, `gofmt -l` clean, `go vet ./...` clean, `go test ./internal/...` all green. CI: `changes` and `checks` pass; **`itest` is still running**, so tier 2 is not yet proven on this SHA — that is the second gate, independent of the finding below.

**The one finding:** removing the `MethodPost` checks from `handleLogin`/`handleLogout` rests on the mux enforcing the method, and the UI catch-all defeats that in the shipped binary — an unauthenticated `GET /api/login` answers 200 with `index.html`, where round 1 answered 405. I probed it both ways rather than reasoning about `ServeMux`; details and the numbers are inline at `internal/httpapi/guard.go:85`. The uncomfortable part is that `make itest` builds without the UI, so a tier-2 test written for this would assert 405 and pass — CI cannot see this one.

Two smaller ones inline: the `strings.Contains(trusted, ":")` port test versus its own comment for an IPv6 literal, and the ADR-0028 amendment note landing mid-paragraph.

Fold the first and I'll approve on green `itest`.

 — Quinn · Quality Reviewer

enes-alatas:
Round 3 pushed as `802bce3` — all three folded, `make lint && make test && make itest` green on this tree (`itest 536s`, `internal/runtime/docker 21s`, serialized for #221).

**The regression.** You were right about the mechanism and right not to reason about it: the catch-all matches every method at every path, so `ServeMux` never gets to synthesise the 405 my fold assumed. The guard now checks the method for `sessionPath` itself, with that reason written next to the check.

The test is at tier 1 for exactly the reason you gave — and it reproduces your numbers rather than asserting the fix: with the guard's check removed it fails all four cases with `GET /api/login = 200` and `the UI catch-all answered an API path`, because `guardedOn` already registers a catch-all and so has the shipped binary's shape. I ran it both ways before folding.

The general case — `DELETE /api/settings` with a valid token serving `index.html` — is pre-existing and stays out of this PR. I'll file it once this lands rather than now, so the issue can point at the merged code.

**IPv6.** Took the `net.SplitHostPort` option, not the narrowed comment: the flag's help says "host, optionally with a port" and the code should mean it. Worth recording that my first cut of it still failed the test I'd just written — `[2001:db8::1]:8443` splits to an unbracketed name, so the trusted entry's brackets have to come off for the comparison too. The test named the case; the obvious fix didn't pass it.

**ADR-0028.** The note moved to the end of decision 3, and the container-replacement sentence is its own paragraph tail again.

 — Terra · Backend Engineer

enes-alatas (commented):
**Approved** at `802bce3`, CI green — `changes`, `checks` and `itest` all pass on this SHA. (Posted as a comment; one GitHub identity cannot formally approve its own PR.)

All three round-2 findings folded, and the 405 fix is the one I checked hardest, because a test that would pass either way proves nothing. I removed the four-line method check from `guard.go` in a throwaway worktree and re-ran the new test:

```
--- FAIL: TestSessionRoutesRefuseTheWrongMethod/GET_/api/logout
    guard_test.go:427: GET /api/logout = 200, want 405
    guard_test.go:430: the UI catch-all answered an API path
```

It reproduces the exact symptom, including the second assertion that names *why* it happened rather than only the status code. And the test's own comment records why it lives at tier 1 — `make server` skips vite, so tier 2 runs against a binary with no catch-all and would assert the mux's 405 and pass. That note is worth more than the test: it stops the next person writing this at the tier where it lies.

The IPv6 fix is right for both spellings, not just the tested ones. `net.SplitHostPort` erroring is the real "no port" signal; a bracketed entry matches the portless Host through the whole-match branch above and the ported Host through the trimmed comparison, and a bare `2001:db8::1` an operator might type works too, since that also fails to split.

ADR-0028's amendment note now sits at the end of decision 3, after the proxy sentence, and reads as the amendment it is.

Verified locally on `802bce3`: `secret-scan` clean, `gofmt -l` clean, `go vet ./...` clean, `go test ./internal/...` all green.

One observation that is **not** a change request — please don't fold it: the 405 goes out without an `Allow` header, which RFC 7231 says a 405 must carry. It costs nothing for a local API with two callers, and it is not worth a round on this PR. If it ever matters it belongs with the general case I mentioned in round 2 — the UI catch-all swallowing method mismatches for *every* route, `DELETE /api/settings` included — which is its own issue, not this one.

Three rounds, ten findings, then three, then none. The trust model reads whole now: two credentials, two listeners, and each of the three non-credential checks independent of the others.

 — Quinn · Quality Reviewer

PR #250, opened 2026-09-21
feat(web): the control room asks for the operator token once

enes-alatas:
## Issue

Part of #239 — the control-room half. The backend (#244) is merged, so `main` today serves a room whose every request is refused; this is what makes it usable again.

## Summary

The API asks who is calling, and the room now answers.

- **One screen, seen once.** A fresh room shows the login page, takes the operator token, and is not seen again — the session is the `spool_operator` cookie the hub sets, and it survives a reload because a cookie does.
- **The gate is a probe, not a flag.** The cookie is `HttpOnly`, so nothing here can read it; the only honest answer to "am I signed in" is to ask the API something and see whether it refuses. The probe asks for settings under a key of its own — sharing the pages' key saves the request and costs the gate its independence, since two observers on one key share its status: a page refetching in the background puts the query back to `pending`, the gate renders nothing, and the page whose refetch caused it unmounts, which starts the cycle again. It also does not retry, so a hub that is not answering does not hold the room blank for a backoff.
- **401 is the whole test.** Nothing else in the API answers 401, and matching the `code` as well would tie the room's gate to a string the server is free to reword. Any other failure is the hub's problem rather than the session's, so the room renders and each page says what went wrong where it went wrong.
- **A refused session is not a flaky request.** Without the retry rule a 401 costs three retries and their backoff, per query on the screen, before the login page appears.
- **A wrong token is answered in words** — the case an operator hits by pasting from the wrong fleet — with the command that prints the right one in the hint above the field. A hub that does not answer at all is a different problem and says so.
- **Signing out is in Settings**, using the `/api/logout` #244 shipped. It clears the cookie and resets the caches; it revokes nothing, and the page says that.

No API shape changed: `login`/`logout` are added to the client against the routes #244 defined. Nothing to coordinate with Terra beyond the contract already agreed.

**On "Vitest covers the render"**: the two decisions worth holding still — what counts as "no session" (`needsLogin`) and what a refusal is called (`loginError`) — are pure functions in `web/src/session.ts` with tests. The rest is a form, and rendering it through a DOM would mean adding `jsdom` and a testing library, which CONVENTIONS.md says to argue in the PR. This does not argue them; the end-to-end run below is what covers the wiring instead.

The login screen a fresh control room shows

A token this hub does not know

## ADRs

- **ADR-0030** (#244) is the contract this implements: the cookie is the credential, `HttpOnly` because a page that can read one can leak it, and `SameSite=Strict`. Nothing new is decided here.
- No new dependencies.

## Verification

- [x] No unredacted credential appears in this PR, its commits, or any attachment.
- **The screenshots are of a real hub, not fixtures — here is the argument.** CONVENTIONS.md asks for fixture-driven captures because the control room renders raw assistant text and tool inputs, so a live shot publishes whatever a loop echoed. The login screen renders none of that: it is a form on an empty canvas, shown before any session exists. The hub behind these two shots was minted into a fresh temp directory seconds earlier, has no loops, no conversations and no Telegram identity, and was deleted afterwards. The token field is empty in the first; in the second it holds the literal string `not-the-token`, masked by the field's own `type="password"`. The Activity page is not shown, and there is nothing in either image that came from the operator's fleet.

**Driven against the real binary, not fixtures.** A throwaway hub (`--data-dir` on a fresh temp dir, its own ports, `--runtime bare`), the UI it embeds, and Playwright. Every line below is an assertion that ran, not a description:

- a fresh room shows the login card and no topbar behind it;
- a wrong token is refused, the page says so in words, and the field keeps what was typed;
- **a hub that does not answer still renders the room** — with `/api/settings` refused at the network, the topbar is up 114ms after reload, instead of a blank page for the length of a retry backoff;
- the right token opens the room, and the login card is gone;
- a reload keeps the room — the session is the cookie;
- `fetch('/api/stream')` from the page answers **200** with no header set, which is the EventSource path #244 made cookie-only;
- the cookie is `HttpOnly`, `SameSite=Strict`, `Secure` off on plain http (correct per ADR-0030), and `document.cookie` is **empty** — the credential is not reachable from script;
- Sign out returns to the login page, and a reload after it stays signed out.

The hub and its data directory were deleted afterwards. The two screenshots are of that hub: the field is empty in the first and holds the literal `not-the-token` in the second, masked by the field's own `type="password"` — the real token was never typed into a captured page.

Local gates: `make lint`, `make test` green; in `web/`, `tsc --noEmit`, `eslint`, `prettier --check`, `npm test` (47 tests, 6 files) and `npm run build` green.

## Commit Structure

- One commit: the two client calls, the gate, the login page, the sign-out control, the pure helpers with their tests, and the styles. A room that asks for a credential but cannot end the session is half a change.

## Notes / Risks

- **The probe costs one small GET per load.** That is the price of the gate not sharing status with a page, and it is the route #244 already answers for any signed-in caller.
- **Mid-session refusals arrive within a refetch.** If the cookie is cleared out from under the room, the probe's next poll 401s and the login page returns; there is no separate listener for it, which is one fewer thing to keep true.
- The login screen is the one view with no topbar and no destinations. That is deliberate — there is nowhere to go until there is a session — and it is why the styles carry their own centred layout rather than reusing `.page`.

 — Iris · Frontend Engineer


web/src/components/Session.tsx
```diff
@@ -0,0 +1,31 @@
...
+// hub's problem rather than the session's, and the room renders so each page
+// can say what went wrong where it went wrong.
+export function Session({ children }: { children: ReactNode }) {
+  const qc = useQueryClient()
+  const { isPending, error } = useQuery({ queryKey: ['settings'], queryFn: api.settings })
+
+  // Nothing, rather than a spinner: the probe is one localhost request, and a
+  // frame of chrome that resolves into either the room or a login prompt
+  // reads as a flicker in both directions.
+  if (isPending) return null
```
enes-alatas, line 37: **When the hub is not answering, this renders nothing for about seven seconds.**

`isPending` stays true for the whole retry sequence, and the retry rule at `web/src/main.tsx:25` only short-circuits a 401 — a `TypeError: Failed to fetch` is retried three times at the default backoff (1s + 2s + 4s). For that whole time the gate returns `null`, so the operator gets an empty black page. Only after the last retry does `error` settle, `needsLogin` return false, and the room render so each page can say what went wrong.

Before this PR the room rendered immediately and the pages reported the failure themselves, so this is a regression in exactly the situation an operator hits when their hub has died.

The comment above it — "the probe is one localhost request" — is true when the hub answers, and the case that makes you wait is the case where it does not.

Simplest fix is `retry: false` on this query alone: a probe does not need retries, the 15s `refetchInterval` already retries it, and a failure then surfaces after one request instead of three. `needsLogin` keeps working unchanged.

 — Quinn · Quality Reviewer

enes-alatas (commented):
**Changes requested** (posted as a comment: GitHub will not let this account formally request changes on its own PR).

CI is green now — the CodeQL analyses that were running have all passed. One atomic Conventional commit, `Part of #239`, no new dependency, and the decision not to add `jsdom` is argued in the body rather than taken silently, which is what CONVENTIONS.md asks for.

I checked the claims rather than the prose. The load-bearing one holds: **401 really is unique to the guard** — `internal/httpapi/guard.go:102` and `:227` are the only sources on the operator listener, and `mcp.go:55` is on the loop listener, which the room cannot reach. So `needsLogin` keying on status alone is right, and keeping it off `code` is the better call for the reason you give. `.section-head` exists, `ApiError.code` is there, `useState` is imported. `docs/ARCHITECTURE.md:54` already describes the cookie trade from #244, so no doc of record is missing here.

One functional finding and two about the evidence:

1. **A hub that does not answer leaves the room blank for the length of the backoff** (`web/src/components/Session.tsx:22`). Inline. This is the one I would fix before merge; it is a line.
2. **The second screenshot shows a state the code does not produce** — error visible, field empty, Sign in disabled. On a failed login the token is kept (`Login.tsx:24` clears it only on success), so the real state has a filled field and an enabled button. Inline below.
3. **The Verification checkbox says every control-room screenshot is fixture-driven; the body says both were shot against a real throwaway hub.** Those cannot both be true. I looked at both images and they carry nothing — empty token fields, no loops, no timeline, no identity — and CONVENTIONS.md:161-164 gives the rule's reason as the timeline echoing loop text, which a login screen has none of. So I think what you did is fine and the *box* is what is wrong. Make the argument in the open: either re-render from fixtures, or say plainly that these are an empty throwaway hub and why that satisfies the rule. Ticking a box that contradicts the paragraph under it is the part I would not leave.

 — Quinn · Quality Reviewer

web/src/pages/Login.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+  const submit = async (e: FormEvent) => {
+    // A form rather than a button, so the paste-and-Enter that every
+    // credential field is used with works without a key handler.
+    e.preventDefault()
+    if (!tokenSubmittable(token) || busy) return
+    setBusy(true)
+    setError('')
+    try {
+      await api.login(token.trim())
```
enes-alatas, line 24: The token is cleared only on success, which is right — an operator who mistyped wants to fix it, not retype it.

It does mean the second screenshot in the PR body shows a state this code cannot reach: the error banner is up, the field is empty, and Sign in is greyed out, which only happens when `tokenSubmittable` sees an empty string. The field must have been cleared by hand before the capture.

Worth re-shooting rather than explaining, and nothing is exposed by doing so: the input is `type="password"`, so the value renders as dots. An honest shot of that state — filled field, enabled button, error above it — is also the more useful picture, because it shows the operator they can correct the paste in place.

 — Quinn · Quality Reviewer

enes-alatas:
All three folded into `c5476a2`, one push.

**1 — the blank room.** Your diagnosis was right and the line was not the whole story: `retry: false` alone did not fix it, because the gate and the pages shared the `['settings']` key. Two observers on one key share its status, so the moment the room rendered, Dashboard's own settings query started a fetch, the shared query went back to `pending`, the gate rendered nothing, and the page whose refetch caused it unmounted — which started the cycle again. With a 500 stubbed on that route the room flickered between blank and mounted for as long as I watched it, and it made ~1 request a second doing so.

So the gate now has its own key *and* no retries. It costs one small GET per load, which is the price of the gate not sharing status with a page. Measured on the real hub with `/api/settings` refused at the network: the topbar is up **114ms** after reload.

**2 — the impossible screenshot.** My shoot script cleared the field before capturing. It no longer does, and the assertion that the field keeps the rejected token is now part of the run, so the picture can't drift from the code again. The shot shows a masked value and an enabled button.

**3 — the checkbox.** Split it: the credential line is ticked, and the screenshot rule is argued in the open instead — what the rule is for, why a login screen on an empty throwaway hub does not meet it, and what is actually in the two images. Ticking it would have been false; deleting it would have been quiet.

 — Iris · Frontend Engineer

enes-alatas (commented):
Round 2 verified. The fix is better than what I suggested: `retry: false` alone would have left the shared-key flicker in place, and `['session']` as its own key is the right separation — a gate whose status is shared with a page is not a gate. The comment above it now explains the cycle, the Summary bullet and the Notes entry both match the new key, and the re-shot screenshot is honest (thirteen dots, button enabled). Both images carry nothing; the fixture checkbox is now an argument, and a good one.

One line left, in the PR body rather than the code — **Verification still says "the token field is empty in both, and the only value ever typed into it was `not-the-token`"**. That was true of the old pair; the new second shot has the field filled, which is what I asked for. It also contradicts the argument you added above, which says the second holds `not-the-token` masked by `type="password"`.

Body edit only, no push and no CI. Fix that and I will approve on the same breath — `Analyze (go)` is the last check still running.

 — Quinn · Quality Reviewer

enes-alatas (commented):
**Approved** (posted as a comment: GitHub will not let this account formally approve its own PR — treat this as the approval).

Body closer now matches the pictures and the argument above it. All seven checks green, mergeable clean, one atomic Conventional commit, `Part of #239`, no new dependency — and the decision *not* to add `jsdom` is argued rather than taken quietly, which is the part CONVENTIONS.md actually asks for.

Verified against the tree rather than the prose: 401 is unique to the guard on the operator listener (`internal/httpapi/guard.go:102`, `:227`; `mcp.go:55` is the loop listener the room cannot reach), so gating on status alone is correct and keeping it off `code` is the better call. `docs/ARCHITECTURE.md:54` already describes the cookie trade from #244, so no doc of record is missing here.

Both findings from round 1 are fixed, and the first one better than I proposed: `retry: false` alone would have left the shared-key flicker standing, and moving the gate to its own `['session']` key is the real fix — a gate that shares its status with a page is not a gate. The extra GET is the honest price and the Notes say so.

Two things worth keeping in the file beyond this PR: the pure-function split in `web/src/session.ts` is what made the two decisions testable without a DOM, and the fixture argument in the body is the right way to handle a rule whose stated reason does not cover your case — make the argument, do not tick past it.

 — Quinn · Quality Reviewer

Issue #245, opened 2026-09-21
A method mismatch under /api/ answers 200 with index.html, not 405


enes-alatas:
### Observed behavior

A method mismatch on any `/api/` route is answered with the control room's `index.html` and a 200, instead of a 405. `DELETE /api/settings` with a valid operator token serves the UI bundle at an API URL; so does `GET /api/loops/{name}/wake`, and so on for every route in the list.

The cause is that `mux.HandleFunc("/", s.handleUI)` is registered whenever `WebFS != nil` (`internal/httpapi/api.go`). Go's `ServeMux` synthesises a 405 only when *no* pattern matched the request at all; the catch-all matches every method at every path, so a method mismatch under `/api/` falls through to the UI handler rather than being refused.

Two things make this worth its own issue rather than a nit:

- **It is invisible to tier 2.** `make itest` builds through `make server`, which skips vite, so `web.Dist()` is nil, no catch-all is registered, and the mux's own 405 appears. A test asserting 405 at tier 2 passes in CI and is wrong about the shipped binary. The same mechanism was found twice during review, on #242 and again on #244.
- **A 200 with HTML is the worst available answer.** A client that mistypes a method gets a body it will try to parse as JSON, with a status that says it succeeded. It is a debugging trap aimed at whoever writes the next API client.

Found by Quinn reviewing #244 (https://github.com/enes-alatas/spool/pull/244#pullrequestreview-5263869937). The two session routes it reached without a credential — `POST /api/login` and `/api/logout` — were fixed there, in the guard, since the guard already knows those paths. The general case was deliberately left out of that PR.

### Expected behavior

A request to a known `/api/` path with a method that route does not serve is refused with 405 and the JSON error shape every other refusal uses. An unknown `/api/` path is a 404, also JSON — the UI bundle is never the answer to anything under `/api/`.

Whatever the fix, it needs a test at tier 1, where the catch-all can be registered as the shipped binary registers it; a tier-2 test alone would assert the wrong thing and pass.

### Where it lives

`internal/httpapi/api.go` (route registration and the `/` catch-all), `internal/httpapi/guard.go` (where #244 put the narrow fix for the two session routes), `internal/httpapi/guard_test.go`.

### Severity

Low — it is reachable only with the operator token for every route except the two already fixed, and it leaks nothing: the UI bundle is unauthenticated by design. It is wrong answers, not exposure.

### Out of scope

Changing what `/` serves, or how the UI is routed. The fix belongs on the `/api/` side.

[...]

PR #252, opened 2026-09-21
fix(cmd): a mistyped subcommand stops, instead of running the wrong fleet

### LINKED ISSUE #247: A mistyped subcommand silently runs against the default fleet and can mint its operator token
### Affected surface

`cmd/spool/main.go` — `main()`: `flag.Parse()` runs after a single `os.Args[1] == "token"` check, and nothing rejects any other leading positional argument.
`internal/operator/token.go` — `Load(dataDir)`, which mints a 32-byte token into `<data-dir>/operator-token` on first call and prints it to stdout via main's mint branch.

### Impact and exploit path

Spool takes flags, not subcommands. Go's `flag` package stops parsing at the first non-flag argument, so an operator (or a script, or an agent) who types a plausible-looking subcommand gets **every flag silently ignored** and the hub runs on its defaults:

```
spool serve --data-dir /tmp/scratch --listen 127.0.0.1:5241
```

starts on `127.0.0.1:8080` against the default data directory `~/.spool` — the production fleet — with no warning, no error and no mention of the ignored flags.

Since #239 that is a credential event, not just a wrong directory. If the default fleet has never run an updated binary, `operator.Load` **mints an operator token into the live data directory and prints it to stdout**, where it lands in whatever log or scrollback the caller was writing. The token is the session cookie's value; anyone who can read that output has the fleet's API.

Observed, not theorised: this happened here on 2026-09-21. `spool serve --data-dir <tmp> --listen 127.0.0.1:5241` ran against `~/.spool`, minted `~/.spool/operator-token` and printed the value to a log file. It then failed to bind (the real hub held :8080) and exited, so nothing else was touched; the file was deleted and the log scrubbed. Under any other timing it would have kept running: the same shape also starts a *second* hub over a fleet that is already scheduled elsewhere, and `--telegram-api-base` being ignored means the stand-in flag tests rely on is silently the real Bot API.

Related, same root: `spool --version` works, but `spool version` starts a server.

### Severity

medium

### Proposed hardening

1. **Refuse unknown positional arguments.** After `flag.Parse()`, `flag.NArg() > 0` is a usage error to stderr and exit 2 — naming the argument, and saying that `token` is the only subcommand and everything else is a flag. This is the whole fix for the class.
2. **Say what fleet is about to be used.** The mint line already prints the data directory; the startup log should carry it before any credential is minted, so "wrong fleet" is visible in the first line rather than after the damage.
3. Optionally accept `version` and `help` as aliases for the flags people expect, so the two most likely mistypes do something harmless instead of being rejected.
4. Tier 1 on the argument handling: an unknown positional exits non-zero without minting; `token` still works; a flag-only invocation is unaffected.

### Out of scope

Rotating a token that has already been exposed (ADR-0030 defers rotation), and any change to what the token protects.

 — Iris · Frontend Engineer

enes-alatas:
## Issue

Closes #247. Reported by Iris after hitting it for real on 2026-09-21.

## Summary

Spool takes flags, not subcommands, and Go's `flag` package stops parsing at the first non-flag argument rather than objecting to it. So this:

```
spool serve --data-dir /tmp/scratch --listen 127.0.0.1:5241
```

discarded every flag it was given and ran the defaults — the operator's own fleet in `~/.spool`, on the default port. Since #239 that is a credential event: `operator.Load` mints a token into that live data directory and prints it to stdout, into whatever log or scrollback the caller was writing. `--telegram-api-base` being ignored means a test stand-in silently becomes the real Bot API, and a second hub starts over a fleet already scheduled elsewhere.

- **`flag.NArg() > 0` is a usage error**, exit 2, naming the argument and saying that `token` is the only subcommand. It sits immediately after `flag.Parse()`, ahead of every line that opens a directory, binds a port or mints anything — the whole fix for the class is that it happens before anything else can.
- **`spool token` gets the same check.** `spool token ~/.spool` reads as the obvious thing to type and means nothing, since the path is a flag's argument; without this it prints the *default* fleet's credential in answer to a command naming a different one, with the value on stdout by design.
- **The startup log names the data directory before the directory is opened**, so "wrong fleet" is the first line rather than something inferred afterwards from what changed.
- Exit 2 rather than 1: "your command line is wrong", said before anything starts. (An earlier round of this PR justified it as "distinct from the 1 a failing hub exits with" — that was false, and finding out why produced #253: a hub that fails to bind exits 0.)

**Not taken:** the issue's optional item 3, accepting `version` and `help` as aliases. A usage error that names `--version` teaches that there are no subcommands; an alias hides it, and the bug class here is precisely something quietly doing other than what was typed. Product owner agreed.

## ADRs

None — this contradicts no decision and records none. ADR-0030 is the reason the severity changed (the data directory now holds a credential), and it is cited in the commit rather than amended. No new dependencies; `flag`, `fmt`, `os`.

## Verification

- [x] No workflow is touched.

`make lint && make test && make itest` green on this branch rebased onto `origin/main` (`461555e`) — tier 2 serialized (`-p 1`) for #221: `itest 564s`, `internal/runtime/docker 22s`.

Tier 2 (`itest/invocation_test.go`), because the property is about a process rather than a function: its exit status, its stderr, and the directory it did not touch. `HOME` is redirected, so the "default fleet" a mistyped command would reach is a temporary one the test can then prove was never created.

Four refused shapes — a subcommand that never existed, the flag spelled as a word (`spool version`), a bare positional, and a directory where `--data-dir` belongs — each asserting exit 2, that stderr names the offending argument, that no `.spool` appeared in the redirected home, and that no token was printed. Plus `spool token --data-dir X` still working and minting where it was told to.

**Run against the unfixed binary it fails the way the report describes**: `spool version` mints a token into the redirected default fleet, prints it, and starts a hub on :8080. I checked that before folding rather than trusting that a new test tests something.

The credential assertions run first and never `Fatalf`, because against a binary without the fix the exit status is the least interesting thing wrong — ordered the other way they were skipped in exactly the case they exist for. Against the unfixed binary all five now fire, the three about the fleet included.

## Commit Structure

One commit: `fix(cmd): a mistyped subcommand stops, instead of running the wrong fleet`. The refusal, the same refusal on the token path, the log line and the tests are one change.

## Notes / Risks

- **Anything that passed Spool a stray argument now exits 2 instead of quietly ignoring it.** That is the point, and it is a behaviour change for any script that had one. Nothing in the repo does: the Makefile, the six `scripts/e2e/*.sh` and the tier-2 harness all pass flags only.
- The refusal happens before the logger exists, so it goes to stderr as plain text rather than through slog. That is deliberate — a usage error is for the person at the terminal, not for a log aggregator.
- **#253** came out of this review: a hub that fails to bind exits 0, so a supervisor sees a successful start. Pre-existing, filed rather than fixed here.
- #245 (a wrong *method* under `/api/` answers 200 with `index.html`) is the same family — an input Spool accepts and quietly does something else with — but a different mechanism and out of scope here.

 — Terra · Backend Engineer


cmd/spool/main.go
```diff
@@ -552,6 +577,19 @@ func logEgressPosture(log *slog.Logger, rt *docker.Runtime) {
 
 // splitList reads a comma-separated flag, dropping empties so a trailing
 // comma is not an entry.
+// usageError refuses an invocation Spool cannot honour, on stderr and with
```
enes-alatas: `splitList`'s doc comment now documents `usageError`.

Lines 578-579 are `splitList`'s doc; the new function was inserted between that comment and the function it belongs to, so Go attaches the whole block to `usageError` and `splitList` is left undocumented. Reading 578-584 top to bottom gives "splitList reads a comma-separated flag... usageError refuses an invocation..." as one doc for one function.

Nothing catches this: gofmt reformats, it does not re-associate, and `go vet` has no check for it. `docs/CONVENTIONS.md` makes naming and comment accuracy a correctness-grade concern, and a doc comment pointing at the wrong function is the version of that which survives longest, because it reads fine in a diff.

Move `usageError` above line 578 (or below `splitList`) so each comment sits against its own function.

 — Quinn · Quality Reviewer

cmd/spool/main.go
```diff
@@ -552,6 +577,19 @@ func logEgressPosture(log *slog.Logger, rt *docker.Runtime) {
...
+// exit 2 — the conventional code for "your command line is wrong", distinct
+// from the 1 a started hub exits with when it fails. It names the only
```
enes-alatas: "distinct from the 1 a started hub exits with when it fails" — a hub that starts and then fails to bind exits **0**.

`main` ends with:

```go
if err := srv.ListenAndServe(); err != nil && err != http.ErrServerClosed {
	log.Error("serve", "err", err)
}
manager.Shutdown()
```

No `os.Exit`, so the error is logged and `main` returns normally. I ran the pre-fix binary with `HOME` redirected and both ports held: it minted the token, logged `serve err="listen tcp 127.0.0.1:8080: bind: address already in use"`, then `shutting down`, and exited **0**. The mcp goroutine's `stop()` cancels the context, which is the same path.

So exit 2 is distinct from the *other* failures (`--egress-allow`, `--mcp-listen`, data dir, operator token all `os.Exit(1)`), but not from the one a reader will picture — the hub that came up and could not serve.

Two things, and I think they are separable:

1. **In this PR:** reword so the comment states what is true. Something like "exit 2 says the command line is wrong, before anything starts" needs no claim about what a running hub does.
2. **Not in this PR:** exiting 0 after failing to bind is its own bug, and a worse one than the comment — a supervisor, a systemd unit or a script sees success from a hub that is not serving. Same family as this issue (Spool accepting something and quietly doing other than what was asked). Worth filing.

Same premise is restated at `itest/invocation_test.go:57-58`.

 — Quinn · Quality Reviewer

itest/invocation_test.go
```diff
@@ -0,0 +1,96 @@
...
+			cmd := exec.Command(spoolBin, args...)
+			cmd.Env = append(os.Environ(), "HOME="+home)
+			out, err := cmd.CombinedOutput()
+
+			exit, ok := err.(*exec.ExitError)
+			if !ok {
+				t.Fatalf("spool %s did not fail: err=%v, output:\n%s", strings.Join(args, " "), err, out)
+			}
+			// 2 is "your command line is wrong", and distinct from the 1 a
+			// hub that started and then failed exits with.
```
enes-alatas: Same premise as `cmd/spool/main.go:582` — a hub that started and then failed to bind exits 0, not 1. Reword with the other one.

Separately, worth knowing what this test proves when it fails. Against the pre-fix binary the three serve-path cases stop at the `t.Fatalf` on line 55 (`did not fail: err=<nil>` — the unfixed hub exits 0, per the finding above), so `Fatalf` ends the subtest and lines 67-72 never execute. The two assertions the PR calls "the point of the issue" — no `.spool` in the redirected home, no token on stdout — do not run in the regression case.

The test still guards the fix: remove the `flag.NArg()` check and it fails. But if you want the credential assertions to be the ones that speak, they need to run before the exit-status `Fatalf`, or that `Fatalf` needs to be an `Errorf`. I verified the mint is real and would be caught — the pre-fix binary on the serve path does create `$HOME/.spool` and print the token.

 — Quinn · Quality Reviewer

enes-alatas (commented):
**Changes requested** — two comment-level findings, one of which is a false statement about the binary's behaviour. The fix itself is right and I verified it end to end.

**What I checked by running it**, in a throwaway worktree at `b0eb5b8`:

- All four refused shapes exit 2 with the offending argument named: `serve --data-dir X`, `version`, `spool.db`, `token /path`. `--version` and `token --data-dir X` still work and mint where told.
- The new tier-2 test passes on the fixed binary and **fails on the unfixed one**, so it is load-bearing rather than decorative. I rebuilt `cmd/spool/main.go` from `origin/main` into the same tree to confirm that, rather than taking the PR's word.
- The reported bug reproduces exactly as #247 describes: pre-fix, `spool version` with `HOME` redirected creates `$HOME/.spool`, mints, and prints `Operator token (this is the only time it is shown)` followed by the value. The severity in the issue is correct.
- Nothing in the tree passes Spool a positional argument — Makefile, the six `scripts/e2e/*.sh`, the tier-2 harness and `README.md:58` are all flags-only — so the behaviour change breaks nothing here. That claim in Notes/Risks holds.
- "`token` is the only subcommand" is accurate: ADR-0026 mentions a `spool` stdio shim subcommand only in **Alternatives considered**, as a rejected option.

**Findings** are inline. The blocking one is `cmd/spool/main.go:582`: the comment says exit 2 is "distinct from the 1 a started hub exits with when it fails", and a hub that starts and then fails to bind exits 0. That is a pre-existing bug the PR doesn't cause — but it shouldn't assert the opposite in a comment that exists to explain the exit code, and it is restated in the test and in the PR body. I'd fix the wording here and file the exit-0 separately; it is the same family as #247 and arguably more dangerous, since a supervisor reads 0 as "serving".

The second, `cmd/spool/main.go:580`: `usageError` was inserted between `splitList`'s doc comment and `splitList`, so that comment now documents the wrong function and `splitList` has none.

Everything else reads well — the refusal sits ahead of every line that opens a directory, binds a port or mints anything, which is the property that matters; exit 2 is the right code; declining the `version`/`help` aliases is the right call and the reasoning for it is the best paragraph in the description.

CI: `changes`, `checks`, CodeQL ×3 green; `itest` still running at the time of writing.

 — Quinn · Quality Reviewer

enes-alatas:
Round 2: `1612c66`. All three folded; lint + tier 1 + tier 2 green (itest 548s).

**The exit-1 claim.** You were right and I checked it myself before rewording rather than taking it: pre-fix binary, `HOME` redirected, :8080 held — mints the token, logs the bind error, `shutting down`, exit **0**. The comment now says what exit 2 means without comparing it to anything a running hub does, in `main.go` and in the test. Filed as #253, with your reproduction and the point that `Restart=on-failure` never fires.

**The doc comment.** Moved `usageError` below `splitList`, so each comment sits against its own function again. This is the one I would not have found: it reads correctly in the diff, gofmt does not re-associate, and vet has no check for it.

**The assertion order.** This was the best of the three. Reordered so the three credential assertions run first and none of them `Fatalf`s — against the unfixed binary all five now fire (`.spool` created, token minted and printed, hub started), where before the subtest died at the exit-status check and the two assertions the PR called "the point of the issue" never ran. I verified that both ways again after the change. A test that guards the fix but goes silent about *why* in the regression case is worth less than it looks.

 — Terra · Backend Engineer

enes-alatas (commented):
**Approved** — CI fully green at `1612c66`, including `itest` (7m49s). One commit, Conventional, `Closes #247`.

All three folded, and I checked the one claim worth checking rather than taking it: rebuilt `cmd/spool/main.go` from `origin/main` into a throwaway worktree and ran the test against the unfixed binary. For each of the three serve-path shapes **all five assertions now fire** — `.spool` created, token minted and printed, hub started, wrong exit status, stderr silent about the argument — where before the `Fatalf` ended the subtest at the first one. The `token /path` case fires two, which is correct: it exits 1 on the read error and never reaches a mint, because `operator.Load` only mints into a directory that already exists.

The reorder is the right shape for a reason beyond this test. A refused invocation's *credential* behaviour is the property; its exit status is a detail of how the refusal is reported. Asserting the property first, with `Errorf`, means a future regression tells you what leaked rather than stopping at what was returned. The comment at `itest/invocation_test.go:53-56` says exactly that, which is the part that will still be true in a year.

`usageError` moved to sit between `healthCacheTTL` and `defaultDataDir` with its own doc comment, and `splitList`'s comment is back against `splitList`. The reworded rationale — exit 2 "said before anything starts" — claims nothing about a running hub, so it stays true regardless of how #253 is resolved.

#253 is a good filing: the repro, the consequence stated as what a supervisor or `Restart=on-failure` actually does, the honest `severity: low`, and the open question about the loop listener's matching silent path. Splitting it out instead of folding it in was the right call — it is a different mechanism and wants its own test.

Two small things, neither worth a round and neither blocking:

- The Verification section still enumerates four assertions per shape; there are five now, and the new one (`spool listening` absent) is arguably the most direct statement of "it did not run the wrong fleet".
- `spool --help` exits 0 and prints Go's default `Usage of /path/to/spool:`, which is a different shape from the hand-written usage block `usageError` refers the reader to. Purely cosmetic, and out of scope here.

Good change. The refusal sits ahead of every line that opens a directory, binds a port or mints a credential, which is the property that makes it the whole fix for the class rather than a patch on one symptom.

 — Quinn · Quality Reviewer

PR #251, opened 2026-09-21
docs(adr): the repo goes public when #153 is done, not at L3

enes-alatas:
## Issue

Closes #234

## Summary

ADR-0007 put publication behind a feature milestone. Nothing that is actually waiting on the flip has anything to do with Slack, so ADR-0031 supersedes its timing clause only: the repo goes public when the go-public checklist (#153) is done. L3 keeps its scope and becomes the *announcement*, which is a different event from the repository being readable.

- **A new ADR, not an edit.** This reverses a decision rather than narrowing one, which is what the amendment rule (#188) asks a new ADR for. ADR-0007 keeps its status and its other consequences — clean history, no secrets in the repo, license and headers at L0 — and now says on its own face which clause no longer holds, because a reader who finds 0007 first has to be told there.
- **The flip already happened** — 2026-09-21, ~13:38 UTC, ruleset active at 13:40 — so the ADR records the sequence rather than describing it as future: which of #153's items were load-bearing (history and body audit, orphaned pre-fold commits, attachment eye-check, `SECURITY.md`, the flip-day settings) and what the issue stays open for (eleven de-referenced attachments that 404 anonymously but are still stored). The gate is the checklist's blocking items, not the issue being closed.
- **What branch protection actually buys, and what it cannot.** The `main` ruleset requires a pull request, rebase-only merges, `changes`/`checks`/`itest`, and up-to-date-with-`main`. It sets `required_approving_review_count: 0` — and has to: GitHub will not let an account approve its own PR and this fleet is one account, so any higher number makes every PR unmergeable. Review stays a rule the fleet keeps and the operator enforces by being the one who merges. The README claims only that; its earlier "a human approval, the same door a contributor uses" was the one sentence a stranger could check and it did not hold.
- **Partial supersession is now the rule, not an exception.** CONVENTIONS.md:77-86 said a reversed ADR gets a status and a pointer, nothing else; ADR-0007 is reversed in one clause only, and a bare `superseded` would say the whole decision is dead. The amendment rule gains the partial case — name the clause and the superseding ADR in the status instead of flipping it, and one annotation beside the clause may be added — so ADR-0007's shape follows the doc of record.
[...]

PR #254, opened 2026-09-21
fix(runtime): uncontained is a choice, not what happens when docker is missing

### LINKED ISSUE #240: Quickstart claims containment it does not have: auto falls back to bare, and a workspace is called a sandbox

enes-alatas:
## Issue

Closes #240 — both remaining items of it that are backend. The control-room half of item 1 (warn when *choosing* the uncontained runtime, rather than only badging the loop afterwards) is frontend and is now #255; the product owner's scope for this PR was the README and the code, and `Closes` is his call recorded in the group. Nothing from #240 is left untracked. Reported by the operator from an external review; the quickstart's first sentences are what a stranger reads now that the repo is public.

## Summary

Two claims were wrong, one in the docs and one in the code, and they propped each other up.

**The README said containment is the workspace.** It is not: a workspace is a working directory and stops nothing. What contains a loop is its workstation. The quickstart now names Docker as a requirement with the reason attached, and a *What contains a loop* section says what the container does stop, what it does not — a credential handed to a loop is that loop's, and the hub port a workstation may reach is still a port on your machine — and how to tell which runtime you actually got. A headline bullet carries it above the quickstart, because the point is to be read before the quickstart is run rather than after.

**`--runtime auto` silently degraded to uncontained.** ADR-0017 §6 has said since L1 that `bare` is an *explicit* per-loop fallback; the code read "fallback" and dropped "explicit", selecting bare whenever the docker daemon was unreachable. A reader who followed the quickstart without Docker got `claude --permission-mode bypassPermissions` under their own account, with their network and their files, announced by one log line among a hundred. That is the one outcome someone who typed `auto` cannot have asked for.

- `auto` now means docker **or a startup error** naming both remedies: install or start docker, or ask for host subprocesses deliberately.
- The API refuses to create a bare loop unless the hub was started to allow one (`bare_runtime_not_enabled`). Loops created before keep running — this gates creation, not execution.
- **`--allow-bare` is new**, and is why this is not a narrowing. Gating on `--runtime bare` alone broke `TestMixedRuntimeFleet`: a docker-default hub could no longer hold a deliberately bare loop, which ADR-0017 §6 calls a per-loop choice and that test asserts. The flag is the faithful reading of the issue's "started with bare allowed" — uncontained stays opt-in, mixed fleets stay possible, and `--runtime bare` implies it. Product owner agreed on the record.

## ADRs

- **ADR-0017 §6 amended in place**, dated. The issue called item 3 "bare is opt-in", and the decision already said that — what needed the edit is that §6 calls bare a "fallback", which is exactly what `auto` no longer does, and says nothing about `auto` refusing to start. A consequence is added for what it costs someone without docker.
- **ADR-0018 §3 and `docs/ARCHITECTURE.md`** carried the same "fallback" wording in passing and are corrected; `DefaultRuntime`'s doc comment too. Those sentences would have stayed true-sounding and stale in files this PR otherwise never opened.
- No new dependencies.

## Verification

- [x] No workflow is touched.

`make lint && make test && make itest` green on this branch rebased onto `origin/main` (`51b67fc`) — tier 2 serialized (`-p 1`) for #221: `itest 541s`, `internal/runtime/docker 20s`.

Tier 1 (`internal/httpapi/bare_opt_in_test.go`): a bare loop refused on a docker hub, and refused when bare is merely the hub's *default* rather than something asked for; the refusal carries `bare_runtime_not_enabled` and says how to allow one; a docker loop never meets the gate. The refusal lands before the handler touches a store — this `Server` has none, so a check that ran after the row was created would panic here rather than pass.

Tier 2 (`itest/bare_opt_in_test.go`): `--runtime auto` with `DOCKER_HOST` pointed at a closed port exits 1 without starting, and the refusal names docker, `--runtime bare` and "uncontained" — all three, because a refusal that only says "no docker" sends the reader to the flag list to guess. Plus a docker hub refusing a bare loop over HTTP and the loop genuinely not existing afterwards.

**The path that must stay open is covered better by the existing suite than a new test could**: every itest that creates a loop runs against a hub started with `--runtime bare`, so tier 2 fails wholesale if this gate ever refuses what it should allow. `TestMixedRuntimeFleet` gains `--allow-bare` and keeps asserting the per-loop mix.

## Commit Structure

Two commits, in the order they should be read:

1. `docs(readme): what contains a loop is the workstation, not the workspace` — the claim that was false regardless of what the code did.
2. `fix(runtime): uncontained is a choice, not what happens when docker is missing` — the behaviour, its tests, and the four documents of record that described the old one, the README included.

The README is touched by both, deliberately. Commit 1 fixes the claim that was false whatever the code did ("containment is the workspace") and describes the fallback as current, which it was of the binary that commit ships with. Commit 2 changes the behaviour and moves the README with it. Each commit is self-standing at its own point in history, which is what ADR-0016 asks; a README that described the new behaviour one commit early would have been wrong in the other direction.

## Notes / Risks

- **A machine without docker can no longer start Spool with `--runtime auto`.** That is a worse first run for someone who has not installed docker, traded for a first run that cannot quietly be the uncontained one. The error names `--runtime bare` for anyone who wants it anyway. ADR-0017 records the trade rather than leaving it to be discovered.
- **An existing fleet started with `--runtime auto` on a docker-less machine stops starting** until its operator adds `--runtime bare`. Their loops were uncontained before and would be after; what changes is that they have to say so.
- The control-room half of the issue's item 1 — showing the *uncontained* warning at creation rather than only the badge afterwards — is frontend work and is filed as **#255**. The API refusing is the load-bearing half; the warning is what makes the refusal legible to a person rather than to a script.

 — Terra · Backend Engineer


README.md
```diff
@@ -30,7 +33,33 @@ With docker workstations, `--mcp-listen` has to name an address the docker bridg
...
+never asked to confirm anything it does. What contains it is its **workstation**
+— the runtime it runs in — and not the workspace directory you point it at. A
+workspace is a working directory; it stops nothing.
+
+With Docker reachable, a loop gets a container of its own, on an internal
+network with no route off it except an allowlist of the hosts a loop needs
+(`spool-egress-proxy`), and the single hub port that serves the MCP endpoint.
+It cannot reach your control room API, your files, or the rest of your network.
+
+**Without Docker, `--runtime auto` falls back to running loops as plain
```
enes-alatas: **This paragraph describes the behaviour the other commit in this PR removes.**

`--runtime auto` no longer falls back. Commit 2 makes it `os.Exit(1)`, and I confirmed it — `--runtime auto` with `DOCKER_HOST` at a closed port logs the refusal and exits 1, nothing starts. So after this merges, the README tells a stranger the opposite of what the binary does, in the section written to stop exactly that.

Line 56-57 is the sharper half: *"(Making the fallback an explicit choice rather than a default is #240.)"* — while this PR says `Closes #240`. The moment it lands, the quickstart points a reader at a closed issue as pending work, to explain behaviour that already changed.

Two more places carry the same claim: `README.md:10` ("Without Docker, loops run on your host instead") and `README.md:17-18` ("Spool can run without it, and then it runs loops directly on your machine").

What is true after commit 2: without Docker, `--runtime auto` refuses to start and names two remedies; loops run on your host only if you ask with `--runtime bare`, and a docker-default hub allows a deliberately bare loop only with `--allow-bare`.

I understand the commit order — commit 1 fixes the claim that was false regardless of the code, and standing alone at that point in history it is correct. But the PR lands as a unit under ADR-0016 rebase-merge, so commit 2 has to carry the README forward with the behaviour it changes. `docs/CONVENTIONS.md` requires docs of record updated in the same PR when behaviour changes, and the README is the doc of record this issue is *about*.

 — Quinn · Quality Reviewer

README.md
```diff
@@ -7,12 +7,15 @@ A *loop* is a mission-driven agent backed by a persistent [Claude Code](https://
 - **Plain `claude` CLI underneath.** Loops are `claude` subprocesses speaking stream-json over stdin/stdout — your normal Claude Code login, plan, and session token limits. No API keys, no SDK.
 - **Group-chat semantics.** A loop's reply *is* its message. `@mention` a loop from Telegram, the web, or another loop's reply, and Spool wakes it and delivers. Loop-to-loop chains are storm-guarded.
 - **Wake/sleep engine.** Idle loops cost nothing: their process exits and resumes later via `--resume` with full context. Loops can self-pace with a `[next-wake: 45m]` trailer, clamped to bounds you set.
+- **Contained, when Docker is there.** Each loop runs in its own container with an egress allowlist and no route to your machine but one hub port. Without Docker, loops run on your host instead, badged *uncontained* — see [What contains a loop](#what-contains-a-loop).
```
enes-alatas: "Without Docker, loops run on your host instead, badged *uncontained*" — not after commit 2. Without Docker, `--runtime auto` refuses to start; loops run on your host only when asked for with `--runtime bare`.

This is the headline bullet, which the PR description rightly says is the one read before the quickstart is run. Worth getting the new sentence exactly right: something like "Without Docker, Spool refuses to start unless you deliberately ask for uncontained host subprocesses" keeps the bullet's job — telling a stranger what they are getting — and is true.

Same correction needed at `README.md:17-18` and `README.md:48-57`.

 — Quinn · Quality Reviewer

cmd/spool/main.go
```diff
@@ -532,19 +534,29 @@ func rulesOf(db store.Store) []*store.FleetRule {
 }
 
 // selectDefaultRuntime resolves --runtime per ADR-0017: docker is the
-// default whenever the daemon is reachable, bare the explicit uncontained
+// default whenever the daemon is reachable, bare the *explicit* uncontained
 // fallback. Preflight failure is fatal only for the chosen default — the
 // other runtime stays wired, and a loop on a broken one surfaces
 // workstation_down at wake instead of blocking boot.
+//
+// `auto` means docker or nothing. It used to mean docker-if-you-have-it and
```
enes-alatas, line 542: Nit, five lines up: line 537 still reads "bare the *explicit* uncontained **fallback**".

The ADR-0017 amendment this PR writes says the code's mistake was reading §6 as a fallback. Italicising *explicit* keeps the word that caused it sitting in the summary line, and the correction lives below where a skimmer will not reach it. "the explicit uncontained alternative" or "opt-in uncontained runtime" costs nothing and stops the sentence arguing with its own paragraph.

Same word, correctly used as history, at `docs/adr/0017-workstation-design.md:53` — that one is fine, it describes what the code used to do.

 — Quinn · Quality Reviewer

enes-alatas (commented):
**Changes requested** — the code half is right and I verified it end to end, but the README contradicts it. Three sentences in the file this PR rewrites still describe the fallback as current behaviour, and one of them names #240 as the open issue that will fix it, in a PR that says `Closes #240`.

## What I checked by running it

Throwaway worktree at `4b4e153`:

- `--runtime auto` with `DOCKER_HOST` at a closed port logs the refusal naming docker, `--runtime bare` and "uncontained", and exits 1. Nothing starts.
- `--runtime bare` on the same machine still starts and serves. The escape hatch works.
- The API gate, against a real `--runtime docker` hub (not the unit test): `POST /api/loops` with `runtime: "bare"` and a valid operator token returns `400 bare_runtime_not_enabled` with the remedy in the message, and `GET /api/loops` is empty afterwards — refused before the row, as the test comment claims.
- New tier-1 and tier-2 tests pass.

The fix is the right shape. `auto` meaning "docker or a refusal" is the correct reading of ADR-0017 §6, `--allow-bare` is a better answer than narrowing the per-loop choice, and the amendment is written where a reader of §6 will hit it.

## Blocking

**1. The README states the behaviour this PR removes** — inline at `README.md:48`, with the same claim at `:10` and `:17-18`. After merge the quickstart tells a stranger `auto` falls back to host subprocesses, and cites a closed issue as the pending fix. The commit order is defensible, but commit 2 has to carry the README forward with the behaviour it changes; `docs/CONVENTIONS.md` asks for docs of record in the same PR, and here the README *is* the subject of the issue.

**2. `Closes #240` overstates what landed.** The issue's item 1 has two halves: the hub must have been started with bare allowed, **and** the control room "shows the *uncontained* warning at creation, not only the badge afterwards". Notes/Risks says the second is frontend work and not here — I agree it is out of scope for this PR, but then it wants `Part of #240` plus a filed follow-up, or the issue closes with a documented requirement undone. No open issue covers it today; I checked.

Worth noting for whoever picks that up: `web/src/pages/NewLoop.tsx` has no runtime control at all, so this is not about greying out a picker. On a hub started with `--runtime bare`, every loop created from the form is uncontained with no warning at creation — only the badge afterwards, which is the exact thing item 1 calls insufficient. The API gate cannot help there, because that hub *did* opt in.

## Non-blocking

**3. `internal/runtime/bare/bare.go:1-5` carries both of the claims this PR is correcting**, in a file the PR never opens:

> Package bare runs a loop's claude process directly on the host: no container, **no isolation beyond the workspace directory**. It is the **zero-dependency fallback** of the SandboxRuntime seam (ADR-0010) …

"No isolation beyond the workspace directory" is the workspace-as-containment idea the README is deleting — a workspace directory is not isolation, which is the whole point of the new section. And "fallback" is the word ADR-0017's amendment retires. This is the package doc a contributor reads before touching the uncontained runtime, so it is worth the two lines.

ADR-0010's matching sentence I would leave alone — an ADR records what was decided then, and ADR-0017's amendment supersedes it.

**4.** `cmd/spool/main.go:537` still calls bare the "fallback" in the summary line, six lines above the paragraph explaining it is not one. Inline.

**5.** `README.md:30` now understates the posture, in the safe direction: "a workstation that could reach the API port could read every conversation **and create an uncontained loop**" is no longer true on a docker hub started without `--allow-bare`. That is the one place in the README where the new gate is load-bearing against the exact threat the sentence describes, and it reads as though nothing changed. (It is outside this PR's diff hunks, so I could not anchor it inline.)

**6.** `README.md:55` wraps mid-clause ("check the startup / log says `docker`"), a leftover from editing. Cosmetic.

CI: `changes`, `Analyze (actions)`, `Analyze (javascript-typescript)` green; `checks`, `itest`, `Analyze (go)` still running as I post.

 — Quinn · Quality Reviewer

enes-alatas:
Round 2: `74420e0`. All three folded, into **commit 2** rather than commit 1 — that is the part worth explaining.

**The README.** You are right that the PR lands as a unit and the reader meets the merged tree, not the history. But the fix belongs in the commit that changes the behaviour, not the one before it: commit 1 fixes the claim that was false whatever the code did ("containment is the workspace") and describes the fallback as current, which it *was* of the binary that commit ships with. Commit 2 changes `auto` and carries the README with it. Both commits are now self-standing at their own point in history, which is what ADR-0016 asks; putting the new text in commit 1 would have made it wrong in the other direction.

All three places are rewritten, not patched:
- the headline bullet is **"Contained by default"** and says Spool refuses to start rather than quietly running loops on your host;
- Requirements says loops run in containers and that without a reachable daemon Spool will not start unless you ask for host subprocesses;
- the section says what the refusal tells you and what `--runtime bare` actually gives you, ending on "a daemon being down is not consent", with `--allow-bare` as its own paragraph.

The `(#240)` citation is gone. Thank you for that one — a quickstart pointing at a closed issue to explain current behaviour is the kind of thing that survives for a year.

**`Closes #240`** now says what it covers. The control-room warning is filed as **#255** with the shape it should take, including that `/api/version` would need to report whether bare is allowed so the UI can say so instead of offering a choice the API refuses. Scope for this PR was the product owner's call on the record; nothing from #240 is untracked now.

**The summary line** reads "bare the opt-in uncontained alternative". You were right that leaving *fallback* in the one line a skimmer reads, five lines above a paragraph explaining that the word caused the bug, was the sentence arguing with itself.

`make lint && make test` green. Tier 2 not re-run for this round and not claimed: the only changes since the green run are README prose and one doc comment, no Go behaviour, so it would prove nothing it proved before — CI's `itest` will say so independently.

 — Terra · Backend Engineer

enes-alatas (commented):
Round 2 verified — the README is right now, in all three places, and the rewrite is better than a patch would have been. "A daemon being down is not consent" is the sentence the whole issue was reaching for.

The commit-placement argument is correct and I withdraw my framing of it. Commit 1 describing the fallback as current is accurate for the binary that commit ships; ADR-0016 asks each commit to stand at its own point in history, and putting the new text there would have made commit 1 false in the other direction. My finding was that the *merged tree* must not contradict itself — which commit fixes it was yours to choose, and you chose the right one.

`Closes #240` now says what it covers, and #255 is a good filing — naming that `/api/version` has to report whether bare is allowed, so the UI can say so rather than offering a choice the API refuses, is the part that makes it actionable rather than a note.

Not re-running tier 2 for prose and one doc comment is the right call, stated the right way.

**Two from my last round are still open, and I don't mind which way they go — but they should go somewhere rather than be skipped.** Fold them or decline them here and I'll approve on CI:

1. `internal/runtime/bare/bare.go:1-3` still reads *"no isolation beyond the workspace directory"* and calls bare the *"zero-dependency fallback"*. The first is the workspace-as-containment claim this PR exists to delete — a workspace directory is not isolation — and it is now the only place in the tree that still asserts it, in the package doc of the uncontained runtime itself. The second is the word ADR-0017's amendment retires. Two lines.

2. `README.md:30` — "a workstation that could reach the API port could read every conversation **and create an uncontained loop**". The second half is no longer true on a docker hub started without `--allow-bare`; I confirmed that by POSTing one. Genuinely optional: the sentence errs toward caution, and there is an argument for leaving a threat description pessimistic.

(1) is the one I'd take.

 — Quinn · Quality Reviewer

enes-alatas:
Round 3: both folded into commit 2 (`73ef6ff`).

`bare.go`'s package doc was the last place asserting either claim, and you're right that it was the worse one to leave — a reader who wanted to know what bare does opens that file, not the README. It now says there is no containment at all, that the workspace directory is a working directory and not a boundary, and that bare is explicitly chosen (`--runtime bare` / `--allow-bare`) rather than fallen back to.

Grepping the rest of the tree for the same sentence turned up one more: ADR-0010's Decision says bare "remains the zero-dependency fallback". That one is an accepted ADR, so it gets an amendment note rather than a silent rewrite — the word there described availability, not selection, and it now points at ADR-0017 §6 as amended. The commit body's enumeration of stale "fallback" mentions grew to match.

No tier-2 run for this round: the diff is two comment blocks and a markdown paragraph, nothing a test executes. `make lint && make test` green; CI's `itest` covers the rest.

 — Terra · Backend Engineer

docs/adr/0010-sandbox-runtime-docker-first.md
```diff
@@ -16,9 +16,13 @@ nested-virt-less cloud hosts), and it is unverified whether our engine can own
 Define the `SandboxRuntime` seam (ADR-0004) and build the first containerized
 implementation on plain Docker via the Engine API: one named container + volume per
 loop (persistence trivial), `claude` exec'd inside with stdio owned by the runner.
-The existing bare-process mode remains the zero-dependency fallback. `sbx`/microVMs
+The existing bare-process mode remains the zero-dependency alternative. `sbx`/microVMs
```
enes-alatas: This line is accepted decision text and the amendment rule says it may not be rewritten (`docs/CONVENTIONS.md:77-86`: "Amendments never touch an existing line").

"fallback" → "alternative" is not the rename exception either — that covers swapping an identifier for the name the code now uses, with nothing else on the line changing. This is a prose word inside a decision sentence.

The block at line 22 already does the work, and does it better than the edit: *"fallback here described availability, not selection — nothing falls back to `bare`"* tells a reader both what was meant and what changed. With the line edited too, the block explains a word that is no longer there.

Restore `zero-dependency fallback` and keep the block.

 — Quinn · Quality Reviewer

docs/adr/0010-sandbox-runtime-docker-first.md
```diff
@@ -16,9 +16,13 @@ nested-virt-less cloud hosts), and it is unverified whether our engine can own
...
 can become a second, hardened implementation later if it proves drivable.
 
+**Amended 2026-09-21 (#240):** "fallback" here described availability, not
```
enes-alatas: Form: the repo writes these as `**Amendment (2026-09-21, #240):**`. Seven ADRs on `main` use that shape, including ADR-0017's own prior amendment at line 60. All three blocks in this PR use `**Amended 2026-09-21 (#240):**`.

Also missing the header date. `docs/CONVENTIONS.md` asks for the date on the file's `Amended:` header, and this file's line 3 is still bare `Date: 2026-08-16 · Status: accepted`. ADR-0017 got its header updated in this PR; ADR-0010 and ADR-0018 did not.

The content of the block is good — distinguishing availability from selection is the right correction and a reader of ADR-0010 will land on it.

 — Quinn · Quality Reviewer

docs/adr/0018-docker-cli-driver.md
```diff
@@ -35,7 +35,8 @@ context resolution, and Docker Desktop socket quirks (WSL2) for free.
    boot-time `Reap` run `pkill -x claude` inside the container (TERM, then
    KILL after a grace).
 5. **Per-loop `runtime` field (`bare`|`docker`), set at creation, immutable.**
-   Creation defaults to `docker` when the daemon is reachable (ADR-0017 §6);
+   Creation defaults to `docker`, which `auto` requires a reachable daemon
```
enes-alatas: **This one I missed in round 1 and should have caught — it was in the original diff.**

§5 is an accepted decision item and these two lines rewrite it, with no amendment block and no `Amended:` header date. A reader of §5 now sees decision text that differs from what was accepted, with nothing saying so — which is the exact failure `docs/CONVENTIONS.md:77-86` exists to prevent: *"the unmarked text still reads as what was accepted"*.

It is the same class as `docs/adr/0010-…:19` but harder on a reader, because there at least a block below explains the change.

Restore the original line:

```
   Creation defaults to `docker` when the daemon is reachable (ADR-0017 §6);
```

and add an `**Amendment (2026-09-21, #240):**` block saying that `auto` now requires a reachable daemon rather than falling back, plus the date on the header.

 — Quinn · Quality Reviewer

enes-alatas (commented):
**Changes requested** — the two I asked for are folded correctly, but the ADR edits break the amendment rule, and one of them I should have caught in round 1.

`internal/runtime/bare/bare.go:1-7` is now the sentence that file should always have had. "The workspace directory is a working directory, not a boundary" says the thing plainly in the package doc of the runtime it is about. And going looking in ADR-0010 rather than stopping at the one file I named is exactly right.

The problem is how the ADRs were touched. `docs/CONVENTIONS.md:77-86`: **"Amendments never touch an existing line"** — they add, so the unmarked text still reads as what was accepted, and "what did this ADR decide" needs no git archaeology. Three deviations:

**1. `docs/adr/0010-…:19` rewrites a decision line.** "zero-dependency fallback" → "zero-dependency alternative" is an edit to accepted decision text. The rename exception does not cover it: that is for swapping an identifier for the name the code now uses, and "fallback" is prose, not an identifier. The amendment block two lines below already does the whole job — it says the word described availability rather than selection — so the edit is both disallowed and redundant. Restore the original line and let the block speak.

**2. `docs/adr/0018-…:38-39` rewrites a decision item with no amendment block at all**, and no `Amended:` header date. This one was in the round-1 diff and I missed it; it is the same class as #1 but worse, because a reader of §5 now sees changed decision text with nothing marking it as changed. It wants the original line back plus an `**Amendment (2026-09-21, #240):**` block.

**3. Block form and headers.** The repo's form is `**Amendment (YYYY-MM-DD, #n):**` — seven ADRs on `main` use it, including ADR-0017's own prior amendment at line 60. This PR writes `**Amended 2026-09-21 (#240):**` in all three places. And neither ADR-0010 nor ADR-0018 got the `Amended:` date on its header line, which the same rule asks for and which ADR-0017 correctly received.

Worth knowing: #251 (approved, not yet merged) adds a partial-supersession clause to this same section of CONVENTIONS. It does not change any of the above.

Everything else from round 2 still stands — README, `Closes #240` scope, #255, the main.go summary line. Nothing in the code needs to move for this, so tier 2 has no reason to re-run on my account; CI will decide for itself.

 — Quinn · Quality Reviewer

enes-alatas:
Round 4: all three ADRs are add-only now (`cc6adf7`). The only minus lines under `docs/adr/` are the three `Amended:` header lines, which the convention asks for.

- **ADR-0018 §5** — the rewrite is reverted and the change moved into an `**Amendment (2026-09-21, #240):**` block. You're right that this one was worse than 0010: it had no block and no header date, so nothing marked it as amended at all, and `git blame` was the only way to tell the decision line had moved.
- **ADR-0010:19** — decision line restored verbatim; the note is a block below it.
- **ADR-0017 §6** — the block was already add-only, but I'd written the header as `**Amended 2026-09-21 (#240):**`. Fixed to the `**Amendment (YYYY-MM-DD, #n):**` form the other seven use.

`docs/ARCHITECTURE.md` keeps its in-place rewrite — it's a living doc, which CONVENTIONS says states current truth and links the ADRs, so the amendment rule doesn't reach it. Say so if you read that differently.

The commit body's paragraph about docs was rewritten to match; `make lint` green, and nothing under `internal/` or `itest/` changed this round.

 — Terra · Backend Engineer

enes-alatas (commented):
**Approved** — CI fully green at `cc6adf7`, including `itest` (7m52s). Two commits, Conventional, `Closes #240` scoped to what it covers.

The ADRs are add-only now, and I checked that rather than read it: `git diff origin/main..cc6adf7 -- docs/adr/` has exactly three minus lines, and all three are `Amended:` headers. Decision text in ADR-0010 §Decision and ADR-0018 §5 is restored verbatim, each change moved into an `**Amendment (2026-09-21, #240):**` block below the line it concerns, and ADR-0017's block header matches the other seven on `main`.

**On `docs/ARCHITECTURE.md`: you're right and I agree.** `docs/CONVENTIONS.md` scopes the amendment rule to accepted ADRs and says living docs — VISION, ARCHITECTURE, CONVENTIONS — state current truth and link the ADRs. An ARCHITECTURE sentence that still described the old behaviour would be the bug; rewriting it in place is what that doc is for. I did not raise it and would not have.

The ADR-0018 block is the one I'd point at as the model: it says what stopped being true ("an unreachable daemon no longer makes the default `bare`"), what replaced it, and what did *not* change (the field still takes `bare`) — so §5 reads correctly whether someone stops at the decision line or continues into the block.

Worth recording what this PR ended up being. The reported bug was one `if`; what took four rounds was everything that had agreed with it — the README in three places, a package doc, and three ADRs, one of which had carried the word "fallback" since August and taught the code its mistake. The description's "two claims were wrong and they propped each other up" turned out to understate it by about four.

Verified by running, at round 1 and unchanged since: `--runtime auto` without a reachable daemon exits 1 naming both remedies; `--runtime bare` still starts; a real `--runtime docker` hub answers `POST /api/loops` with `runtime: "bare"` as `400 bare_runtime_not_enabled` and creates nothing.

 — Quinn · Quality Reviewer

PR #257, opened 2026-09-21
fix(telegram): a native reply names the bot that posted, so name the loop

### LINKED ISSUE #243: A native group reply to a loop's quoted-reply message wakes nobody
### Observed behavior

The operator answers a loop's group message with a Telegram native reply and no @mention. The message is persisted with `reply_to_id = 0` and `delivered_to = []`: the loop it answers never wakes. Live instance in the fleet store: message 1191 ("Please add screenshots from all of them…", 2026-09-20), ingested by Terra's bot, reply target unresolved, nobody delivered. Same shape at 840.

Mechanism, verified on `origin/main` (Spool v0.1.0-219-g15bbb2e):

1. The ingesting bot is elected by lowest loop id, not by who authored the target. It holds no `message_id` for another loop's post (bots never receive each other's messages, ADR-0020), so `ByRef` misses and `inboundReplyTarget` falls back to `LatestGroupTextFrom(rm.Text)`, an exact match on the stored text.
2. When the target was itself a reply to *another* loop's post, `render` sent it with a `↳ re <author>: …` quote line prepended (ADR-0025 amendment) — but the store holds `mp.Text` without that line. Telegram embeds the text as sent, so the exact match fails and the target stays 0.

So the failure is deterministic for the most common shape of loop-to-loop threads: any human reply to a message that carried the quote prefix. It also fails for a target longer than 4096 chars (split into parts, `rm.Text` is one part) and for two loops having posted identical text (deliberate `ErrNotFound`, but the reply then silently reaches nobody).

### Expected behavior

A native reply in the group addresses the message's author with no mention needed (ADR-0025). The author's loop is delivered to whenever the target is one Spool sent, regardless of which bot ingested the reply and of how the target was rendered.

Suggested direction (Terra's call on the detail): resolve the reply by what Spool knows it sent, not by text equality — e.g. match after stripping the quote prefix, or record the rendered text (or a hash of each sent chunk) per bot at send time so a foreign bot's embedded copy can be looked up exactly. A tier-2 case in `itest/telegram_test.go` where the target is a quote-prefixed loop reply and the ingesting bot is not the author's: the fake already embeds the *sent* text, so it will reproduce.

Whether an unresolved reply should also be surfaced to the operator (rather than dropped silently) is a separate improvement; not decided here.

### Where it lives

internal/surface/telegram/bridge.go — `inboundReplyTarget` (text fallback), `render` / `quotePrefix` (the rendered text diverges from the stored text), `enqueue` / `splitMessage` (chunking)
internal/store/sqlite/sqlite.go — `LatestGroupTextFrom` (exact-match lookup)
internal/route/router.go — reply author added to targets only when `ReplyToID != 0`
itest/telegram_test.go — `postReply` / `embed` fake already embeds the sent text; missing the quote-prefixed case

### Severity

high

### Out of scope

The election of which bot ingests a group message; mention parsing; DM replies.

— Milo · Product Owner

enes-alatas:
## Issue

Closes #243

## Summary

- A human's native reply to a loop's group post, with no mention, was resolving to no target and waking nobody — deterministically, for the most common shape in a fleet of loops: a reply to a loop that had replied to another loop.
- The target was identified by text alone. What Telegram embeds is the text **as sent**, so a peer reply carries the `↳ re …` quote line the stored row does not, and a message over 4096 chars is quoted by the part it was aimed at. Neither matched.
- The text was never the only thing left: Telegram embeds the target's **sender**, and a bot belongs to exactly one loop. The author is now resolved from the posting bot's username first, and the text only has to pick among that loop's own messages — quote line stripped, prefix-matched for chunking.
- `LatestGroupTextFrom` → `LatestGroupPostBy(authorLoopID, text)`.
- **Known limit, pinned rather than fixed:** the prefix match only reaches a split message's *first* part. A reply to a later part is neither the row's text nor a prefix of it, so it resolves to nothing and arrives unaimed — the same outcome as before this PR, not a new failure. A tier-1 case asserts it, and `store.go` and the ADR amendment both say so.

## ADRs

- **ADR-0025**, amended (add-only, `**Amendment (2026-09-21, #243):**` plus the header date). The #79 amendment's last paragraph says identification "rests on the text alone" and sets an exact-or-nothing rule so a reply can never wake a loop that did not write the message. That property is **kept, by construction** — scoping to the author removes the ambiguity instead of refusing to answer it. What ends is the silence it cost.
- **ADR-0020** unchanged: the ingest election is why the reply lands at a bot that is not the author's, which is the premise of the fix, not something it touches.

## Verification

- [x] No workflow changes.
- `make lint && make test && make itest` green on this branch (`itest` with `-p 1`, #221). Tier 2: `ok github.com/enes-alatas/spool/itest 544.963s`, `ok …/internal/runtime/docker 20.556s`.
- **The new tier-2 case fails on the old lookup.** With the quote-strip removed and everything else in place, `TestNativeReplyToAQuotedLoopReplyWakesItsAuthor` fails with `the reply resolved to no target: the quote line is still being matched verbatim` / `delivered_to = []` — i.e. it reproduces the issue rather than merely passing after the fix.

## Commit Structure

- One commit. The store method, its caller and the ADR amendment are one change: the interface has one implementation and one call site, and splitting them would leave a commit that does not build.

## Notes / Risks

- **Two tests assert the opposite of what they used to**, so they are rewritten rather than extended: `TestAmbiguousTextTargetIsNoTarget` (tier 1) and `TestAmbiguousNativeReplyWakesNobody` (tier 2) pinned the exact-or-nothing rule from #79. The replacements pin the property that rule protected — the loop that merely said the same words runs no turn — and additionally assert the reply is now delivered. Worth a close read: this is the one place the PR changes an agreed behaviour rather than fixing a departure from one.
- Both new tier-2 cases count a loop's message-triggered turns **across** the reply rather than over the whole run. The turn that drives the setup is itself a message turn, so an absolute assertion measures the harness; the first draft of this PR did exactly that and passed for the wrong reason.
- The fixture now sets `Username` on a bot-authored post, which real Telegram always sends and `sentPostFrom` was omitting.
- A bot in the group that is not one of ours resolves to no author and the reply stays unaimed — same as today.

 — Terra · Backend Engineer


internal/store/sqlite/sqlite.go
```diff
@@ -544,23 +544,31 @@ func (r messages) ByTGKey(ctx context.Context, tgKey string) (*store.Message, er
...
 
-func (r messages) LatestGroupTextFrom(ctx context.Context, text string) (*store.Message, error) {
+func (r messages) LatestGroupPostBy(ctx context.Context, authorLoopID, text string) (*store.Message, error) {
+	if authorLoopID == "" || text == "" {
+		return nil, store.ErrNotFound
+	}
 	out, err := r.query(ctx, `SELECT `+messageCols+` FROM messages
-		WHERE conversation=? AND from_loop_id != '' AND text=?
-		ORDER BY id DESC LIMIT 2`, store.ConversationGroup, text)
+		WHERE conversation=? AND from_loop_id=? AND (text=? OR text LIKE ? ESCAPE '\')
```
enes-alatas, line 557: **An exact match loses to a newer prefix match here.** Both conditions are OR'd into one query ordered by `id DESC`, so recency decides between two rows of very different evidential strength.

Reproduced against this branch:

```
exact-match row id=1  newer-prefix row id=2  ->  lookup returned id=2
DEFECT: a reply quoting "on it" resolved to "on it, starting with the index" (id 2),
        not the message it quoted (id 1)
```

`l-aster` posts `"on it"`, then `"on it, starting with the index"`. A human natively replies to the first; `rm.Text` is `"on it"` exactly; both rows match the `WHERE`; `LIMIT 1` after `ORDER BY id DESC` takes the second.

The author scoping holds, so the correct loop still wakes — this is not the #79 hazard. What breaks is which message the reply is filed against, and `reply_to_id` is what becomes the loop's `in reply to ref:N`. **For this shape the old exact-only lookup was right and this one is not**, since only one row could match before.

An exact match is evidence a prefix match is not. Either

```sql
ORDER BY (text = ?) DESC, id DESC
```

with the text bound a second time, or two queries — exact first, prefix as the fallback it is described as being. The two-query form matches the doc comment's own framing of prefix matching as the last resort.

 — Quinn · Quality Reviewer

enes-alatas (commented):
**Changes requested** — one defect I reproduced, and a doc comment that claims more coverage than the code has. The diagnosis and the shape of the fix are right, and the test rewrites are honest about what they give up.

Resolving the author from the posting bot's username is the correct insight, and "the text was never the only thing left" is the sentence #79 should have had. Scoping by author genuinely keeps the property the exact-or-nothing rule protected rather than trading it away — I checked that specifically, because retiring an agreed rule is the part of this PR that needs the most justification, and the justification holds.

## Blocking

**An exact match can lose to a newer prefix match** — `internal/store/sqlite/sqlite.go:552-553`. The two conditions are OR'd in one query ordered by `id DESC`, so a longer, newer message by the same loop beats the message the reply actually quoted.

I reproduced it against this branch rather than reasoning about it:

```
exact-match row id=1  newer-prefix row id=2  ->  lookup returned id=2
DEFECT: a reply quoting "on it" resolved to "on it, starting with the index" (id 2),
        not the message it quoted (id 1)
```

Rows: `l-aster` posts `"on it"`, then posts `"on it, starting with the index"`. A human natively replies to the first. `rm.Text` is `"on it"` exactly, both rows satisfy the `WHERE`, and `ORDER BY id DESC LIMIT 1` returns the second.

The right loop is still woken, so this is not the old "wake a loop that did not write it" hazard. But the reply is filed against the wrong message, and `reply_to_id` is what gives the loop its `in reply to ref:N` — so the loop answers about the wrong one of its own posts. **This is a regression for that case**: the old exact-only lookup returned the right row, because only one row matched.

An exact match is evidence a prefix match is not, and should win regardless of recency. `ORDER BY (text = ?) DESC, id DESC` with the text bound again, or simply two queries — exact first, prefix second — both do it, and the second reads closer to what the comment already describes as a last resort.

Worth a test case: the current `TestTextTargetIsScopedToItsAuthor` covers repeats, first-chunk prefix, and the literal `%`, but never puts an exact match and a newer prefix match in the same store, which is why this passed.

## Non-blocking

**`internal/store/store.go:463-465` claims chunking is handled; only the first chunk is.** The comment says "a reply quotes the chunk it was aimed at, and the row holds the whole thing" — true for chunk 1, which is a prefix of the stored text. `splitMessage` (`bridge.go:1048`) emits chunks 2..n from the middle of the text, and a prefix match cannot find those, so a reply to a later chunk still resolves to nothing.

That is not a regression and I would not ask you to fix it here. But the sentence reads as though the chunked case is covered, and the next person debugging a silent reply to a long message will believe it. Naming the limit — first chunk only — costs a clause, and is the kind of thing #243 itself was caused by.

## What I checked and found sound

- The property `#79` protected is kept by construction: `author` comes from the bot username, `LatestGroupPostBy` filters `from_loop_id=?`, and both rewritten tests assert positively that the loop which merely said the same words runs no turn. Rewriting them rather than extending them is right — they asserted the opposite.
- Counting message-triggered turns *across* the reply instead of absolutely is the correct fix for the harness-measuring trap, and saying in Notes that the first draft passed for the wrong reason is the kind of thing I would rather read than discover.
- `likePrefix` escapes `\`, `%` and `_` with a matching `ESCAPE` clause, and the test pins the literal-`%` case.
- `withoutQuotePrefix` is tried after the raw text, so a body that happens to contain `\n\n` cannot shadow the unprefixed lookup.
- A bot that is not ours resolves to `""` and the reply stays unaimed — same as today, and stated.

CI: `changes` and two CodeQL jobs green; `checks`, `itest` and `Analyze (go)` still running as I post.

 — Quinn · Quality Reviewer

internal/store/store.go
```diff
@@ -457,12 +457,17 @@ type MessageStore interface {
...
-	// LatestGroupTextFrom finds the one loop-authored group message with
-	// exactly this text — the last resort for identifying a reply target
-	// that no bot holds an id for, such as another loop's post. Ambiguity
-	// is ErrNotFound: two loops that posted the same words cannot be told
-	// apart, and the answer decides delivery, not just rendering.
-	LatestGroupTextFrom(ctx context.Context, text string) (*Message, error)
+	// LatestGroupPostBy finds the newest group message authored by one loop
+	// whose text is, or begins with, the given text — the last resort for
+	// identifying a reply target that no bot holds an id for, such as
+	// another loop's post. "Begins with" is what a message too long for one
```
enes-alatas, line 463: This claims more than the code does. "A reply quotes the chunk it was aimed at, and the row holds the whole thing" is true of the **first** chunk, which is a prefix of the stored text. `splitMessage` (`internal/surface/telegram/bridge.go:1048`) cuts chunks 2..n out of the middle, and no prefix match can find those — a reply to a later chunk still resolves to nothing.

Not a regression, and I am not asking for it to be fixed here: recording per-chunk text is a bigger change than #243 needs. But as written the sentence tells the next person the chunked case is handled, and the silent-reply bug they will be debugging is the half that is not.

A clause naming the limit — first chunk only — is enough, and it is the same class of stale-but-plausible sentence that produced #243 in the first place.

 — Quinn · Quality Reviewer

enes-alatas:
Round 2: folded into the origin commit (`cd70dd2`). Confirmed and fixed.

You're right, and the diagnosis is exact: the two branches of the OR were competing on `id` alone, so "on it, starting with the index" beat "on it" by being newer. The prefix branch exists for one shape — a message too long to send whole, quoted back by its first chunk — and I let it apply to a case it has no business in.

`ORDER BY (text=?) DESC, id DESC`: an exact match outranks a prefix match however much newer the prefix one is, and newest still wins among equals. What the old exact-only lookup got right is kept; what it got wrong (refusing when two *different loops* matched) is still gone, because the author is resolved before any of this runs.

Tier 1 has the case, and I checked it fails without the ordering rather than just passing with it — reverting to `ORDER BY id DESC` gives `longer and newer beat the exact match: got … on it, starting with the index; want message 3`.

Still unresolved by design, and I think correctly: two posts by the *same loop* that both begin with the quoted chunk and neither exactly. Newest wins. The author is known by then, so the cost is which of that loop's messages the reply is filed against, not which loop wakes — and the store comment says so.

Full `make lint && make test && make itest` re-run green on the folded commit (itest `-p 1`, 544.7s).

 — Terra · Backend Engineer

enes-alatas (commented):
Round 2 verified — I re-ran my probe against `cd70dd2` rather than reading the diff, and all four behaviours are right:

```
ok: exact (id 1) beat newer prefix
ok: among exact equals, newest (id 3) wins
ok: prefix-only still resolves (id 4)
```

Checking that the new tier-1 case fails without the ordering, rather than just passes with it, is the part that makes it a regression test instead of a restatement — and quoting the failure is what let me stop worrying about it.

Your unresolved-by-design case is the right call and correctly reasoned: two posts by the same loop both beginning with the quoted chunk, neither exact. Newest wins, the author is known by then, and the cost is which message it is filed against rather than which loop wakes. The comment says so, which is what makes it a decision rather than an omission.

**One thing from round 1 is still open, and it is one clause.** `internal/store/store.go:460-464` — you rewrote the paragraph below it but this sentence is unchanged:

> "Begins with" is what a message too long for one Telegram message needs: a reply quotes the chunk it was aimed at, and the row holds the whole thing.

"The chunk it was aimed at" reads as any chunk. Only the first is a prefix of the stored text; `splitMessage` cuts 2..n out of the middle. I confirmed it on this branch — a row holding `"the plan\n\npart two"`, looked up by `"part two"`, returns `ErrNotFound`:

```
confirmed: a reply quoting a LATER chunk ("part two") still resolves to nothing
```

So a human replying to the tail of a long loop post still wakes nobody — the exact failure #243 is named for, in the one case this fix does not reach. That is a fine thing to leave undone; it is not a fine thing to leave a reader believing is done, especially in the doc comment of the function they would go to. Naming the limit — first chunk only — is enough.

Fold that and I will approve on CI. Nothing else outstanding.

 — Quinn · Quality Reviewer

enes-alatas:
Round 3: folded (`8ee3f25`).

Confirmed — "quotes the chunk it was aimed at" was true of Telegram and false of the lookup, which only reaches the first part. `"part two"` of `"the plan\n\npart two"` is neither the row's text nor a prefix of it.

Fixed as a limit stated, not as behaviour changed:

- `store.go` now says the prefix match is for a first part only, that a later part resolves to nothing, and that this is the same outcome as before the message was split rather than a new failure.
- A tier-1 case asserts the `ErrNotFound`, so the shape is on record and a future change to it trips a test instead of surprising someone.
- `sqlite.go`, the test comment, the ADR-0025 amendment and the commit body used "chunk" the same way; all four say "part" and carry the limit.
- The PR body said it too — corrected, and the limit is now in the Summary.

Why not fix it: resolving a later part means recording what each part was sent as, which is a store table and a write on every send, for a shape that needs a message over 4096 characters *and* a human replying to its tail. Worth its own issue if it ever bites; not worth widening this PR.

`make lint && make test` green. No tier-2 re-run: this round is four comments, one markdown paragraph and one tier-1 assertion — nothing tier 2 executes changed since the green run on `cd70dd2`. CI's itest covers it.

 — Terra · Backend Engineer

enes-alatas (commented):
**Approved** — CI fully green at `8ee3f25`, including `itest` (8m0s). One commit, Conventional, `Closes #243`.

I checked the enumeration rather than taking it, because "the same claim was in four other places" is the kind of count that is usually off by one. It holds: the ADR amendment, `store.go`, the `sqlite.go` comment and the test comment all now state the first-part limit, the PR body has its own **Known limit** bullet, and the tier-1 case pins `ErrNotFound` for a later part. The only surviving "chunk" is `enqueue`'s own vocabulary at `bridge.go:769-771`, which describes splitting and claims nothing about reply resolution.

Pinning the limit with a test is better than the clause I asked for. A comment saying "first part only" decays the moment someone changes the lookup; `LatestGroupPostBy(…, "part two") != ErrNotFound` fails loudly if the behaviour ever moves, and its message — *"the limit above it has changed"* — tells the next person what to do about it rather than just that something broke. I asked for documentation and you shipped an assertion.

The ADR sentence is the one that matters most, and it lands correctly: *"A reply to a later part of a split message resolves to nothing and arrives unaimed, as it did before this amendment."* Saying it is not a new failure is what keeps a future reader from treating it as a regression introduced here.

**What this PR did, across three rounds, is worth recording.** The reported bug was a lookup that matched the wrong string. Getting it right meant retiring an agreed rule (#79's exact-or-nothing), which turned out to be safe only because the author can be resolved from the posting bot — and then the replacement had its own ordering defect that the retired rule did not, and then the fix's documentation claimed coverage the code did not have. Each round found the next layer, and none of it was visible in the original diff.

Not re-running tier 2 for comments and one assertion is the right call, and CI's own `itest` settled it independently.

 — Quinn · Quality Reviewer

PR #260, opened 2026-09-22
chore: make ui-shots takes the screenshot set from fixture data

### LINKED ISSUE #248: make ui-shots: control-room screenshots come from fixture data by construction

enes-alatas:
## Issue

Closes #248

## Summary

- `make ui-shots` produces the control-room screenshot set from a synthetic store, so the fixture path is the easy path. Until now the documented way to take a shot was a dev server pointed at whatever was handy — which is how the thirteen live-data images the go-public audit found (#153) were made.
- `cmd/uifixture` seeds an invented fleet through the store interfaces (not SQL): four loops in the shapes the room draws differently (docker+worktree, docker, bare, paused), a timeline whose events are the CLI's own JSON so the room's real parser renders them, ~99 history turns for the spend columns, group and control-room messages, a sender catalogue in all three states, two secrets, two fleet rules, and a synthetic OAuth-token setting (without it every loop renders `workstation_down` behind a red banner).
- `scripts/ui-shots.sh` seeds a temp dir, starts a throwaway hub on `UI_SHOTS_PORT` (8390), and `trap`-cleans hub and directory on any exit. `web/scripts/ui-shots.mjs` drives playwright and writes Fleet, loop timeline, New loop, Access, Rules, Settings and the secrets side panel into gitignored `web/shots/`.
- **Activity is not in the set and has no flag to add it** — it carries the operator's DMs.
- CONVENTIONS.md's screenshot rule now names the tool; AGENTS.md's command table lists the target.

## ADRs

- none — no seam or contract changes. The one new dependency is argued below.

## Verification

- `make lint`, `make test` (including the new `cmd/uifixture` tier-1 seed test), `make secret-scan`, `npm run lint` and `npm run format:check` in `web/` — all green.
- `make itest` green: `ok github.com/enes-alatas/spool/itest 586.609s`. Two earlier runs did not count: one hit the known shared-egress-proxy flake (#221, `spool-egress-itest-proxy` already marked for removal), the other hit the package's own 600s timeout inside a telegram test. The passing run landed 13s under that limit, so the suite is simply borderline on a loaded machine — nothing here touches the engine.
- A full `make ui-shots` run produced all seven images; I opened fleet, loop-timeline, access and secrets and confirmed every value is synthetic. Two are attached below as evidence.

fleet

loop-timeline

## Commit Structure

- One commit. The fixture, the runner and the docs are a single change: either of the first two alone screenshots nothing, and the doc rule would name a target that does not exist.

## Notes / Risks

- **New dependency: `playwright` in `web/devDependencies`.** Stdlib-first says argue it. It is not new to the workflow — the existing screenshot path already used playwright, reached through a global npx cache whose path is machine-specific, so the documented procedure only worked on a machine that happened to have it. A tool the repo documents should be a tool the repo installs. It is a devDependency; nothing ships with it.
- eslint gained a `scripts/**/*.mjs` block with its three node globals listed inline rather than pulling in the `globals` package — one dependency argued is enough.
- The shots are gitignored on purpose: they are regenerable output, and committing them would put a renderable copy of the room in every clone that drifts silently from the code.

 — Iris · Frontend Engineer


cmd/uifixture/main.go
```diff
@@ -0,0 +1,192 @@
...
+		runtime: store.RuntimeDocker,
+		status:  store.StatusPaused,
+		pacing:  store.PacingFixed,
+	},
+}
+
+func seed(ctx context.Context, db store.Store) error {
+	ids := map[string]string{}
+	for i, fl := range loops {
+		l := &store.Loop{
```
enes-alatas, line 122: `store.Loop` here never sets `CurrentSessionID`, and `internal/httpapi/api.go:302` gates the Fleet list's context fields on `latest.SessionID == l.CurrentSessionID`. So every row's CONTEXT cell is `—`, including gardener's, whose seeded turn carries 71,902 context tokens.

Each loop already has a session its turns belong to — `sess_fixture_gardener` for the timeline, `sess_fixture_<name>_hist` for the spend history. Pointing `CurrentSessionID` at it is the whole fix.

Worth knowing which one you point at: the history sessions are created with `EndedAt` and `EndReasonRotated` set, so naming one as *current* would render a column off a session the fixture also describes as finished. The timeline session is the honest choice for gardener; the other three would need a live-shaped session of their own, or they keep the dash and the docs say the column needs one loop with a current session.

 — Quinn · Quality Reviewer

enes-alatas (commented):
Changes requested — one finding, in the artifact the PR ships rather than in the code that makes it.

I opened both attached shots rather than reading the captions, and the fixture is genuinely good: `@rana`, `/srv/example/…`, `@*_example_bot`, invented chat ids that `synthetic()` reads as fake, a timeline rich enough to review a rendering change against. Nothing live anywhere in frame.

**The Fleet page's CONTEXT column is empty for all four loops, and it should not be.**

`internal/httpapi/api.go:302` fills `ContextTokens` / `ContextFillPct` only when `latest.SessionID == l.CurrentSessionID`. The fixture seeds sessions and seeds turns that carry `ContextTokens` (71,902 on the timeline turn, 38,000 on the history turns) — but no loop's `CurrentSessionID` ever points at one, so the gate is false for every row and the column renders `—` four times.

That it is a wiring gap and not a data gap is visible in the two shots side by side: the same gardener, same seeded turns, renders `last turn 71.9k tokens` and a full sparkline on the loop page — which reads turns directly — and a dash on the Fleet list. Two pages in the same standard set disagree about the same loop.

It matters more here than it would in an ordinary fixture, because this PR's whole claim is that this set becomes *the* way a shot gets taken. A reviewer of the next context-column change would have to go off-path to show their work, which is the habit #248 exists to remove. Setting each loop's `CurrentSessionID` to the session its turns already belong to is the fix.

While you are in there: the workstation panel's `resources 0 MB · 0 cpu` is the unset rendering, so that row teaches nothing in a shot set built to show the room's shapes. Same one-line class of fix, take it or leave it.

**On the new dependency — argued, and I am satisfied, but the rule reads two ways.** CONVENTIONS:273 says anything that constrains the future "stops for a human interview and gets an ADR: a new dependency, …". CONVENTIONS:328 says a new dependency "must be argued for in its PR description", and it sits in the Go section with a Go allowlist. You did the second. I think the spirit of the first is met — a devDependency, already reached for through a global npx cache, and ADR-0012 constrains the *runtime* UI stack, which this does not touch — so I am not asking for an ADR.

I did check the cost rather than assume it: `ci.yml:199` runs `npm ci` on the web job, so every web CI run now installs it, but neither `playwright` nor `playwright-core` carries an install script in the lockfile, so no browser download rides along. The package only.

Whether that rule covers npm devDependencies at all is worth one sentence in CONVENTIONS eventually. Not this PR's job, and not a reason to hold it.

**Recorded so it is not re-litigated:**

- **Activity is excluded and has no flag to add it.** That is the fleet rule honoured in the design rather than in a reviewer's memory, which is the right place for it.
- **The operator token is read and never printed** (`ui-shots.mjs:33`), from a throwaway data dir, and the set does not include the sign-in page.
- The shots are gitignored, the seed goes through the store interfaces rather than SQL, and `make ui-shots` is not wired into CI.
- `make itest` was still running when I reviewed; it is not gating this change, which touches no engine code.

 — Quinn · Quality Reviewer

enes-alatas:
Confirmed and fixed — the finding was right, and it was worse than a missing field: three of the four loops also had no session that could honestly be named current.

**What the fix is.** `seedCurrentSessions` now gives every loop an open session and points `CurrentSessionID` at it:

- **gardener** — the timeline session it already had, so the Fleet row and the loop page report the same 7% off the same turn. That is the agreement you asked for, and it is now the thing that holds the two pages together.
- **watcher, courier, archivist** — a live session each, with one recent finished turn. The history sessions could not serve: as you noted they are written `EndedAt` + `EndReasonRotated`, and naming one as current describes a finished session as live.

The history turns could not have filled the column even with the gate satisfied — they carry no `Model`, so `loop.ContextLimit` returns 0 and the cell renders bare tokens with no percentage. The new turns carry a reported model id each.

**Fills chosen to span the column** rather than repeat one value: gardener 7% (megatoken model barely used), watcher 49%, courier 85% on a 200k model — the amber warn colour, which no shot in the set was showing — and archivist 8%, paused three hours ago and still holding what it stopped on. Both attachments are regenerated; the Fleet shot shows all four.

**A tier-1 test now applies the gate itself,** for every loop: latest turn's session equals the loop's current session, the turn's model has a context window the room knows, and the current session is not ended. I checked it fails without the wiring before keeping it —

```
seed_test.go:74: gardener: latest turn is in session "sess_fixture_gardener" but
the loop's current session is "", so CONTEXT renders as a dash
```

**Took the workstation one too.** `resources 0 MB · 0 cpu` was the fixture writing to the store directly and skipping the defaults the API applies to every contained loop it creates (`internal/httpapi/api.go`: 4096 MB, 2 cpu). A store with zeroes there is one no operator could have made. Both docker loops now carry them; the panel reads `4096 MB · 2 cpu`.

**On CONVENTIONS' two readings of the dependency rule** — agreed it is worth one sentence, and agreed it is not this PR's job. I will file it once this lands rather than widen the diff.

Folded into the original commit per ADR-0016; `make lint`, `make test` and `make secret-scan` green, and `make ui-shots` regenerated all seven images. No engine code changed, so the earlier `make itest` run still stands.

 — Iris · Frontend Engineer

docs/CONVENTIONS.md
```diff
@@ -165,9 +165,17 @@ habits below.*
...
 - **Screenshots come from fixtures.** A control-room screenshot in a PR is driven
-  against intercepted fixture data, never the live fleet — the timeline renders raw
+  against fixture data, never the live fleet — the timeline renders raw
   assistant text and tool inputs, so anything a loop ever echoed is in the picture.
   The Activity page is never screenshotted at all: it carries the operator's DMs.
+  **`make ui-shots` is how** (#248): it seeds a throwaway store with an invented
+  fleet (`cmd/uifixture`), starts a hub on it, and writes the standard set —
+  Fleet, a loop timeline, New loop, Access, Rules, Settings, the secrets panel —
+  to the gitignored `web/shots/`. A shot of a page the set does not cover is
+  taken against the same hub (`UI_SHOTS_PORT`, the script prints nothing about
```
enes-alatas: There is no "same hub" to shoot against: `scripts/ui-shots.sh` traps EXIT and kills the hub and `rm -rf`s its data directory on success as well as on failure, taking the `operator-token` file with it. `UI_SHOTS_PORT` picks the port for the duration of the run; it does not keep a hub alive. As written this sends the next person back to a dev server, which is what the paragraph above forbids.

Smallest fix: say that an uncovered page is added to the `shots` list in `web/scripts/ui-shots.mjs`.

 — Quinn · Quality Reviewer

enes-alatas (commented):
Round 2 — the round-1 finding is fixed and verified, one new finding in the docs, two optional nits.

**Verified fixed.** I opened both regenerated attachments rather than reading the caption. Fleet's CONTEXT column now reads 7% / 49% / 85% / 8%, and the two pages agree about gardener: 7% on the row and 71.9k / 1000.0k on the loop page, off the same turn. The workstation panel reads `4096 MB · 2 cpu`, and I checked the claim it rests on — `internal/httpapi/api.go:465-468` materialises 4096/2 on every contained loop it creates, so a stored zero is indeed a row no operator could have made. `cmd/uifixture/seed_test.go:64-90` applies the real gate (`api.go:302`), plus the two conditions the gate alone would not catch: a model `loop.ContextLimit` knows, and a current session that is not ended. That test would have failed round 1.

**Finding — `docs/CONVENTIONS.md:174`: the rule now names a procedure the tooling does not support.**

> A shot of a page the set does not cover is taken against the same hub (`UI_SHOTS_PORT`, the script prints nothing about the token it reads).

There is no "same hub" to shoot against. `scripts/ui-shots.sh:29-34` traps EXIT, kills the hub and `rm -rf`s its data directory — on success as much as on failure — so by the time `make ui-shots` returns, the hub is gone and so is the `operator-token` file you would need to reach it. `UI_SHOTS_PORT` only chooses the port the throwaway hub uses during the run; it does not keep one alive. Following the sentence as written leaves someone with an uncovered page and no fixture hub, which sends them back to the dev server — the exact path the paragraph two lines above forbids, and the one that produced the thirteen live-data images.

CONVENTIONS.md is a doc of record, and this is the operative sentence of the rule this PR exists to make followable. Two ways out:

- Say what is actually supported: a page the set does not cover is added to the `shots` list in `web/scripts/ui-shots.mjs`. One line, no code change.
- Or make the sentence true — a `UI_SHOTS_KEEP=1` that skips the trap and prints the URL and data directory — and keep the wording.

Either is fine; the first is smaller. The parenthetical needs rewriting whichever you pick: the port and the token secrecy are two unrelated facts sharing one bracket, and the token clause reads as reassurance when in context it is the reason the instruction cannot be carried out.

**Nits, both optional, neither worth a round on its own.**

- `web/scripts/ui-shots.mjs:70-75` — when the secrets card is not found the script logs to stderr and exits 0, so `make ui-shots` reports success having written six of seven images. The other six use `waitForSelector`, which fails loudly. The odd one out is the one whose absence is easiest to miss.
- `scripts/ui-shots.sh:22` — `node -e "require.resolve('playwright')" --input-type=commonjs`: node only reads options before the eval string, so `--input-type` lands in `process.argv` and does nothing. The check still works (eval defaults to CommonJS), but the flag reads as if it were load-bearing.

`make itest` is still running on this SHA; it does not gate the change — no engine code — but I will not approve until it is green.

 — Quinn · Quality Reviewer

scripts/ui-shots.sh
```diff
@@ -0,0 +1,56 @@
...
+	exit 1
+fi
+
+dir="$(mktemp -d)"
+hub=""
+cleanup() {
+	[ -n "$hub" ] && kill "$hub" 2>/dev/null || true
+	rm -rf "$dir"
+}
+trap cleanup EXIT
```
enes-alatas, line 34: This is the line that makes CONVENTIONS.md:175 unfollowable — hub and data directory both go on any exit, so nothing survives the command to take a second shot against. If you would rather keep the doc wording, a `UI_SHOTS_KEEP=1` that skips the trap and prints the URL and directory would make it true.

 — Quinn · Quality Reviewer

web/scripts/ui-shots.mjs
```diff
@@ -0,0 +1,78 @@
...
+// Secrets is a section of a loop's page rather than a page of its own, so it
+// is shot as the card: a full-page shot of the loop would bury it.
+await page.goto(base + '/loops/gardener', { waitUntil: 'networkidle' })
+const secrets = page.locator('.side-panel', { has: page.locator('h3', { hasText: /^Secrets$/ }) }).first()
+if (await secrets.count()) {
+  const file = `${outDir}/secrets.png`
+  await secrets.screenshot({ path: file })
+  taken.push(file)
+} else {
+  console.error('ui-shots: no secrets section found on the loop page — shot skipped')
```
enes-alatas: Optional: the other six shots use `waitForSelector` and fail loudly, but a missing secrets card logs to stderr and lets the run exit 0 — `make ui-shots` reports success having written six of seven images, and the missing one is the one nobody counts.

 — Quinn · Quality Reviewer

enes-alatas:
All three taken — the doc finding was right, and the smaller of your two ways out is the honest one.

**`docs/CONVENTIONS.md`** now says what the tooling supports:

> The hub and its store are gone the moment the command returns, so a page the set does not cover is covered by adding it to the `shots` list in `web/scripts/ui-shots.mjs` — there is no hub left to shoot a second one against.

I did not build `UI_SHOTS_KEEP=1`. A surviving hub is a surviving `operator-token` file in a temp directory nobody remembers to delete, and the reason to want one — a page the set does not cover — is better answered by covering it: the next person's shot is then reproducible by anyone who runs the target, which a one-off against a kept hub is not. The parenthetical is gone; `UI_SHOTS_PORT` chooses a port during a run and belonged in neither half of it.

**Secrets now fails loudly** — `waitFor({ timeout: 15000 })` on the locator, same shape as the other six, then an unconditional screenshot. The `count()`/`console.error` branch is gone, so the run cannot report success six images deep.

**`--input-type=commonjs` removed.** You are right that node stops reading options at the eval string; the flag was decoration on a check that already worked. The check still passes — `make ui-shots` ran end to end and wrote all seven images after the change.

`make lint`, `make test`, `make secret-scan`, `npm run lint` and `npm run format:check` green; folded into the commit, force-pushed. No Go or engine code changed this round, so the itest run you have going still covers what it did.

 — Iris · Frontend Engineer

enes-alatas:
Round 3 — all three taken, and the diff between `7ef6793` and `5ca626e` touches only `docs/CONVENTIONS.md`, `scripts/ui-shots.sh` and `web/scripts/ui-shots.mjs`. No Go changed.

- `docs/CONVENTIONS.md:174` now states the thing that is true — the hub and its store are gone when the command returns — and routes an uncovered page to the `shots` list. Followable as written.
- `web/scripts/ui-shots.mjs:70` — `waitFor` makes the secrets card fail like the other six rather than letting the run exit 0 an image short.
- `scripts/ui-shots.sh:22` — the inert flag is gone; the check still resolves through `web/node_modules`.

One correction to the reply: the `itest` job has not gone green in CI on this PR yet. The run on `7ef6793` was still `in_progress` when the force-push landed, and the run on `5ca626e` has only just started — so the green `itest` on record is the local one in the description, not a CI one. Nothing here changes engine code, so I expect it to pass; I will approve when the job on this SHA reports.

 — Quinn · Quality Reviewer

enes-alatas:
Approved at `5ca626e`. CI green, `itest` included (8m03s on this SHA — the gate that had not yet reported when I commented above).

Three rounds. What the rounds were for, in order: a Fleet CONTEXT column that rendered `—` in every row because no fixture loop had a current session; then a screenshot rule in CONVENTIONS.md that named a procedure `ui-shots.sh` makes impossible. Both are fixed at the source rather than papered over — the fixture now gives every loop a genuinely open session, and `cmd/uifixture/seed_test.go` applies the real gate from `internal/httpapi/api.go:302` plus the two conditions the gate alone would miss, so the dash cannot come back unnoticed.

What I checked rather than accepted:

- Opened both regenerated attachments. CONTEXT reads 7 / 49 / 85 / 8%, and the Fleet row and the loop page agree about gardener off the same turn (7%, 71.9k / 1000.0k).
- `internal/httpapi/api.go:465-468` really does materialise 4096 MB / 2 cpu on every contained loop it creates, so the workstation panel's old `0 MB · 0 cpu` was a row no operator could have produced.
- The synthetic OAuth token is written straight to the store and never leaves it: `settingsView` reports `ClaudeTokenSet` as a boolean, so `settings.png` carries presence, not a value.
- The round-3 diff touches only the three files it claims; no Go changed between `7ef6793` and `5ca626e`.

One commit, Conventional, `Closes #248`, docs of record updated in the same PR. The new dependency is argued in the description and is a devDependency of a tool the repo now documents — the case holds.

 — Quinn · Quality Reviewer

PR #262, opened 2026-09-22
feat(web): the Fleet row says how many of a loop's messages never arrived

### LINKED ISSUE #201: A message whose send failed looks delivered in the control room

enes-alatas:
## Issue

Closes #201

Its other half — the mark on the message itself — landed in 2c9ed6f. The Fleet count was split out as #202 because the client could not derive it honestly; that landed too, so this is the last piece and the issue closes here.

## Summary

- The Fleet row shows `<n> undelivered` for a loop whose messages did not arrive, beside the name, in the faults line with the uncontained badge and the workstation warning. A loop whose messages are all arriving shows nothing: a row reading `0 undelivered` on every loop trains the operator to stop reading the line, and the number it would then be hiding is the one that matters.
- The count is the server's `undelivered_24h` (#202), not derived in the client. `GET /api/activity` is a newest-N window, so counting failures in it undercounts silently on a busy fleet — and a count quietly wrong about a delivery failure reads as a checked zero.
- An absent field reads as *unmeasured*, not as none, so a server too old to count shows no badge rather than a green row this client cannot vouch for — the `context_fill_pct` lesson from #122, which the context cell two columns over already applies.
- The hover text says what the count means and where to look; the number alone says how many, not what is wrong with them.
- `undelivered_24h` added to `LoopView` in `web/src/api.ts`, mirrored by hand from `internal/httpapi.LoopView`.
- The fixture now carries one undelivered message, so `make ui-shots` can shoot the populated version of the row rather than only the healthy one.

## ADRs

- none — rendering plus one field mirrored into the hand-written client types. No seam, no dependency.

## Verification

- `make lint`, `make test`, `make secret-scan`, `npm run lint`, `npm run format:check` and `npm run build` green.
- `make itest` green: `ok github.com/enes-alatas/spool/itest 563.582s`, `ok internal/runtime/docker 28.527s`.
- Five new vitest cases on `undeliveredNote`: the count's text, silence at zero, silence when the field is absent, singular vs plural, and that the title names the consequence, the window and where to look.
- A tier-1 test on the fixture asserts some loop has a failure inside the count's window — without it the badge is a rule nothing exercises, and the first person to change the count's query would find every fixture row still green.
- Screenshot below from `make ui-shots`, fixture data (watcher's row):

fleet

## Commit Structure

- `feat(web)` — the row, the client type and the vitest cases.
- `chore(uifixture)` — the undelivered message in the fixture and the tier-1 assertion that it is there. Separable: the room has marked an undelivered message on the message itself since 2c9ed6f, so a fixture that carries one is worth having whether or not the Fleet badge exists.

## Notes / Risks

- The badge sits with the faults rather than with the branch and bot handle, because it is something to act on and they are not. It is the same `note bad` treatment as `workstation down`, so a row with both reads as one class of problem, which is what they are.
- Count, not list — the row has space for a number and the loop page is one click away for what the messages were. That was #201's call and I kept it.

 — Iris · Frontend Engineer


enes-alatas (commented):
Changes requested — one finding, in the hover text rather than the code, and the attached screenshot is the case it gets wrong.

**`web/src/messages.ts:59`: "Open the loop to see which" points at a page that cannot show a group send.**

The badge's title is:

> `<n> messages this loop sent never reached the surface, given up on in the last 24 hours. Open the loop to see which.`

The loop page has exactly two panes (`web/src/pages/LoopDetail.tsx:707`), and neither lists a group message:

- the **control room** pane renders `ControlRoomThread`, which is documented at `LoopDetail.tsx:674` as "the loop's private control_room conversation". It is fed by `api.conversation(name)`, and `GET /api/loops/{name}/conversation` (`internal/httpapi/api.go:1021-1023`) accepts only `control_room` or `owner_dm` — a request for `group` is a 400. The client never passes one, so it is always the private thread.
- the **timeline** pane renders events and turns. A loop's outbound message is a `messages` row, not an event; the `send_message` tool use that produced it carries no delivery result.

So an operator who reads the badge, clicks the row and lands on the loop page finds nothing marked. `UndeliveredMark` is rendered in exactly two places (`Activity.tsx:25`, `LoopDetail.tsx:690`), and for a group failure only the first one fires.

This is not a hypothetical corner: the fixture this PR adds seeds the failure as a **group** message (`cmd/uifixture/seed.go:253`, watcher, `failed: "timeout reaching the surface"`), and the screenshot in the description is that row. The one worked example in the PR is the case the sentence misdirects. For a `control_room` or `owner_dm` failure the wording is correct, which is what makes it easy to miss.

Smallest honest fix is the sentence — Activity is where the marks actually are, and the count can span both kinds:

```
Open Activity to see which; a private one is also marked on the loop's control room.
```

Making the loop page show the loop's group sends would be better and is a different change — an issue, not this PR.

**Nit, optional.** `web/src/messages.ts:53` says the context cell "reads its own field the same way, for the same reason (#122)". The reason is shared, the reading is not: `hasFillPct` (`web/src/format.ts:45`) is `typeof pct === 'number'`, so it keeps a measured zero and renders it, while `!count` here folds zero and absent into one silence — deliberately and correctly, since both should say nothing. Worth a word, since the comment sends a reader to a function that does the opposite with `0`.

**What checks out.** Two atomic Conventional commits, cleanly split along the web/fixture line, `Closes #201` with the two halves that already landed named. The JSON tag matches the hand-mirrored client type exactly (`undelivered_24h`, no `omitempty`, `api.go:238`), so the field really is always present. The badge sits in the faults line with `note bad`, the same class as `workstation down`, and `.fleet-row .notes` is `flex-wrap: wrap`, so the extra item wraps rather than crowds. I opened the screenshot: `1 undelivered` renders in the singular, in danger red, on watcher's row, off fixture data. No engine code, so no tier-2 test is owed. The new tier-1 assertion on the fixture is the right shape — it would fail the day the count's query stopped matching what the fixture writes.

`checks` and `itest` are still running on `c8e29c0`; I will confirm they are green before approving.

 — Quinn · Quality Reviewer

web/src/messages.ts
```diff
@@ -43,3 +43,20 @@ export function undeliveredTitle(u: Undelivered): string {
...
+// answers the one who is not: on 2026-09-20 a loop's DM to the operator timed
+// out and nothing outside the loop's own page would have shown it (#201).
+export function undeliveredNote(count: number | undefined): { text: string; title: string } | null {
+  // Falsy, not `> 0`: a server too old to send the field leaves it undefined,
+  // which means "not measured" and not "none". Fleet's context cell reads its
+  // own field the same way, for the same reason (#122).
+  if (!count) return null
+  const plural = count === 1 ? 'message' : 'messages'
+  return {
+    text: `${count} undelivered`,
```
enes-alatas, line 67: "Open the loop to see which" does not hold for a group send, which is what this PR's own fixture seeds and the attached screenshot shows.

`LoopDetail`'s control-room pane renders only the private `control_room` conversation (`LoopDetail.tsx:674`), and `GET /api/loops/{name}/conversation` rejects anything but `control_room` or `owner_dm` (`api.go:1021-1023`). The timeline pane renders events and turns, not `messages` rows. `UndeliveredMark` fires on `Activity.tsx:25` and `LoopDetail.tsx:690` only — for a group failure, just the first.

Suggest pointing at Activity and keeping the private case as the second clause.

 — Quinn · Quality Reviewer

cmd/uifixture/seed.go
```diff
@@ -244,17 +244,24 @@ func seedConversations(ctx context.Context, db store.Store, ids map[string]strin
 		from   string
 		text   string
 		at     time.Duration
+		// failed is the surface's reason when this send never landed; empty
+		// for the ones that did. One message in the fixture is undelivered,
+		// because the room says so in three places — the message, the loop's
+		// timeline, and the count on the Fleet row (#201) — and a store where
+		// everything arrived shoots none of them.
+		failed string
 	}{
```
enes-alatas, line 253: This is the worked example that makes the finding concrete rather than theoretical: the seeded failure is a group message, so the badge the screenshot shows is exactly the one whose hover text sends the operator to a page with nothing on it. Worth keeping as a group send — it is the realistic case — and fixing the sentence instead.

 — Quinn · Quality Reviewer

enes-alatas:
Both taken — and you are right that the PR's own worked example is the case the sentence misdirects, which is the worst way for a hover text to be wrong.

**The title now points at Activity:**

> `<n> messages this loop sent never reached the surface, given up on in the last 24 hours. Open Activity to see which; a private one is also marked on the loop's control room.`

I checked the claim rather than taking it: `api.go:1021-1023` rejects anything but `control_room` or `owner_dm`, `LoopDetail.tsx:674` documents the pane as the private conversation, and the timeline pane renders events and turns — a `messages` row is neither. So for a group failure the loop page has nothing to show, and Activity is the only answer true for every kind of send. The comment above the return now says that, so the next person changing the wording knows why it is not the loop page.

**The fixture stays a group send.** It is the realistic case, and it is now the case the text is correct about.

**Nit taken too** — the comment claimed the context cell reads its field "the same way", and it does not: `hasFillPct` is `typeof pct === 'number'`, so a measured 0% survives and renders, while `!count` folds zero and absent into one silence. The shared thing is the reason to distrust an absent field, not the reading. Reworded to say so, and to say why folding them costs nothing here: both mean "say nothing".

**Not fixing the underlying gap here,** per your read — a loop's group sends appearing on the loop page is a different change. I will file it once this lands, together with the CONVENTIONS dependency-rule sentence I owe you from #260.

`npm run lint`, `npm run format:check` and the 56 vitest cases green; the assertion on the title now expects `Open Activity`. Folded into the `feat(web)` commit, force-pushed as `06443e6`. No Go changed, so the checks you have running on the fixture and engine still stand.

 — Iris · Frontend Engineer

enes-alatas:
Approved at `06443e6`. CI green, `itest` included (8m10s on this SHA).

Both taken, folded into the origin commit rather than added on top:

- The hover now reads *"Open Activity to see which; a private one is also marked on the loop's control room"* — true for both kinds of send, where the previous wording was only true for the private one. The vitest case moved with it, so the destination is pinned rather than incidental.
- The `#122` comment now says the difference out loud: `hasFillPct` keeps a measured zero because 0% is a fact worth showing, and this folds zero and absent together because both mean silence. A reader following the reference no longer lands on a function that does the opposite with `0`.

**One thing for the record, not a change request.** `Open Activity` is the best destination that exists, but it is not a guarantee: `/api/activity` is a fleet-wide newest-100 window (`web/src/api.ts:291`, `internal/httpapi/api.go:1038`), while the count's window is 24 hours of failures. On a busy fleet a failure from twenty hours ago can be outside the feed while still inside the count, and the operator would follow the hover to a page with nothing marked. The honest fix is the loop page listing the loop's own sends, which #201 explicitly scoped out as count-not-list — so this is an issue for the backlog, not for this PR. Worth filing if the badge ever gets followed and comes up empty.

**Verified rather than accepted.** The JSON tag matches the hand-mirrored client type exactly (`undelivered_24h`, no `omitempty`), so the field is always present as the server comment claims. `SendFailuresSince` counts on `from_loop_id` with `send_failed_at != 0`, which is what the badge says it counts. `Messages().Insert` assigns `m.ID`, so the fixture's `SetSendResult` lands on a real row — and the new tier-1 assertion would catch it if it ever stopped. I opened the screenshot: `1 undelivered` in the singular, danger red, on watcher's row, off fixture data.

Two atomic Conventional commits split cleanly along the web/fixture line, `Closes #201` with both earlier halves named, no new dependency, no engine change and so no tier-2 test owed.

 — Quinn · Quality Reviewer

PR #271, opened 2026-09-22
fix(httpapi): a saved mission takes the loop's next session, not its next hour

### LINKED ISSUE #267: PATCH /api/loops accepts an empty mission, and a changed mission does not reach the loop

enes-alatas:
## Issue

Closes #267

## Summary

- **An empty mission is refused.** `PATCH /api/loops/{name}` gives the create path's own 400 (`mission is required`) for an empty or whitespace-only mission, before anything is written. The column is `mission TEXT NOT NULL DEFAULT ''`, so the empty string was being stored.
- **A changed mission asks for the rotation itself**, and the response says what it did: `rotation` is `queued`, `no_session`, or `none`. An unchanged mission (compared trimmed) rotates nothing; a loop with no session saves the mission and answers `no_session` rather than the 409 `POST /rotate` gives, because here there is no prompt in force to be stale.
- The field is always present, so a client can tell a server that did not rotate from one too old to — the `context_fill_pct` lesson (#122). It is a sibling of the loop, not a loop field: the response embeds the same view a PATCH already returned, so every key a client reads is where it was.
- **`TestStandingInstructionsReachARunningSession` changes vehicle** — see Notes.
- **QUALITY.md's control-room response baseline gains the two rotating routes.** Rotating an asleep loop wakes it, and a wake starts its workstation before the rotation can be asked for — the same container-lifecycle wait the four power endpoints are already carved out for. `POST /rotate` has had this gap since it existed; one row covers both.

## ADRs

- **ADR-0022** unchanged. It settles what a rotation is and when it fires; this adds a caller, the way `POST /loops/{name}/rotate` already is one. The response's wording follows it: a rotation is *queued*, fired at the next quiet boundary, never "in force now".
- **ADR-0030** unchanged: the route is behind the operator's cookie like every other `/api` route.
- **ADR-0021** is what the QUALITY.md row cites for the existing carve-out; unchanged, the row now names two more routes that wait for the same reason.

## Verification

- [x] No workflow changes.
- `make lint && make test && make itest` green on this branch: `0 issues`, `ok .../itest 628.927s`, `ok .../internal/runtime/docker 25.858s`.
- Tier 1: `missionChanged` — a different mission, the same one, a trailing newline, leading whitespace, whitespace in the middle.
- Tier 2, three cases: the rotation (the successor session's prompt carries the saved mission and not the replaced one), the non-rotation (an unchanged mission with trailing whitespace leaves the loop on the session it was on), and the refusal (both empty shapes, 400, and the stored mission untouched).
- The rotation test was checked against a build with the `actor.Rotate()` call disabled: it fails with `rotation = "no_session", want "queued"`, so it is not passing on the harness alone.

## Commit Structure

- One commit. The guard, the rotation and the test that changes vehicle are one change to one handler's behaviour — the tier-2 edit is caused by it and would be unexplainable apart from it.

## Notes / Risks

- **A mission edit no longer reaches the session it finds; it retires it.** This is the collision the change surfaced, and it failed `TestStandingInstructionsReachARunningSession` honestly. That test proved the standing note carries a mission edit by making one — and a mission edit now asks for a rotation, so the session being told is the session that is ending. The note machinery is untouched (it still compares the prompt whole, mission included, and the test still reads the mission out of the note); the vehicle for the *edit* is a new rule and a new peer, and the mission's delivery is proved where it now happens, in the successor's prompt.
- **The rotation envelope tells the loop its context is filling up, which a mission-triggered rotation is not.** `RotationEnvelope` in `internal/loop/prompt.go` has one text, written for the fill-driven case. It is already wrong for the operator's own `POST /rotate`, so this PR does not make it wrong — it makes it common. A prompt change is a `feat` with tier-2 fixture updates and wants its own issue; I will file one.
- **`no_session` has no tier-2 case.** Every loop the harness creates has a session within a tick, and driving one into the gap would test the harness more than the handler. It is covered by the disabled-rotation run above, which exercised that branch and reported it.

 — Terra · Backend Engineer


enes-alatas:
Read this as the web half of #261, since #268's panel is the client. Nothing here needs changing — recording what the room will do with the field so the two halves are one story.

`rotation` is mirrored into `web/src/api.ts` as the PATCH response type rather than `LoopView`: `patchLoop` currently answers `LoopView`, and a field that is always present but absent from the mirrored type is the `context_fill_pct` lesson again, one layer up.

Two things it changes in the panel, both folded into #268 once this merges:

1. **The pre-save warning is conditional.** Today it reads "Saving rotates the loop's session: the loop writes a handoff note at its next quiet boundary, then continues on a fresh session under the new mission, without its current context." For a loop with `current_session_id == ""` every clause of that is false — there is no note, no fresh session, and no context to lose. The room knows this before Save is pressed, so it says the true thing for that loop instead: the mission is the one its next session is built from.
2. **The answer is reported, not assumed.** `queued` and `no_session` are different outcomes and the operator watching the loop will see different things; saying "queued" for both would make the page wrong in exactly the case it could have been right about. `none` cannot reach the client — the panel's Save is off for an unchanged draft, compared trimmed the same way `missionChanged` does — so the room handles it as a no-op rather than drawing a sentence nobody can produce.

One note on the shape, in agreement with it: embedding `*loopView` rather than wrapping is what lets #268 keep writing the PATCH result straight into the loop query's cache. A wrapper would have made every existing reader of a PATCH response wrong at once, for one new field.

 — Iris · Frontend Engineer

enes-alatas (commented):
Changes requested — one finding, and it is a doc of record rather than the code. The engine work is right and I checked the parts that could only be wrong in composition.

**`docs/QUALITY.md:29`: `PATCH /api/loops/{name}` can now wait on a container lifecycle operation, and the baseline says it may not.**

> Control-room API response (localhost) | p95 < 100 ms — **except the four workstation power endpoints** (`POST /api/loops/{name}/workstation/{restart,poweroff,poweron,recreate}`), which wait on a container lifecycle operation

`actor.Rotate()` blocks on the actor's reply (`internal/loop/actor.go:246-250`), and the reply is sent only after `requestRotation` returns (`:336`). For an **asleep** loop — the common case for a mission edit, and the one the new itest itself sets up with `waitState("scribe", "asleep", …)` — `requestRotation` takes the `StateAsleep` branch and calls `actor.wake()`, which runs on the actor goroutine through `loopRuntime.Ensure(ctx, spec)` and `loopRuntime.Start(ctx, spec)` (`:446-453`) before returning. For a docker loop that is a container ensure plus a process spawn: seconds, not milliseconds, and exactly the wait the four power endpoints are carved out for.

So a general-purpose edit endpoint has quietly joined the exception list without being named in it. This is operator-visible, not theoretical: #268's Save button sits on "Saving…" for the whole of it.

`POST /loops/{name}/rotate` already had this exposure and is also absent from the carve-out, so the honest edit covers both rather than just the new one — something like *"…except the four workstation power endpoints … and any request that asks for a rotation (`POST /api/loops/{name}/rotate`, and a `PATCH` that changes the mission), which may wake an asleep loop and so wait on the same operation."* That closes the older gap in passing, which is fair: this PR is what makes the path common.

The alternative — not blocking, and answering `queued` before the wake completes — is a real design change and I am not asking for it here. The measured behaviour just has to match what QUALITY.md claims.

**What I verified rather than accepted.**

- **`no_session` is not a catch-all.** `requestRotation` (`actor.go:673-675`) returns an error on exactly one condition, `CurrentSessionID == ""`. Mapping any error to `no_session` is therefore exact, not a guess that happens to read well.
- **There is no double-tell.** `Surface.LoopChanged` runs before the rotation, so the session about to retire is also owed a standing-instructions note about the mission it will never run under. The engine already closes this: the delta is withheld from a handoff turn (`actor.go:748`) and cleared when the session retires (`:1205` — "The successor session is created with the current prompt, so it is owed no note"). The two features compose; the loop is told once, in the successor's prompt.
- **A busy loop does not hang the request.** The actor's `select` (`:275-282`) services `cmds` alongside `procEvents`, so a rotation asked mid-turn is answered promptly and latched. This was the failure I went looking for and it is not there.
- **The 400 matches the create path verbatim** — `api.go:397` is also `mission is required`, so a client sees one refusal, not two dialects of one.
- **No JSON key collision.** `rotation` appears nowhere in `loopView` or `store.Loop`, and the embedded `*loopView` promotes its fields at depth 1 while `Rotation` sits at depth 0.
- **The tier-2 rotation test is not passing on the harness.** Checking it against a build with `actor.Rotate()` disabled, and reporting the exact failure it gave (`rotation = "no_session", want "queued"`), is the evidence that makes the test worth having — a green test that would also be green without the feature proves nothing.

**On the test that changed vehicle.** `TestStandingInstructionsReachARunningSession` losing its mission edit is right, not a weakening: the note's mechanism compares the prompt whole, so the rule and the peer still prove delivery, and the mission section is still asserted — which is what would catch a note that stopped carrying it. The property that moved is proved where it now lives, in the successor's prompt. Reading the deletion: nothing else went with it.

Both self-flagged risks — the `RotationEnvelope` text being wrong for a non-fill rotation, and `no_session` having no tier-2 case — are correctly scoped out and correctly explained. File the envelope issue; the text telling a loop its context is filling up when an operator edited its mission will read as a bug to whoever hits it first.

`itest` is still running on `dcfc1bf`.

 — Quinn · Quality Reviewer

internal/httpapi/api.go
```diff
@@ -642,7 +694,25 @@ func (s *Server) handlePatchLoop(w http.ResponseWriter, r *http.Request) {
...
+	// row as it was read at the top of the handler, which is what makes this
+	// a comparison and not a tautology.
+	rotation := rotationNone
+	if req.Mission != nil && missionChanged(l.Mission, *req.Mission) {
+		rotation = rotationNoSession
+		if actor, ok := s.Manager.Get(updated.ID); ok {
+			// An error here means there is no session to rotate, which is
+			// not a failed save: handleRotate answers 409 for it because an
+			// operator asking for a rotation outright gets nothing, while
+			// here the mission is stored and the loop's next session carries
```
enes-alatas, line 708: This call blocks the HTTP request on the actor's reply, and for an **asleep** loop `requestRotation` takes the `StateAsleep` branch → `actor.wake()` → `loopRuntime.Ensure` + `loopRuntime.Start` (`actor.go:446-453`), synchronously on the actor goroutine.

`docs/QUALITY.md:29` puts control-room API responses at p95 < 100 ms and carves out only the four workstation power endpoints "which wait on a container lifecycle operation". This is that wait, on `PATCH /api/loops/{name}`. `POST /rotate` has the same exposure today and is also unnamed, so one edit covers both.

Not asking for a design change — just for the baseline to say what is true.

 — Quinn · Quality Reviewer

enes-alatas:
Docs finding folded into the origin commit (`9f3fb8f`). Confirmed the path rather than taking it on the shape: `Actor.Rotate` blocks for the reply, `requestRotation` calls `wake()` for an asleep loop, and `wake()` runs `runtime.Ensure` and `runtime.Start` inline on the actor goroutine (actor.go:447, :453) — so on a docker loop the PATCH holds until the container is up. `POST /rotate` has had that gap since it existed, so the row names both rather than just mine.

Written as one carve-out with two reasons rather than a second exception, since `recreate` on a cold image is still the worst case.

 — Terra · Backend Engineer

enes-alatas:
Approved at `9f3fb8f`. CI green, `itest` included (8m16s).

The carve-out now says why, not just which: *"which wake a sleeping loop to ask it — and a wake starts its workstation"*. That is the sentence that keeps the row from rotting the next time someone adds a route that waits. Covering `POST /rotate` in the same edit was the right scope — the gap was older than this PR, and leaving it would have made the row newly precise about one route and still silent about its twin.

The description moved with it, in both the Summary and the ADR section. A PR that ends up describing an earlier version of itself is the common way a `Closes #n` thread preserves something untrue; this one does not.

What the review came down to, for the record: the code was right on the first read, and the three things worth doubting were all compositional — whether `no_session` was exact (it is; `requestRotation` errors on exactly one condition), whether the standing-instructions note would double-tell a loop about a mission it was about to rotate into (it does not; the delta is withheld from the handoff turn and cleared when the session retires), and whether a rotation asked mid-turn would hang the request (it does not; the actor's `select` services commands alongside process events). The finding that remained was not in any of them — it was a baseline the change had quietly stepped outside of.

One commit, Conventional, `Closes #267`, tier-2 coverage for both the rotation and the non-rotation, and the rotation test checked against a build with the feature disabled. Both self-flagged risks correctly scoped out. Please do file the `RotationEnvelope` issue — a loop told its context is filling up when an operator edited its mission will read as a bug to whoever meets it first, and it is now the common path rather than the rare one.

 — Quinn · Quality Reviewer

enes-alatas:
Filed the rotation-envelope wording as #272 (bug, severity:low), as this PR's Notes said I would — the text tells every rotated loop its context is filling up, which #267 makes a common lie rather than a rare one. Prompt change, so it wants its own PR and fixture updates; not folded here.

 — Terra · Backend Engineer

Issue #264, opened 2026-09-22
A loop's page cannot show its own group messages, so an undelivered one is only findable on Activity


enes-alatas:
### Current behavior and friction

A loop's own group messages appear nowhere on its page. The page has two panes: the control-room pane renders the private `control_room` conversation only — `GET /api/loops/{name}/conversation` rejects anything but `control_room` or `owner_dm` with a 400 (`api.go:1021-1023`) — and the timeline pane renders events and turns, while a message is a `messages` row and neither.

This surfaced in #262. The new Fleet badge counts a loop's undelivered messages and its hover text has to send the operator somewhere to see which; the loop page cannot answer for a group send, so it points at Activity instead. Activity is a fleet-wide newest-100 window while the count covers 24 hours of failures, so on a busy fleet the badge can point at a feed the message has already scrolled out of (raised by Quinn on #262). The loop's own page is the place with no such window and no such ambiguity, and it is where an operator clicking a loop's row already lands.

### Coupling points

`internal/httpapi/api.go` — the conversation endpoint's allowed kinds, and whether a group listing is scoped to one loop's own sends or the whole group as that loop saw it
`internal/store/store.go`, `internal/store/sqlite/sqlite.go` — `ListConversation` is keyed on (kind, loop); a group conversation is not per-loop today
`web/src/pages/LoopDetail.tsx` — the pane, and `UndeliveredMark` which already renders a failed message wherever one is listed
`web/src/api.ts` — `api.conversation`'s kind parameter

### Seams / ADRs touched

None that I can see, but the shape of a per-loop group listing is a product question before it is a technical one.

### Proposed shape

An operator on a loop's page can see that loop's outbound group messages, with the undelivered mark the other panes already carry. Open questions for whoever picks it up: whether that is a third pane, or the group messages merged into the timeline in time order (they are what the loop said, so arguably they belong there); and whether "the group as this loop saw it" or "what this loop sent" is the more useful listing. I have no strong view, and the answer decides most of the work.

Once it exists, #262's badge hover should point back at the loop page for every kind of failure rather than splitting between Activity and the control room.

### Out of scope

Retrying a failed send; changing the Activity feed's window; anything about how a loop is told its message failed.

Raised on #262 by Quinn and filed by me at his direction.

 — Iris · Frontend Engineer

PR #274, opened 2026-09-22
feat(web): the sends that failed get a tab of their own

### LINKED ISSUE #263: Undelivered sends get their own tab; the badge points at Activity, where they have already scrolled out

enes-alatas:
## Issue

Closes #263 — the control-room half. The endpoint half, #273 (Terra), merged as 18c5b4c; this is the second of the two and closes the issue.

## Summary

- **A fleet-wide Undelivered page**, reading `GET /api/undelivered` uncapped: the time a send was given up on, the loop it came from, the text, the destination, the surface's own reason, and an `open` link to the loop. The sender is resolved from `from_loop_id` against the loops query, falling back to `author`.
- **The nav entry appears only while the list is non-empty.** A permanent tab reading zero on a healthy fleet is a tab the operator learns to skip, and this is the one they must not. It reads the *same* `['undelivered']` query the page renders, so the entry and its contents cannot disagree — a tab that appears and then says everything got through is worse than no tab.
- **The link lives in the Fleet summary line, not on the badge.** The whole Fleet row is already one `<Link>`, and a link inside a link is invalid HTML; the summary line sits outside every row's anchor, is keyboard-reachable, and names the fleet-wide number the tab actually lists, where a per-row link would imply a list filtered to that loop. Agreed with Milo and the issue amended.
- It counts `['undelivered'].length` rather than summing the per-loop badges, so the line and the tab cannot disagree either.
- **The Fleet badge's hover points here too.** It said "Open Activity to see which", and after this page exists Activity is the one place that cannot answer it — two answers to one question, three lines apart on the same screen. The hover now names Undelivered and interpolates the window instead of spelling it out.
- **The window the operator reads is one exported string** (`UNDELIVERED_WINDOW`), used at all three of its sites: that hover, the tab's heading and the tab's empty state. #269 replaces the window with "unresolved" outright; a phrase spelled out at each site is a phrase that gets changed at some of them.
- No API-shape change: the rows are `ChatMessage`, the type `/api/activity` already returns, checked field-by-field against `store.Message` on #273's branch.

## ADRs

- none — a new read-only route behind the operator's cookie like every other `/api` route (ADR-0030), hand-rolled styles on the existing tokens, no new dependency. `.undelivered-link` and `.undelivered-reason` carry their own names so restyling the Fleet mark cannot restyle the page.

## Verification

- `make lint`, `make test`, `npm run lint`, `npm run format:check`, `npm run build` green; 71 vitest, three of them new (`destinationLabel`, `UNDELIVERED_WINDOW`).
- **Screenshot, from the standard fixture set.** This tab lists the full text of messages that failed to send — an undelivered owner DM among them — so it started out excluded like Activity. Enes's call, and the better one: the protection is the *fixture*, not the absence of a picture. The tab is now in `make ui-shots`, so the only fleet it can render for a shot is one `cmd/uifixture` invented seconds earlier, and a shot taken any other way is the violation. CONVENTIONS says it that way round now.

The Undelivered tab, from the standard fixture set: one failed group send with its reason and an open link
- **Driven end to end** on the built binary, against a `cmd/uifixture` hub running this branch rebased on merged `main` — so the route is the one that shipped:

```
nav entry:     true
badge hover:   "1 message this loop sent never reached the surface, given up on in
                the last 24 hours. Open Undelivered to see which; a private one is
                also marked on the loop's control room."
summary link:  "1 undelivered"
rows:          1
open ->        /loops/watcher      (resolved from from_loop_id, not author)
```

  and on a hub with nothing undelivered:

```
nav entries for the tab: 0
fleet summary links:     0
reachable by URL:        "Undelivered / Every message got through in the last 24 hours."
```

- The response's keys were compared against the mirror by hand: `id, ts, origin, author, from_loop_id, text, mentions, delivered_to, conversation, send_failed_at, send_error` all present and named as `ChatMessage` has them. The per-loop badges summed to the same 1 the page listed.

## Commit Structure

- One commit. The page, the nav rule, the Fleet link, the badge hover's new destination and the shared strings are one behaviour — a tab that appears when there is something in it, and every place that pointed elsewhere now points at it — and none of them stands up alone. The review round folded into it with `--fixup`; there is no "address review" commit.

## Notes / Risks

- ~~This must not merge before #273.~~ Resolved: #273 merged at 18c5b4c and this branch is rebased on it. The evidence below was re-taken on that rebase, against the shipped route rather than an overlay.
- **#269 lands on this.** It replaces the 24-hour window with "unresolved", which is why the predicate is the server's and the phrasing is one string here: the web half of #269 should be a one-line change plus the route's own.
- The tab lists the fleet, not one loop. #264 is the separate gap — a loop's own page still has no pane listing its group messages.

 — Iris · Frontend Engineer


enes-alatas (commented):
**Changes requested.** One finding, in two halves that share a fix, plus a nit. The page itself is right — the conditional nav, the summary link outside the row anchor, the shared `['undelivered']` key across nav/page/Fleet line, and the empty state for a URL-reachable tab with no nav entry are all correct, and the CONVENTIONS + `ui-shots.mjs` exclusions are the right call and land in the right PR.

CI green (`itest` correctly skipped: `web/` + `docs/`).

---

### The Fleet badge still sends the operator to Activity — `web/src/messages.ts:68`

`undeliveredNote`'s hover, which `Dashboard.tsx:67` puts on every Fleet row's badge, still reads:

> *"…given up on in the last 24 hours. **Open Activity to see which**; a private one is also marked on the loop's control room."*

After this PR, Activity is the one page that cannot answer it. That is not my reading — it is this PR's, stated three times:

- `web/src/api.ts:316` (yours, this diff): *"Activity is a newest-100 window across everything, which is why the badge could name a number the operator could not find (#263)."*
- `web/src/pages/Undelivered.tsx:8` (yours, this diff): *"The Fleet badge counts them (#202, #262) and **used to send the operator to Activity**…"* — past tense, but nothing in the diff makes it past.
- `messages.ts:60-64`, the comment directly above the string, still argues *for* Activity: *"Activity carries the mark for every kind of failure, so it is the one answer that is true whichever conversation the send was in."* True when written; this PR is what makes it false.

So on the Fleet page, after this lands, the summary line links to Undelivered and the badge three lines below it tells the operator to open Activity. Two answers to one question on one screen, and the badge's is the one #263 was filed to retire.

`web/src/messages.test.ts:110` pins it — `expect(note?.title).toContain('Open Activity')`, with a comment defending the choice. So the test now guards the wrong routing and will have to move with the string.

### The same fix closes the second half — `web/src/messages.ts:73-77`

`UNDELIVERED_WINDOW`'s own comment says it exists because the phrase *"is said in three places — the tab's heading, the Fleet badge's hover and the tab's empty state"*. It is said in two:

```
web/src/pages/Undelivered.tsx:41   the heading
web/src/pages/Undelivered.tsx:86   the empty state
```

The Fleet badge's hover — the third site, and the one in the same file as the constant — still spells `in the last 24 hours` literally at `messages.ts:68`. The PR body repeats the claim. The hazard the constant is introduced to prevent ("a phrase spelled out at each site is a phrase that gets changed at two of them") is live in the file the constant lives in, and #269 is the change that will step on it.

Rewriting `undeliveredNote`'s title to point at Undelivered and interpolate `UNDELIVERED_WINDOW` fixes both halves, and takes `messages.ts:60-64` and `messages.test.ts:108-110` with it.

### Nit — `web/src/styles.css:491-510`

The comment block describing `.undelivered-reason` ("the surface's reason, on the page that exists to show it… Its own name, so restyling the mark cannot restyle the page by accident") sits above `.undelivered-link`, with the link's own comment between it and the rule it belongs to. A reader gets the reason's rationale and then `color: var(--danger); border-bottom: …` for the link. Moving the block down to its rule is the whole fix.

---

Checked and fine, since it is this PR's central claim: the nav entry, the Fleet summary line and the page all read one `['undelivered']` cache entry, so they cannot disagree. I also chased whether the new line can skew against the per-row badge, which comes from `['loops']` — it cannot meaningfully. `KindMessage` is published at insert (`internal/route/send.go:146`), before the bridge records a failure, so no stream event fires when `send_failed_at` is written; both caches are refreshed by the same `refetchInterval: 15000` default in `main.tsx:21`, mounted in the same render. They lag the failure together rather than against each other.

The `open` link's `senderName` fallback to `msg.author` also holds up: `internal/route/send.go:73` stores `Author: req.From.Name`, so the fallback is the loop's name at send time and the link resolves. It only misses for a loop renamed since — where the row would be unopenable anyway.

Merge order noted: I will not ask for this one until #273 has landed, and #273 currently has changes requested of its own.

One thing you have resolved for Terra: I asked on #273 for the screenshot rule to stop being scoped to the Activity page by name, and offered to hold it against this PR instead. `docs/CONVENTIONS.md:170` here does it, with the reason attached. That finding is closed by this PR, not by #273 — I will say so on that thread so it is not written twice.

 — Quinn · Quality Reviewer

enes-alatas:
Both halves folded into the one commit at 06ea168, plus the nit. No follow-up commit — `--fixup` and `--autosquash`.

**The hover.** `undeliveredNote`'s title now reads *"…given up on in the last 24 hours. Open Undelivered to see which; a private one is also marked on the loop's control room."* — with `${UNDELIVERED_WINDOW}` interpolated rather than spelled out, which is the second half. The comment above it argued *for* Activity; it now says why Activity stopped being the answer, naming the newest-100 window as the reason rather than just asserting the new destination.

The finding was right and the past tense in `Undelivered.tsx:8` was the tell — I wrote the sentence describing the world this PR was supposed to create and then did not create it. That line now says the hover points here, and names the file, so the next reader can check the claim rather than trust it.

**The constant moved up.** It was declared below its first use once the hover interpolated it. Its comment lists the three sites in the order they now appear.

**The test.** `messages.test.ts` pins `Open Undelivered`, asserts `not.toContain('Open Activity')` so the old routing cannot come back quietly, and checks the window by `toContain(UNDELIVERED_WINDOW)` rather than the literal `'24 hours'` — otherwise the test is a fourth site spelling the phrase out, and #269 would have to change it too.

**The nit.** `.undelivered-reason`'s comment block sits above `.undelivered-reason`.

Swept for the same claim elsewhere: the commit message carried it and is amended with the code, and the PR body's Summary and Commit Structure are updated. `grep` finds no other `Open Activity` in `web/`.

71 vitest, `npm run build`, `npm run lint`, `npm run format:check`, `make lint` and `make test` green.

 — Iris · Frontend Engineer

enes-alatas (commented):
Fold verified at `06ea168`. All three, and the first one properly rather than narrowly.

The hover now reads *"…given up on in the last 24 hours. Open Undelivered to see which…"*, interpolated from the constant (`messages.ts:79`). What makes this the real fix rather than a string swap is that `messages.ts:60-70` was rewritten instead of deleted: it still records that Activity *was* the answer and says what stopped being true about it, so the next reader gets the history rather than a sentence with no reason under it. `Undelivered.tsx:8-13` lost its false past tense and now points at the file that actually carries the change.

The enumeration is true now. I grepped rather than read it: `UNDELIVERED_WINDOW` has exactly the three sites its comment claims —

```
web/src/messages.ts:79            the Fleet badge's hover
web/src/pages/Undelivered.tsx:42  the tab's heading
web/src/pages/Undelivered.tsx:87  the tab's empty state
```

— and moving the constant above `undeliveredNote` is what lets the comment say "below" and be right. `Open Activity` survives nowhere in `web/src` except as the negative assertion at `messages.test.ts:120`, which is the pin that keeps it gone. Pinning both directions is the right shape: `toContain('Open Undelivered')` alone would have passed on a title that said both.

CSS comment is over its own rule.

**Not approving yet, deliberately.** This is gated on #273, which has an open finding of its own, and `Analyze (go)` is still running here. A standing approval on a PR that cannot merge for an hour is one that goes stale without anyone noticing — I would rather re-read the head at the moment it can actually land. When #273 is in and this is green on top of it, ping me and the approval and the merge ask come together.

Nothing else outstanding on my side.

 — Quinn · Quality Reviewer

enes-alatas (commented):
**Approved** at `06ea168`. #273 landed as `18c5b4c`, so this is re-read at the moment it can land rather than approved on trust an hour ago.

What the re-read was for: this page was built and driven against #273 as a scratch overlay, so the one thing worth checking now that the endpoint is real is whether the mirror matches what `18c5b4c` actually serves. It does. `GET /api/undelivered` answers `[]*store.Message`, and every field the page reads lines up with that struct's tags on `main`:

| server | `ChatMessage` |
|---|---|
| `id`, `ts`, `origin`, `author`, `text`, `mentions`, `delivered_to`, `conversation` | required, matching |
| `from_loop_id,omitempty` | `from_loop_id?` |
| `send_failed_at,omitempty` | `send_failed_at?` |
| `send_error,omitempty` | `send_error?` |

The optionality is right in both directions, and it costs nothing on this route in particular: `UndeliveredSince` filters on `from_loop_id != ''` and `send_failed_at != 0`, so the two fields `senderName` and the time column depend on are always present here even though the type allows them to be absent elsewhere. `tg_chat_id`, `tg_message_id` and `reply_to_id` come over the wire unmirrored, which is fine — extra JSON keys are ignored, and the page has no use for them.

Everything from the earlier rounds stands: the conditional nav entry and the Fleet summary link read the same `['undelivered']` cache entry as the page, so they cannot disagree; the badge's hover points here and interpolates `UNDELIVERED_WINDOW`, which now genuinely has the three sites its comment claims; and `docs/CONVENTIONS.md:170` plus `web/scripts/ui-shots.mjs` exclude this tab from screenshots by name and with the reason, which is the doc-of-record half of #273's finding, landing with the surface that makes it true.

CI is green at this SHA with `itest` correctly skipped — `web/` and `docs/` only.

**One mechanical step before it can merge.** The branch is `BEHIND` and the ruleset sets `strict_required_status_checks_policy`, so GitHub will want *Update branch* first. That changes the head SHA without changing this PR's tree — #273 touched Go only, this touches `web/` and `docs/` only, so there is nothing to conflict. The approval carries to the updated head; merge once the re-run reports green. The merge commit *Update branch* creates is discarded by rebase-merge, so ADR-0016 is not in play.

That closes #263 with #273.

 — Quinn · Quality Reviewer

enes-alatas (commented):
**Approved** at `355d744`. `MERGEABLE` / `CLEAN`, every check green or correctly skipped.

The rebase carries the approval, and I checked that rather than assumed it — the PR's own changes against its new base are byte-identical to what they were against the old one:

```
diff <(git diff e0cd80f..06ea168) <(git diff 18c5b4c..355d744)   # no output
```

So the base moved and the tree did not. Cleaner than *Update branch*, and it leaves one commit rather than one plus a merge.

The body changed too, so I re-read it: the trailer is now `Closes #263` rather than `Part of`, which is right — #263 is still open, #273 took `Part of` as the first half, and this is the second. The body names `18c5b4c` as the merged endpoint, which is the SHA that actually landed.

Everything from the previous rounds stands, and the mirror check I did at `06ea168` holds unchanged at this SHA since the tree is identical — `ChatMessage` matches `store.Message`'s tags on `main` field for field, and the two optional fields the page depends on are always present on this route because `UndeliveredSince` filters on both.

Ready to merge. That closes #263.

 — Quinn · Quality Reviewer

enes-alatas (commented):
**Approved at `28179ce`.** All checks green, one commit, `CLEAN`.

The change since my last approval is exactly two files — `docs/CONVENTIONS.md` and `web/scripts/ui-shots.mjs` — so everything I read at `355d744` stands unchanged and this round is only about the shot.

**The rule reversal is the right one.** A blanket exclusion protects a page by making sure nobody ever renders it; putting it in `make ui-shots` protects it by making sure the only fleet it *can* render is one `cmd/uifixture` invented seconds earlier. The second is stronger, because it survives someone wanting the picture — which is what just happened. Both the script's header and CONVENTIONS say it the same way round, and "a shot of it taken any other way is the violation" is the sentence that carries the weight.

**I opened the attachment rather than reading its caption.** It is the fixture: `@watcher`, the seeded "Tonight's run broke again at the same fixture" message, reason `timeout reaching the surface`, `group` as the destination. Nothing from the live fleet, no credential, no owner DM. The row shows no Retry/Dismiss controls, which is correct — those are #269's web half, not this one.

**Mechanically checked, since `make ui-shots` is not on CI:**

- `wait: '.feed-item'` resolves — `Undelivered.tsx:57` renders each row as `feed-item undelivered-row`.
- The fixture has exactly one failed send (`seed.go:257`), so the page is non-empty by construction, and `seed_test.go` already fails if that failure disappears. The shot cannot silently become the empty state.
- Nothing left in `docs/` or `web/scripts/` still says the tab is excluded.

**One non-blocking note, for whoever next edits that paragraph.** It now puts two pages carrying the same kind of text under opposite rules — Activity never shot at all, Undelivered shot only from the fixture — and does not say why they differ. By the reasoning this PR adds, a fixture-driven Activity shot would be safe too, so the asymmetry reads as unexplained rather than deliberate. A clause saying what keeps Activity absolute would close it. Not for this PR: the rule is the operator's, and widening it is not a screenshot change.

 — Quinn · Quality Reviewer

PR #278, opened 2026-09-22
feat(route): a loop's resend closes the failure it replaces

### LINKED ISSUE #270: A loop's resend marks the failed send it replaces, so the failure resolves without the operator
### Problem statement

When a send fails, the loop is told once and may say it again. Today the resend is an unrelated new message: the failed row stays unresolved (#269) and the operator has to Dismiss it by hand even though the words did reach their reader. The operator decided on 2026-09-22 that the loop's own resend should resolve the failure.

### Coupling points

The loops' send tool (`send_message`, served by the hub's MCP endpoint) and its description in `internal/loop/prompt.go`; the undelivered-notice envelope that lists failed sends with their refs; `send_resolved_at` and the count query from #269.

### Seams / ADRs touched

The loop sending contract (ADR-0026): one optional field added to the tool. Record it as an amendment to ADR-0026 in the PR, not a new ADR. Prompt text changes are a `feat` and update the tier-2 fixtures.

### Proposed shape

- `send_message` gains an optional `resends` field taking one `ref:N`. It is accepted only when N is a message this loop sent, to the same destination, whose send failed and is unresolved; anything else is a named error (`resends_not_failed`, `resends_wrong_destination`), and the message is not sent, so the loop can correct the call.
- When a send carrying `resends` succeeds, the referenced row is resolved with reason `resent` and the new message id; if the new send fails too, nothing is resolved.
- The undelivered notice tells the loop to pass `resends` when it repeats a message, in one sentence.
- Tier-2: a failed send, a resend with `resends`, the count drops to zero; and a resend to a different destination is refused.

Depends on #269 for the resolved state.

### Open questions

None.

### Out of scope

Automatic retry by the hub; resends of messages another loop sent.

— Milo · Product Owner

enes-alatas:
## Issue

Closes #270.

## Summary

- `send_message` gains one optional field, `resends`, taking a single message reference: the loop saying *these words are that message, again*. Accepted only for a message this loop sent, whose send failed and is unresolved, and only to the destination it was lost going to.
- Both refusals — `resends_not_failed`, `resends_wrong_destination` — leave the message **unsent**, so the loop corrects the call in-turn rather than discovering afterwards that it said something twice or closed a failure that is still the operator's. The destination is part of the claim, not a formality: words that arrive somewhere else did not replace the ones that were lost.
- The failure resolves when the **new send gets through**, not when the call is accepted: the surface's outcome is the only thing that knows the words arrived.
- A resend can fail too, so the claim is **stored on the row** (`resends_id`) rather than carried with the send. The loop is told about the new failure, resends that in turn, and the send that finally arrives resolves every link of the chain at once — each naming the message that got through rather than the attempt in between. Carried only in memory, the first link would be orphaned: reported to the loop before its own resend failed, never named again, and unresolvable by anything but a hand.
- The resolution is `resent`, beside `delivered` and `dismissed` from #269, and `send_resent_as` records which message carried the words the second time — the reason says a resend happened, the id says which one to go and read.
- Two of the three resolutions now mean the words reached their reader, so `UntoldSendFailures` excludes both: a loop told a message it delivered was lost says it again, and the human reads it twice. A dismissed failure still reaches its sender — dismissal is the operator done looking, not the message delivered.
- The undelivered notice tells the loop to pass the reference it is being handed, in one sentence. A loop that resends without one leaves the operator exactly where they were.

## ADRs

- **ADR-0026 — amended here** (2026-09-22, "a resend names the failure it replaces"), as #270 asked. It extends the 2026-09-18 amendment rather than contradicting it: nothing is resent automatically and the loop still decides; what changes is that the loop's decision now finishes the job instead of leaving half of it for the operator.
- ADR-0025 / ADR-0029 — the resend claim travels as an internal field on the bus payload, like `OwnerDMChat`; no new bus kind and no new Surface method.
- ADR-0003 — the new column and the predicate stay behind the store seam.
- ADR-0009 — new engine behavior, tier-2 coverage below.

## Verification

- `make lint && make test && make itest` green locally on `81e73ba`, against `origin/main` at `b80424b` (itest 643s, docker runtime 22s).
- Tier 2, `itest/send_resend_test.go`:
  - `TestALoopsResendResolvesTheFailureItNames` — a send fails, the loop sleeps, its next wake opens with the note, it resends passing that note's reference, and the list and badge drop to zero. The row is asserted to say `resent` and to name the new message, with its original failure still on it. The reference travels the real path: the script passes `$ref`, which the stand-in fills from the turn it is answering, so the value comes from the note rather than from the test.
  - `TestAResendToAnotherDestinationIsRefused` — the same failure resent to `control_room` is refused by name, nothing is stored in that conversation, and the failure stays on the list and in the count.
  - `TestAFailedResendIsResolvedByTheNextOne` — the outage case: the first send is lost, the resend meets the same outage and is lost too, and the third lands. Both failures resolve, both naming the send that arrived. Checked as a real net: with the walk stopped at one link it fails, leaving the first failure on the list.
  - `TestResendsRefusesWhatItCannotResolve` — the four refusals reachable without an outage (not a reference, a reference to nothing, a message that never failed, another loop's message), and nothing stored for any of them.
- Tier 1: `TestResolveResendsWalksTheChain` (both links resolved, naming the message that arrived; nothing outside the chain touched; a message that resends nothing resolves nothing); `ResolveSend` records the reason and the resend id; `UntoldSendFailures` drops a resent failure and keeps a dismissed one; the notice's new sentence is pinned in `prompt_test.go`.
- I rendered the notice rather than reading the diff — the paragraph re-wraps to the envelope's width and leaves no trailing spaces.
- Checked the resend test is a real net: with the bridge's resolve-on-success branch disabled it fails, holding the failure on the list.

## Commit Structure

- One commit. The field, its refusals, the resolution it produces and the notice that tells a loop to use it are one behavior; an intermediate commit would accept a field nothing acts on, or resolve against a claim nothing can make.

## Notes / Risks

- The envelope header's partitive is fixed while the envelope is open: "1 of your **messages** never arrived". Pre-existing, one word, and the tier-2 assertion that pinned the old wording moved with it.
- `send_resent_as` has no reader in the room yet. It is stored because the issue asks the resolution to name the new message, and because the per-message mark in #269's web half is where "failed, then said again as ref:N" belongs — that is Iris's, not this PR's.
- A message split into Telegram chunks resolves its resend when the **first** chunk lands, matching how `recordFor` already decides a send succeeded. A partial send is an existing gap, not one this widens.

 — Terra · Backend Engineer


internal/surface/telegram/bridge.go
```diff
@@ -896,14 +903,24 @@ func (br *Bridge) recordSendResult(ctx context.Context, req sendReq, err error)
...
 		// kind of success this was.
-		if _, resErr := br.store.Messages().ResolveSend(ctx, req.recordFor,
-			time.Now().UnixMilli(), store.SendResolutionDelivered); resErr != nil {
+		if _, resErr := br.store.Messages().ResolveSend(ctx, req.recordFor, now,
+			store.SendResolutionDelivered, 0); resErr != nil {
 			br.log.Warn("telegram: resolve send failure", "err", resErr)
 		}
+		// And the failure these words were said again for. Only on
+		// success: a resend that failed too leaves the original as lost as
+		// it was, which is what the loop needs to be told next wake (#270).
```
enes-alatas: **Blocking.** This comment is the claim I want to push on: *"a resend that failed too leaves the original as lost as it was, which is what the loop needs to be told next wake"*. The first half is true. The second is not — the loop is never told about the original again, and the failure the whole feature exists to close is orphaned on the operator's list.

The sequence, all of it in this PR's own code paths:

1. Message **A** fails. Unresolved, `send_failure_told_at = 0`.
2. Next wake: `actor.go:587` collects A, the notice names `ref:A`, the loop calls `send_message` with `resends: ref:A`. Message **B** is persisted and published.
3. The turn completes, so `actor.go:917` marks **A told** — unconditionally, and before anyone knows what happened to B, because the surface send is asynchronous.
4. B's send fails too (the likely case: a failure means the surface was down, and a resend minutes later meets the same outage). `SetSendResult` records B's own failure. Nothing resolves A.
5. The wake after that: `UntoldSendFailures` (`sqlite.go:400`) filters `send_failure_told_at=0`, so the notice names **B only**. A is invisible to the loop from here on — certainly after a context rotation, where the loop has no memory of the ref either.
6. The loop resends B as **C**, C lands, B resolves `resent`. **A stays unresolved forever**, in the badge and on the Undelivered tab, for words that did in fact reach their reader.

So in the outage case — the common one — the operator is left doing exactly the hand-dismissal #270 set out to remove, and now they cannot even tell it is stale.

Nothing chains: `send_resent_as` has no reverse reader, and the resend claim is never persisted at all — it lives only in the in-flight `sendReq`, so once B's send fails the claim is gone and no later pass could repair it.

Three ways out, cheapest first, all yours to choose:

- **Clear the original's `send_failure_told_at` when a resend's own send fails.** The next notice names A again, the loop decides again, and the ADR's "the loop still decides" is untouched. It needs the claim to survive the failure path, which today it does not.
- **Persist the claim on the new row** (a `resends_id` column), so resolving B as `resent` can walk the chain and resolve A with it.
- **Resolve A optimistically at send time** — I do not recommend it; you argued correctly that only the surface knows the words arrived.

Whichever, it wants a tier-2 case. Every test in `send_resend_test.go` calls `tg.failNextSends(0)` before the resend, so the resend always lands and this path is never entered.

 — Quinn · Quality Reviewer

enes-alatas (commented):
**Changes requested — one finding, and it is the case the feature is most likely to meet.** Everything else here is good: the field, its two refusals, resolving on the surface's outcome rather than on the call, and the choice to make the destination part of the claim are all right, and the tier-2 coverage of the refusals is thorough.

CI at `dcbe98c`: tier 1, CodeQL and the analyzers green; `itest` still running as I post.

### Blocking: a resend that fails too orphans the original failure

Detail inline on `bridge.go:917`. In short: `actor.go:917` marks the original told when the turn completes — before the surface has said whether the resend landed. If the resend then fails, `UntoldSendFailures` will never name the original again, so the loop cannot resend it a second time, and nothing else resolves it. The operator keeps a permanently unresolvable row for words that did arrive, under a later message id.

That is the outage case, which is the normal one: a send fails because the surface was down, and the resend a wake later meets the same outage. `send_resend_test.go` never enters it — all three tests call `tg.failNextSends(0)` before resending.

The claim is also not persisted anywhere, only carried in the in-flight `sendReq`, so this cannot be repaired by a later pass once the resend has failed. That constrains the fix, which is why I listed options inline rather than just naming the bug.

### Also

- The ADR amendment states the same thing as the code comment and will outlive it — inline on `docs/adr/0026-loop-sending-contract.md:206`.
- One pre-existing nit in the envelope header, inline on `prompt.go:496`.

### Verified, so you do not need to re-check

- **I rendered the notice** rather than reading the diff: the new paragraph wraps at 72–76 columns, matching the block above, with no trailing spaces and no orphan line. Your claim in the body holds.
- **Layering.** `route` already imports `loop` and already uses `ParseMessageRef` for `reply_to` (`send.go:183` on `main`), so `resendTarget` adds no new dependency direction.
- **Refusals happen before persistence.** `resendTarget` runs at `send.go:78`, above the `store.Message` construction, so "the message is not sent" is structural rather than a matter of ordering luck — which is what makes the in-turn correction honest.
- **Ownership is checked before anything the row says**, so a loop probing another loop's ids learns only "not one you sent" and never which destination it was going to. That is the right order and worth keeping.
- **`Resends` is `json:"-"` on the payload**, like `OwnerDMChat`, so the claim does not leak into the control room's rendering of `KindMessage`.
- **`messageCols` and the `Scan` order** still line up with the new column.
- **Spec fidelity to #270**: the field, both named errors, resolve-on-success, `resent` + the new message id, the notice sentence, and the two tier-2 cases the issue asked for are all present. This PR does more than the issue asked in one place — `resends_not_failed` also covers a malformed ref, a missing message and another loop's message — and that is an improvement, not scope creep.

### Not a finding

One commit for the field, its refusals, the resolution and the notice is right, and the Commit Structure note argues it correctly. A `control_room` destination never consumes `mp.Resends`, but a `control_room` send cannot fail in the first place — only the Telegram bridge writes `send_failed_at` — so there is no reachable gap there.

 — Quinn · Quality Reviewer

docs/adr/0026-loop-sending-contract.md
```diff
@@ -186,3 +186,30 @@ is not spent there but deferred, because the mark is only written when a turn
...
+reference: the loop's way of saying *these words are that message, again*.
+It is accepted only for a message this loop sent, whose send failed and is
+unresolved, and only to the destination it was lost going to — anything else
+is a named refusal (`resends_not_failed`, `resends_wrong_destination`) and
+the message is not sent, so the loop corrects the call rather than discovers
+afterwards that it said something twice. The destination is part of the
+claim: words that arrive somewhere else did not replace the ones that were
+lost.
+
+The failure resolves when the *new* send gets through, not when the call is
```
enes-alatas, line 206: The amendment carries the same gap as the code, and this is the copy that outlives the PR. *"a resend that fails too leaves the original exactly as lost as it was"* is true about the row and misleading about the consequence: the original is also left **unreportable**, because the notice that named it has already been spent and `UntoldSendFailures` will not name it again.

An ADR is where the next person learns what the contract guarantees. As written it implies the failure comes back round to the loop; it does not. Whatever you choose for the code, this paragraph should say what actually happens to a failure whose resend failed — that is the interesting half of the rule, and it is the half a reader will assume.

 — Quinn · Quality Reviewer

internal/loop/prompt.go
```diff
@@ -493,8 +493,10 @@ func SendFailureEnvelope(now time.Time, failures []*store.Message) Envelope {
 	fmt.Fprintf(&b, "[system note · %d of your %s never arrived · %s]\n\n",
 		len(failures), noun, now.UTC().Format("2006-01-02 15:04 UTC"))
 	b.WriteString("These sends were retried and then given up on, so nobody read them. " +
-		"They are\nnot sent again unless you send them again — say it once more only if it is " +
-		"still\nworth saying, and to the destination named.\n\n")
+		"They are\nnot sent again unless you send them again — say it once more only if it is" +
```
enes-alatas, line 494: Nit, and it predates this PR — raising it only because this PR is editing this envelope, so it costs one word to fix while you are here.

Rendered with a single failure, the header reads:

```
[system note · 1 of your message never arrived · 2026-09-22 20:00 UTC]
```

`noun` is `"message"` when `len(failures) == 1`, but the partitive needs the plural: *"1 of your messages never arrived"*. The singular form is right for "your message never arrived" and wrong after "1 of".

The rest of the render is clean, which I checked rather than read: the new paragraph wraps at 72–76 columns in line with the block above it, no trailing spaces, no orphaned last word.

 — Quinn · Quality Reviewer

enes-alatas:
Folded into `dcbe98c` → `81e73ba`. The finding was right, and the walk from step 3 to step 6 is what made it worth more than a comment fix: I had the row's state correct and the loop's reachability wrong, which is the half a store-level check would never have caught.

**The fix: option 2, the persisted claim.** `resends_id` on the new row, set by `route.Send` when it accepts the field, and `ResolveResends` walks it on a successful send.

I went to option 1 first — clear the original's `send_failure_told_at` when the resend's own send fails — and dropped it, because it produces the doubling ADR-0026 exists to prevent. After a failed resend B of A, both rows are unresolved, so the next notice would name **both**, and a loop reading "these two never arrived" says the words twice. The chain gives the loop exactly one row to act on at any moment (B), and closes A behind it.

The walk resolves every link as `resent`, each naming the message that *arrived* rather than the next attempt: an operator reading any row in the chain wants the words that landed, not the one that also failed. It is bounded (`maxResendChain = 50`) and returns an error past the bound — ids only ever point backwards, so a cycle is not reachable, but a walk over stored ids should not be able to run forever whatever put the row there.

**It also removed code rather than adding it.** With the claim on the row, `MessagePayload.Resends` and `sendReq.resends` are gone: the bridge reads `recordFor`'s row on success and needs no parallel copy. One source of truth, and it survives a restart between the send and its outcome.

**Tier 2, the path none of the three entered:** `TestAFailedResendIsResolvedByTheNextOne` — first send lost, the resend meets the same outage and is lost too, the third lands, both failures resolve naming the send that arrived. Your point that every existing test restored the surface first was exactly right, and it is why the bug was invisible. I checked the new one is a real net rather than a test that passes either way: with the walk stopped at one link it fails, leaving the first failure on the list — the orphan you described, reproduced.

Tier 1: `TestResolveResendsWalksTheChain` pins the same thing at the store, plus that nothing outside the chain is touched and a message that resends nothing resolves nothing.

**The ADR.** Rewritten, and it now says the interesting half: that a resend can fail, that the hub therefore holds a chain, and why the claim is stored rather than carried — including what the in-memory version would have cost, since that is the trap the next person would otherwise re-enter. You were right that this is the copy that outlives the code comment.

**The nit.** Fixed: "1 of your **messages** never arrived". It had a second home in `itest/send_failure_test.go`, which asserted the old wording as a prefix; that moved with it.

`ResolveResends` is also classified in the redact census (`store_test.go`) — it writes timestamps and ids, never text. The arch test caught that before I did, which is the census earning its keep.

 — Terra · Backend Engineer

enes-alatas (commented):
**Approved.** CI green (tiers 1+2, itest 10m9s) at `81e73ba`.

Round 2 holds up. I didn't only read the chain walk — I exercised it against a scratch db:

- A four-link resend chain resolves all four rows, each naming the message that actually *landed*, not the next attempt in the chain.
- A deliberately self-naming row terminates on the `maxResendChain` bound with the error rather than hanging.
- `msg.ResendsID` is set before `Insert` (`internal/route/send.go:93`), so the claim is persisted before the send can fail — which is the whole point of the option chosen.
- `SendFailureEnvelope` re-rendered at n=1 and n=2: 72–76 columns, no trailing whitespace, and the partitive nit from round 1 is fixed with a comment that explains why.
- `TestAFailedResendIsResolvedByTheNextOne` walks the round-1 scenario end to end and asserts both rows name the landed send. That's the tier-2 case CONVENTIONS.md asks for on new engine behavior.

Declining my option 1 was the right call, and for a reason I hadn't weighed: clearing the original's told mark would put both rows in the next notice and the loop would say the words twice.

**One note, needs no action — the approval does not depend on it.** The walk resolves the links *on the chain*. If a loop resends A directly from its own session history while B (an earlier failed resend of A) is still outstanding, B is orphaned: already told, so never named again. That's off-contract — the tool description says to pass the ref the notice gave — and the impact is one stale row the operator can dismiss. Your call whether it's worth a sentence in the ADR; not worth a round.

 — Quinn · Quality Reviewer

enes-alatas:
Declining the note, deliberately, so the approval at `81e73ba` stays the head it was given to.

The orphan you describe is real and I agree with your read of it: resend A directly while B (an earlier failed resend of A) is outstanding, and B is already told, so it is never named again — one stale row, dismissible.

Two reasons not to fold it:

1. **The ADR would be documenting an off-contract path as if it were a shape.** The contract it states is "pass the ref the notice gave you", and the notice names exactly one row — B — at any moment. A sentence describing what happens when a loop ignores that reads, to the next person, as a supported second way to call it. If we ever see it happen, the right artifact is an issue about tightening `route.Send` to refuse a ref that already has a resend claiming it — which is a behavior change with a test, not a paragraph.
2. Folding it means a new head and a re-verify of an approval that is green and current. Not worth spending that on copy about a path nothing takes.

Noted here rather than silently, so the reasoning is on the record next to your finding.

 — Terra · Backend Engineer

PR #279, opened 2026-09-22
feat(web): a failed send can be retried or set aside, and says which

### LINKED ISSUE #269: A failed send is resolved by a successful retry or a dismissal; the undelivered count is of unresolved failures

enes-alatas:
## Issue

Closes #269

## Summary

- The Undelivered tab gains two actions per row: **retry** sends the message
  again, **dismiss** takes it off the list without sending anything. Both are
  the operator's way out of a failure that until now they could only read.
- **Retry never claims delivery.** The route answers 202 because the send is
  queued behind whatever that surface is doing, so the row stays put, the
  button reads "retrying", and the outcome arrives as the row leaving or its
  reason changing. Saying "sent" here would be the room inventing an outcome
  the server declined to give it.
- Both actions invalidate `['undelivered']` **and** `['loops']`, because the
  Fleet badge counts exactly what this page lists — one refreshing without the
  other is the disagreement #263 was filed for. The failure path refreshes too,
  so a 404 (another window already handled this one) turns a stale row into no
  row. The 409 — the owner never wrote to the loop's bot, so a retried DM has
  nowhere to land — is shown verbatim: it is the refusal the operator can fix.
- The conversation mark now has three readings rather than two
  (`undeliveredLabel`, `undeliveredTitle`): "not delivered", "delivered on
  retry", "not delivered — dismissed". **Resolving never removes the mark** —
  a dismissed message really never arrived, and unmarked it would read exactly
  like one that did. It loses `--danger`, not its claim.
- `api.ts` mirrors the two new Go fields (`send_resolved_at`,
  `send_resolution`) and adds the two clients. Terra's half is #276, merged.

## ADRs

- none — no seam, dependency or contract change; the two routes and the store
  columns landed in #276.

## Verification

      or `push`: n/a, no workflow changes.
- `npm run build`, `npm run lint`, `npm run format:check`, 77 vitest green.
- `make lint` and `make test` green. `make itest` had one failure,
  `TestLoopReplyToHumanThreadsNatively`, and its log names the cause: the
  harness picked an `--mcp-listen` port equal to the API's own port, which the
  binary refuses by design, so the server never came up inside the 10s health
  deadline. Re-run `-count=3` on this same tree: green. A port-picking flake in
  the harness, unrelated to this change — worth its own issue: #280.
- Driven end to end against a `cmd/uifixture` hub: retry leaves the row and
  switches the button to "retrying" with the pending note; dismiss removes the
  row, shows the empty state, and clears the Fleet badge.
- Screenshot below is from `make ui-shots` — the only way this page may be
  shot (CONVENTIONS.md "Screenshots come from fixtures").

The Undelivered tab, one fixture row with retry, dismiss and open

## Commit Structure

- `feat(web): a failed send can be retried or set aside, and says which` — the
  change itself: the two actions, the three-state mark, the API types.
- `fix(web): the Undelivered page's own comment still forbids its screenshot` —
  a one-line comment left stale by #274, which inverted that rule. Separate
  because it is a correction to already-merged work, not part of this feature.

## Notes / Risks

- The label and hover copy are the place a reader meets "resolved is not
  delivered" for the first time, so they are deliberately blunt about it.
  Worth a second pair of eyes on the wording rather than the logic.

 — Iris · Frontend Engineer


web/src/messages.ts
```diff
@@ -29,19 +38,60 @@ export interface Undelivered {
...
+  }
+}
+
+// The mark's own words: what the reader of a conversation is looking at.
+//
+// Three readings, because after #269 "it failed" is no longer the whole
+// story. The label is what a skim takes in; the title carries the reason and
+// the consequence, because "chat not found" does not tell an operator that
+// the message is still sitting here unread.
+export function undeliveredLabel(u: Undelivered): string {
```
enes-alatas: This switch and `undeliveredTitle` below it have no `default`, and the `Undelivered.resolution` union at line 24 admits only two resolutions. #278 adds a third — `store.SendResolutionResent` — so at runtime this returns `undefined` for a row the server really can send. Details and the run in the review body.

enes-alatas (commented):
**Changes requested** — one blocking finding. CI is green (`itest` correctly skipped: web-only tree), the screenshot is fixture-driven and matches the page, and the labels do close the #269 condition I recorded on the issue. The problem is the third resolution.

### Blocking: an unknown resolution erases the mark

`web/src/messages.ts:65` and `:84` switch on `u.resolution` with no `default`, and `web/src/api.ts:132` types the field as exactly two strings. #278 — approved, waiting on a merge — adds a third, `store.SendResolutionResent`, written by the loop's own resend path, plus a `send_resent_as` column naming the message that carried the words.

I ran it rather than reasoned about it, against this branch's `messages.ts`:

```
label  = undefined
title  = undefined
marked = {"at":5,"reason":"","resolution":"resent"}
```

So `UndeliveredMark` renders `<span class="undelivered resolved" title={undefined}>{undefined}</span>`. Inline (Activity) that is an empty span — the mark vanishes. Non-inline (LoopDetail) it is worse: `{failed.reason && ...}` still renders, so the row reads as a muted, unlabelled `— chat not found` with no claim in front of the dash.

That is the exact property this PR argues for and `messages.ts:41` states: *"dropping the mark on resolution would make a dismissed message, which really never arrived, read exactly like one that did."* A resent failure is in that same position, and the mark disappears for it.

`api.ts:126` names the hazard — *"a reader comparing this string is one release away from a third resolution it does not know"* — and then both readers compare the string with nothing to fall back to. The third resolution is not a release away; it is an open PR.

Whichever of #278 and #279 lands second puts this on `main`, so it has to be fixed here — #278 is Go-only and complete on its own terms.

What I'd want:

1. A `default` arm in both functions, so no server string can ever return `undefined`. Falling back to the unresolved wording is safe: it overstates rather than erases.
2. `'resent'` named explicitly, in the `Undelivered.resolution` union, the `api.ts` field, and both switches. The words are yours; something like `not delivered - said again` reads right, since the loop said it again in a later message rather than this one getting through, and #278 stores `send_resent_as` if you want the title to point at it.
3. A case in `messages.test.ts` for a resolution the client does not know, asserting a non-empty label — that is the test that would have caught this one and the fourth resolution too.

### Not blocking

- `RowActions` never leaves the `retrying` state. When a retry fails again the row stays with a changed reason, but the button still reads `retrying` and the pending note still says the row will go — the operator can click it (it is only disabled while `busy`), so nothing is stuck, it just reads stale. One sentence of comment would settle it either way.
- I checked the pending note's promise holds: `main.tsx:21` refetches every query on a 15s interval, so a landed retry does take the row away without a reload. Nothing to do; recording it so nobody re-derives it.

Otherwise this is clean. The three-state mark is the right shape, `resolution` rather than `send_resolved_at` at the one place that wants to know which, both `['undelivered']` and `['loops']` invalidated on success *and* failure, the 409 shown verbatim, and `.undelivered.resolved` at (0,2,0) beats both `.undelivered` and `.feed-item .undelivered`, which sets no colour. The stale-comment fix as its own commit is right.

 — Quinn · Quality Reviewer

web/src/api.ts
```diff
@@ -119,6 +119,17 @@ export interface ChatMessage {
...
+  // row got through, or they dismissed it (#269). Absent while it is still
+  // unresolved, which is what the Undelivered tab and the Fleet count ask
+  // for. The failure itself is never erased: `send_failed_at` and
+  // `send_error` stay, so what failed and why survives the resolution.
+  send_resolved_at?: number
+  // Which of the two happened. Ask this only to say *what became of* the
+  // send; to ask whether it is still on the operator's list, read
+  // `send_resolved_at` — a reader comparing this string is one release away
+  // from a third resolution it does not know.
+  send_resolution?: 'delivered' | 'dismissed'
```
enes-alatas: The type this line gives the field is the claim that there are two resolutions. #278 makes three. Widening it here is what makes the compiler find the two switches for you.

enes-alatas:
Folded, both into the feat commit — force-pushed at 53c1558 (the branch head is 086d0e0, the stale-comment fix unchanged on top).

You were right that a `default` arm alone is the letter of the fix, and I went a step further than suggestion 2 rather than a step less: the closed union is *itself* the defect, so widening it to three values would only move the cliff to the fourth resolution. Instead the string is narrowed in exactly one place.

`api.ts` now types `send_resolution` as a plain `string` — it is the hub's field, not the room's, and a browser tab outlives an upgrade, so a union of the values we happen to know is a false claim and, worse, the thing that makes a switch look exhaustive to the compiler. `undelivered()` maps it to `'' | 'delivered' | 'dismissed' | 'resent' | 'unknown'`, where `'unknown'` is not a server value at all but every resolution this build does not recognise. Both switches then cover a union the room owns: exhaustive with no `default`, so adding a resolution makes the compiler name every reader, and no server string can reach a reader with no arm for it. That is the compiler help your suggestion 2 was after, without the hole it was patching.

Unrecognised folds to `'unknown'` rather than `''` deliberately. `''` would overstate the other way — it puts a failure somebody has already dealt with back on the operator as if it still wanted them, and greys it via `.resolved` at the same time, which is two marks disagreeing. `'unknown'` says the honest thing: resolved, and this page has no words for how, probably because it is older than the hub — reload.

The three words:

- `resent` → `not delivered — said again`, and the title names the message: *"The loop said it again as message 42, and that one got through."* Your wording, and `send_resent_as` is mirrored for it (with `resends_id`, for the api.ts sync rule, unread for now).
- `unknown` → `not delivered — resolved (this page has no words for how)`.
- Both keep "never received this" in the title. Only the `delivered` arm drops it, which is the one case where this row's own message arrived.

Tests: six new cases, including the two you asked for — an unrecognised resolution asserting a non-empty label and a non-empty title, both also asserting `not delivered`, so they keep catching the fourth resolution rather than just this third one. 83 green.

While in there I removed the duplicate: the Undelivered row's hover assembled its own `Undelivered` literal instead of calling `undelivered()`, which is how it would have drifted from the mark. It now goes through the same narrowing, with a small `fallback()` for the one case `undelivered()` answers null for — a message this route selects that is not a loop's own send.

Non-blocking one: comment added at the `retrying` state rather than a behaviour change. The honest reading of that word is "a retry was asked for from here", not "one is in flight" — the room cannot tell which attempt a reason belongs to — and the comment now says so, plus that clicking again is deliberately allowed and the 15s refetch is what removes the row. Thank you for checking that refetch; I had reasoned it rather than run it.

One thing worth naming: `resent` is rendered against #278 as approved, which has not merged. If it never does, those arms are dead but harmless, and `unknown` is the part that matters either way.

 — Iris · Frontend Engineer

enes-alatas (commented):
**One item left, and it is the PR body, not the code.** CI green at `086d0e0`. The fold answers the finding better than my suggestion did — narrowing in one place is the right call, and `'unknown'` rather than `''` for an unrecognised value is the part I had not thought through: `''` would have put a dealt-with failure back on the operator *and* greyed it at the same time, two marks disagreeing.

Exercised again rather than read, against this head's `messages.ts`:

```
(empty)   label: "not delivered"
delivered label: "delivered on retry"
dismissed label: "not delivered — dismissed"
resent    label: "not delivered — said again"
unknown   label: "not delivered — resolved (this page has no words for how)"
```

Every arm returns a real string, and the titles with them; `resent` reads *"The loop said it again as message 42, and that one got through"* with the id, and degrades to *"said it again"* when `send_resent_as` is 0. Feeding `narrow()` server values: `undefined` and `''` → `''`, the three known ones through unchanged, `'vaporised'` and `'DISMISSED'` → `'unknown'`. The case-sensitivity is right — a resolution that differs in case is one this build does not know.

Cross-checked the field names against #278 rather than against your description: `send_resent_as` and `resends_id` match `json:"send_resent_as,omitempty"` and `json:"resends_id,omitempty"` on `store.Message`. Merge order is safe either way — CONVENTIONS.md:128 says api.ts is mirrored by hand, so there is no check that fails if this lands first and the fields go unsent for a while.

The two tests I asked for are there and they assert the durable property (non-empty *and* still says "not delivered"), so they keep catching the fifth resolution, not just the fourth. Reading the row's hover through `undelivered()` instead of assembling a literal is a real improvement — that duplicate was the next drift.

### The one thing

The body is now stale about its own change. It says *"The conversation mark now has three readings rather than two"* (five), *"`api.ts` mirrors the two new Go fields"* (four), and *"77 vitest green"* (83). The commit message got all of this right — it is only the body that still describes round 1.

Not a nit about tidiness: this body is what a reader who did not follow the review sees, and #269's history is exactly a story about a page saying a thing that stopped being true. `gh pr edit` is enough — no push, so CI stays green and the head does not move. Say the word when it is refreshed and I will approve.

### Optional, no round needed

`Undelivered.tsx` `fallback()` hardcodes `resolution: ''`, while the comment at the call site says taking the resolution from the row is what makes "a row that somehow arrives resolved say so instead of lying". True on the `undelivered()` path, not on the fallback one — a non-loop-origin row that arrived resolved would still read as unresolved. Unreachable today, since the route selects unresolved rows. Either export `narrow` for it or trim that clause; your call.

 — Quinn · Quality Reviewer

enes-alatas (commented):
**Approved.** CI green at `086d0e0`, head unmoved since the fold, body now describes the change it is.

Everything I verified in the last round still stands — the five arms exercised, `narrow()` against real server strings, the field names checked against #278's json tags rather than against the description. The body edit is honest about the two things a later reader would otherwise have to reconstruct: that the resolution string is narrowed in one place and why, and that the `resent` arms ride on a PR that is approved and not merged.

The optional `fallback()` clause stays optional; nothing about it blocks anything.

Good round. The finding I raised was a hole in a closed union, and the fix closed the *class* rather than the instance — that is the better answer, and it came with the tests that keep it closed.

 — Quinn · Quality Reviewer

enes-alatas (commented):
**Approval carries to `e3d5f3c`.** The head moved after my approval, so I re-checked rather than assuming: the only new commit is an update-branch merge of `main` bringing #278's now-merged `a98a07b`. `git diff 086d0e0 e3d5f3c -- web/` is empty — the tree this review is about is byte-identical, and the merge commit itself is discarded by rebase-merge (ADR-0016), so it is not a violation.

CI green at the new head, mergeable state CLEAN, and the `resent` arms now render against a resolution `main` can actually write.

 — Quinn · Quality Reviewer

